PR #1, opened 2016-04-29
Move validation model to dataactcore

bsweger:
This PR contains the changes for moving the validation database models to core with the rest of the db models:

* Model file moves to dataactcore/models
* Existing validation migration scripts are merged into the core migration chain
* Validation db initialization process updated to use the alembic scripts

**Note:**: The migration scripts in this PR are updated to work with the new migration templates that expect upgrade/downgrade methods for all four databases (user, error, job, and validation). Any new migrations generated after this PR is merged will be formatted correctly (see the mako template), with no need for manual intervention.


jworcestBAH:
All looks fine, I did remove a drop constraint command in one of the migrations that was causing errors.  Even if moving to an earlier version there's no real reason to drop that constraint anyway.  Merging.

PR #2, opened 2016-04-29
Db 453 session

jworcestBAH:
Session handling now goes through headers instead of cookies to be consistent with IE security settings.  Also includes fixes to session expiration handling.


dataactcore/scripts/databaseSetup.py
```diff
@@ -41,3 +43,5 @@ def runMigrations(alembicDbName):
         if "relation" and "already exists" in e.message:
```
bsweger: That's strange. I think I pushed a change to this bit of code that checks the errorType instead of looking in e.message (in response to your comments in a code review). Is this a victim of the repo merge?

bsweger:
@jworcestBAH Done reviewing. This all looks good after reviewing and giving it a spin locally via Postman. Do you think we need a unit test for the scenario of hitting a login-required route w/o the appropriate session header?

jworcestBAH:
That's a good idea, I'll add a unit test for that.

jworcestBAH:
New tests added.  I also included one after logging out to ensure that logout cancels the session appropriately.

dataactcore/scripts/databaseSetup.py
bsweger: I was wrong about this--was thinking of something else!

bsweger:
@jworcestBAH  @nmonga91 @afrasier I reviewed this on Friday, and the code looks good. Had a unittest issue that @jworcestBAH was looking into. Once that is resolved or ruled out as a problem, this branch is ready to merge.

bsweger:
Unit test is all fixed up. Merging!

PR #4, opened 2016-05-02
hotfix: add missing job_status constraints to db migrations

bsweger:
The cascade delete on the job_tracker submission id FK was not captured in the alembic db migrations. This script adds those.


dataactcore/migrations/versions/87d17514eeed_job_status_constraints.py
```diff
@@ -0,0 +1,78 @@
...
+def upgrade_job_tracker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(u'job_status_submission_id_fkey', 'job_status', type_='foreignkey')
+    op.create_foreign_key(None, 'job_status', 'submission', ['submission_id'], ['submission_id'], ondelete='CASCADE')
+    ### end Alembic commands ###
+
+
+def downgrade_job_tracker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'job_status', type_='foreignkey')
```
jworcestBAH: I removed a similar line from an earlier migration because it causes the error "Can't emit DROP CONSTRAINT for ForeignKeyConstraint...None...it has no name".  I suspect that "None" is causing problems, the earlier version had the same thing.  Try upgrading to head and then downgrading past this version to see if you get the same error.  That said, dropping this particular constraint seems unnecessary as it's certainly not a bad thing for an earlier version to be cascading on delete, so we could just pass on the downgrade.

bsweger:
@jworcestBAH thanks for pointing that out, and it makes sense b/c the constraint isn't explicitly named in the models file. will update the model, re-generate the migration and update this branch if all works as expected.

bsweger:
@jworcestBAH ok, updating the job_status model to use an explicit constraint name for the foreign key, seems to generate an alembic migration file that will downgrade without error. thanks for the catch!

jworcestBAH:
Migrations now work correctly, merging.

PR #5, opened 2016-05-02
Dynamo status check

nmonga91:
@jworcestBAH or @bsweger


bsweger:
@nmonga91 Thanks so much for finding a solution that doesn't require another install by local users! quick question since I've not used the multiprocessing module before--will this work as intended on machines that only have a single processor? Do such machines exist these days?

nmonga91:
@bsweger Most single processor computers still have multiple cores which allows for true multithreading. All this is doing code wise is starting an separate thread to test the dynamo response. I don't foresee a problem with any hardware. You can still invoke multithreading (they would just be different processes) on a single processor/single core machine.

PR #7, opened 2016-05-03
adding default for rule_timing_id in MultiFieldRule

nmonga91:


jworcestBAH:
This change is fine, but don't we also need a change to the existing migration script?

nmonga91:
You're right, let me make that change

jworcestBAH:
Looks good and seems to work, merging

PR #8, opened 2016-05-03
Pointing to new file paths for supervisor and uwsgi

jworcestBAH:
This fixes the location of the code for the new repo structure in the supervisor and uwsgi config files


nmonga91:
Paths look good

PR #10, opened 2016-05-03
Adding default value to correct error with rule_timing migration

klrBAH:
Hotfix into staging to fix migration issue.


jworcestBAH:
Looks good

PR #12, opened 2016-05-05
Db 449 incorrect password

nmonga91:
@jworcestBAH or @bsweger


tests/loginTests.py
```diff
@@ -29,6 +29,17 @@ def test_inactive_login(self):
         response = self.login_inactive_user()
         self.assertIn("401 UNAUTHORIZED", response.status)
 
+    def test_expired_locked_login(self):
+        """Test broker expired lockout"""
+        response = self.login_expired_locked_user()
+        self.assertIn("expired", response.json["message"].lower())
```
jworcestBAH: Can we also check that the status is 401 here? Same thing in the following test.

nmonga91: Yes, I'll add those in

jworcestBAH:
Looks good, merging

PR #13, opened 2016-05-09
Db 458 rss update

jworcestBAH:
This includes the update to the released version of the RSS, a new rule type for requiring one of a set of fields, and a file documenting the current state of RSS rules.


bsweger:
@jworcestBAH finished reviewing, and all the code/tests look good. Only one question...do you think we should publish the rule documentation file as a .csv instead of an .xlsx so it's in a non-proprietary format (and so it renders on github, is easier to track changes)?

jworcestBAH:
Changed rule doc to csv

bsweger:
Thanks for switching to .csv!

PR #14, opened 2016-05-09
updating error messages for incorrect password attempts, and case for generic invalid login method.

nmonga91:
@bsweger or @jworcestBAH


bsweger:
Looks good--merging!

PR #15, opened 2016-05-09
Had a second zip file of test files, and at least one was out of date

jworcestBAH:
Just updating the zip of sample files, had a duplicate file and it was out of date.


bsweger:
Thanks for catching this!

PR #16, opened 2016-05-10
fixes to ensure Alembic downgrades function correctly

bsweger:
As part of the db simplification tech debt story (DB-376), make sure Alembic downgrades work. There were a few more recurrences of the scenario where a downgrade fails because the constraint it's trying to drop doesn't have a name.


jworcestBAH:
Looks good, merging

PR #17, opened 2016-05-10
Db 370 - allow pipe and comma as CSV delimiters

nmonga91:
@bsweger or @jworcestBAH


bsweger:
@nmonga91 Finished reviewing. I really like the simple way we're inferring the appropriate delimiter on the users' behalf. Just a few questions:
1. Do we need a test case for scenario of mixing commas and pipes in the header row?
2. Are all of our valid test files pipe-delimited now? If so, should we leave at least one as comma-delimited so both are tested?
3. After everything is done, will we need a new copy of test_files.zip in the repo for the testing team (and for local installers?)

jworcestBAH:
Agreed, if we could revert one of the valid files to comma-delimited and make one of the mixed files pipe-delimited that would give us a good spread for testing.

nmonga91:
@bsweger my responses:
1. I agree, I'll include a test for mixed delimiters. 
2. I can revert one of them to comma delimited
3. Yes, i believe so since technically we have new test files with new delimiters

I incorporate the above with @jworcestBAH's comment as well for a mixed file to be pipe-delimited

jworcestBAH:
Mostly looks good, but I have two questions about the mixed delimiter error.  First, it doesn't appear to be generating an error report, which it needs to do for the user to see the error in a useful way.  Also, why is the error type showing as a 'single_row_error', that doesn't really seem to fit.

klrBAH:

nmonga91:
Code has been updated per @jworcestBAH's comments

jworcestBAH:
Now works and generates an error report with reasonable message, merging.

PR #18, opened 2016-05-10
Db 447 rabbit doc

jworcestBAH:
This includes the documentation for RabbitMQ and updates to the documentation for the windows installation process, which requires a few extra steps and workarounds for two errors.  This is by no means ideal, but does at least get us to the point of having a working process for a windows install.


doc/CONTRIBUTING.md
```diff
@@ -106,6 +108,13 @@ Don't worry about setting DynamoDB endpoints or creating tables: the broker's co
 
 **Note:** The local version of DynamoDB is not recommended for production.
 
+### Install RabbitMQ
+
+RabbitMQ is used to pass jobs to the validator, and requires Erlang to be installed before RabbitMQ.
+
+1.  Install Erlang based on the [download instructions](https://www.erlang.org/downloads)
+2.  Choose an installation guide based on your OS for [RabbitMQ](https://www.rabbitmq.com/download.html).  Be sure to install Erlang before installing RabbitMQ.  The default user and password is "guest"/"guest", be sure to update the config.yml file if you change this.
```
bsweger: Since this is before the instructions for setting up the config file, what about something like "By default, the RabbitMQ username and password are both set to `guest.` If you change either of these, make a note of it; you'll need that information when setting up the broker's configuration file."

doc/CONTRIBUTING.md
```diff
@@ -106,6 +108,13 @@ Don't worry about setting DynamoDB endpoints or creating tables: the broker's co
 
 **Note:** The local version of DynamoDB is not recommended for production.
 
+### Install RabbitMQ
+
+RabbitMQ is used to pass jobs to the validator, and requires Erlang to be installed before RabbitMQ.
```
bsweger: Do we need to put a note here that this won't work on anything higher than Windows 7? Or should that note go somewhere closer to the top of the install directions, so those folks don't get halfway through only to be foiled by RabbitMQ?

doc/CONTRIBUTING.md
bsweger: Should we set the RabbitMQ username/pwd in the sample config file to guest/guest, since we're telling folks they'll only have to make a change if they update these values when installing?

bsweger: It might also be a good idea to add the `job-queue` section of config.yml as a place that users should pay attention to (these "special attention" sections are listed under the _Create Broker Config File_ section.

bsweger: Can we add clarification (or another step) to make it very explicit that users will have to not only install RabbitMQ but also run it?

bsweger:
@jworcestBAH Done reviewing. Thanks for making these docs more Windows-friendly! I added some comments/questions to the docs. The thing is, after doing all this, I *still* can't get the test_file_submission unit test to run with the local RabbitMQ. If it's working for you, I'll follow up with @nmonga91 to debug.

doc/CONTRIBUTING.md
jworcestBAH: It does work on Windows 10, and for Windows 8 we have one failure and one success.  The statement from Pivotal was "We've seen this reported on Windows a few times but don't know how to reproduce."  So we don't know how widespread it is and we don't know that it is specific to Windows 8.  Broader testing would be helpful, not sure what the process for that would be.

doc/CONTRIBUTING.md
jworcestBAH: Changed the example values to guest/guest, and updated the text referring to the config file.  In the typical case the user will not be changing those and so doesn't need to pay attention to it in the config file (it's actually complicated to change them).  RabbitMQ actually runs immediately after installation, no extra steps are needed.  Stopping and restarting the service can be more complicated, the information is available in the install guide we link to if they want to do that, but the typical user just needs to install and move to the next step.

jworcestBAH:
So I can't do a local rabbit queue on the laptop where I have all the code.  I could set everything up on my home computer where rabbitMQ does work, but that would take some time.

klrBAH:

jworcestBAH:
Katie I'm not sure what you're referring to.  Was this comment supposed to be on a different PR?

bsweger:
Changes look good. We can iterate on this as we hear feedback from local installers.

PR #19, opened 2016-05-10
hotfix for overriding job db name

bsweger:
@nmonga91 Can you take a look at this, since you're the most familiar with the job queue? Didn't catch this before, but it turns out that the changes introduced for the job queue break the overriding of the database name (for instance, when unit tests are running). This didn't come to light until I was changing the database names as part of another story.

Will this updated code work with the queue logic?


nmonga91:
Everything looks good!

PR #21, opened 2016-05-10
DB 376: Standardize Database Table Names

bsweger:
This PR standardizes some older table names in the error and job tracker databases:
 
**Error Data**
* `file_status` --> `file`
* `status` --> `file_status`
* `error_data` --> `error_metadata`

**Job Tracker**
* `job_status` --> `job`
* `status` --> `job_status`
* `type` --> `job_type`

We had also discussed minor changes to a few tables in the user db (`users` and `permission_type`). I left those out of this PR due to time, because they weren't explicitly in the story discussion, and because they are less egregious in general.

@jworcestBAH @nmonga91 This is ready for review. However, I'd prefer to see if we can run @kevinli-work's front-end unit tests against this branch before merging, since the changes are pretty extensive.


dataactbroker/handlers/jobHandler.py
```diff
@@ -14,13 +14,13 @@ class JobHandler(JobTrackerInterface):
...
     connection -- sqlalchemy connection for executing direct SQL statements
     session -- sqlalchemy session for ORM usage
-    waitingStatus -- status_id for "waiting"
-    runningStatus -- status_id for "running"
-    finishedStatus -- status_id for "finished"
-    fileUploadType -- type_id for "file_upload"
-    dbUploadType -- type_id for "db_upload"
-    validationType -- type_id for "validation"
-    externalValidationType -- type_id for "external_validation"
+    waitingStatus -- job_status_id for "waiting"
```
jworcestBAH: So most of these no longer exist as instance variables, I'll push a commit to remove these lines.

klrBAH:

dataactcore/migrations/versions/19c18c1545c3_file_status_job_unique.py
```diff
@@ -1,4 +1,4 @@
-"""file_status_job_unique
+"""file_job_unique
```
jworcestBAH: I'm assuming nothing breaks with this no longer matching the file name?  I'll test alembic migrations later to make sure.

jworcestBAH:
Broader question, what should our process be for migrating DBs to this?  I see you've made the changes throughout past migration scripts rather than creating a new migration, so it seems like we can't apply alembic to current DBs.  Do we need to completely reset the databases to do this?

bsweger:
@jworcestBAH That's a good question, and something I should have asked the team. As I was developing, I made the assumption that foundational changes like re-naming tables would be best served by completely resetting the impacted databases (error and job), but obviously that could be problematic since we're live. Thanks for raising this point--let's discuss in the channel.

bsweger:
Ok, based on our convo at sprint review, sounds like we're okay this time to drop `error_data` and `job_tracker` as we cut over to the new table names. I recognize that this is a deploy nuisance for @klrBAH. Sorry for the trouble--hopefully we won't be changing any other table names for a while.

bsweger:
Also, will fix up the merge conflicts.

dataactcore/scripts/resetAlembicVersion.py
```diff
@@ -0,0 +1,36 @@
...
+        session.execute(u)
+        session.commit()
+
+parser = argparse.ArgumentParser\
+    (description="Reset alembic version tables across broker databases.")
+parser.add_argument(
+    'version', help="Version to set the Alembic migration tables to.")
+v = vars(parser.parse_args())['version']
+reset_alembic(v)
+
```
jworcestBAH: This is a nice script to have, thanks for adding this.

jworcestBAH:
Looks good, merging

PR #33, opened 2016-05-13
Docstring update

jworcestBAH:
Resubmitting the docstring PR, this time to the correct target.


dataactbroker/README.md
```diff
@@ -555,14 +555,12 @@ Example output:
 
 ## Test Cases
 
-To run the broker API unit tests, navigate to the main project folder (`data-act-broker`) and type the following:
+To run the broker API unit tests, navigate to the main project folder (`data-act-broker-backend`) and type the following:
```
bsweger: Good catch! This made me realize something else...the runTests.py command will not run both the broker and validator tests, which is probably what we want. However, the validator tests won't run unless you're actually in the tests folder (we should probably fix that in a tech debt task at some point). I'll push a change that updates these directions accordingly.

jworcestBAH: Could we just change these directions to instruct people to be in the tests folder rather than the main project folder?  Is there a benefit to running them from there?

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -19,10 +19,12 @@ class CsvAbstractReader(object):
     def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
         """ Opens file and prepares to read each record, mapping entries to specified column names
         Args:
-            bucket : the S3 Bucket
+            region: AWS region where the bucket is located
```
bsweger: Is this method only intended to non-local use. If so, should we say so here (I noticed you had the local/non-local distinction in a few other docstrings, which is very helpful).

bsweger:
@jworcestBAH Done reviewing. Thanks so, so much for taking the time to update everything. Now that we're current, we can make an extra effort to spot docstring inconsistencies in code review so we don't have to do a big cleanup again.

dataactvalidator/filestreaming/csvAbstractReader.py
jworcestBAH: It looks like those values are passed to getWriter, which only uses them if self.isLocal is false.  This function is overloaded by CsvLocalReader.openFile and CsvS3Reader.openFile, which therefore need to have the same arguments,  Those arguments are simply ignored if the object is a CsvLocalReader rather than a CsvS3Reader.  I'll add a note to these arguments that they are not used if running locally.

bsweger:
Thanks for the clarifications! Merging...

PR #36, opened 2016-05-16
Db 547 staging field names

jworcestBAH:
Staging database now records mappings of column IDs to field names.  As this involved creating a model object for the staging database, it has now been added to alembic as well.


dataactvalidator/scripts/initialize.py
```diff
@@ -62,7 +61,6 @@ def loadValidator():
 
 def setupDB():
     setupValidationDB()
-    setupStagingDB()
```
bsweger: Where will the staging db be created now? I've been setting up the app by [running the broker and validator initialize.py scripts](https://github.com/fedspendingtransparency/data-act-broker-backend/blob/db-547-staging-field-names/doc/CONTRIBUTING.md#initialize-broker-backend-applications), and when I do that in this branch, no staging db is created.

bsweger:
@jworcestBAH Done reviewing. My main question is to ask about the strategy of storing the field name mapping info in a separate staging table instead of as a column in staging job table itself, since it's just a single column. Is it because we need a permanent record of that information? Do we expect to store other information in the new staging table at some point as a way to pass info to the datastore?

dataactvalidator/scripts/initialize.py
jworcestBAH: It is called in setupAllDB, I had been under the impression that was called elsewhere.  Looking at the broker initialize, it seems that lists the DBs to be set up individually, so I've re-added the staging call here.

jworcestBAH:
The issue is that it's a single entry per table.  If it were another column within the table, we'd have to store that map separately for each record, or have a blank record to specifically have the data in that one column and nothing else populated, which to me seems like a rather awkward solution.  That said, the structure of the staging database is going to be rethought after the meeting next week with Timothy, so it may get reshuffled as part of that.

bsweger:
@jworcestBAH Ah yes, of course. I attribute my question above to a caffeine deficit. Looks good!

PR #38, opened 2016-05-17
Db 497 consolidate setup

jworcestBAH:
This branch consolidates the db setup for validation, staging, and job queue to be consistent with other DBs, and refactors initialize scripts to work correctly and include a -d option for deploying on AWS.


dataactvalidator/filestreaming/schemaLoader.py
```diff
@@ -81,16 +84,19 @@ def loadCrossRules(filename):
                 fileId = validationDb.getFileId(record["file"])
                 validationDb.addMultiFieldRule(fileId,record["rule_type"],record["rule_text_one"],record["rule_text_two"],record["description"],record["rule_label"],record["rule_timing"])
 
+    @classmethod
```
bsweger: I really like this revamp of the field and rule file loads. Do you think it makes sense to take it one step further and add top-level path of the broker backend and the path to the validator conifg in `dataactcore/config.py` so we don't have to worry about constructing that path down here?

bsweger:
@jworcestBAH done reviewing!

dataactvalidator/filestreaming/schemaLoader.py
jworcestBAH: CONFIG_BROKER['path'] now holds the path to the base directory for the install.

jworcestBAH:
Updated path handling in config.py to use abspath(__file__) throughout

bsweger:
Looks good--thanks for adding the broker path to the config file. That will come in handy!

PR #39, opened 2016-05-17
Updated rule doc

jworcestBAH:
Updating the rule documentation based on discussion with schema team.  @bsweger:  See the comments on rules A18 and A19 for your current story.


bsweger:
@jworcestBAH looks good--thanks for keeping this up to date and for documenting the clarifications!

PR #40, opened 2016-05-17
Not including 'u' for unicode in the map put in the staging table

jworcestBAH:
Small hotfix to the staging table change to not include 'u' in front of every field name.

### DIFF
--- a/dataactvalidator/interfaces/stagingTable.py
+++ b/dataactvalidator/interfaces/stagingTable.py
@@ -83,7 +83,7 @@ class StagingTable(object):
             else:
                 classFieldDict[newKey] = Column(fieldTypeName)
             # First record will hold field names
-            fieldNameMap[newKey] = key
+            fieldNameMap[str(newKey)] = str(key)
         # Add column for row number
         classFieldDict["row"] = Column(Integer, nullable=False)
 


dataactvalidator/interfaces/stagingTable.py
```diff
@@ -83,7 +83,7 @@ def createTable(self, fileType, filename, jobId, tableName=None):
             else:
                 classFieldDict[newKey] = Column(fieldTypeName)
             # First record will hold field names
-            fieldNameMap[newKey] = key
+            fieldNameMap[str(newKey)] = str(key)
```
bsweger: In Python 2.7, I think this would break if there actually was a non-ascii character in `newKey` or `key`. That said, this is a pattern we have all over the code base, so I vote to leave it as a further incentive to update to Python 3.

bsweger:
Merging this one. It made me realize that we have a lot of `str(something)` in the code base which could prove to be problematic before we upgrade to Python 3.

jworcestBAH:
@bsweger:  Google seems to think casting to str this way is still a thing in Python 3, what makes this problematic?

bsweger:
Oh, it's great for Python 3. Should said string have any non-ascii characters in Python 2, I believe this will break.

PR #42, opened 2016-05-18
Db 33 - Adding agency admin specific endpoints and updates

nmonga91:
@bsweger or @jworcestBAH


dataactbroker/handlers/accountHandler.py
```diff
@@ -377,13 +392,33 @@ def listUsersWithStatus(self):
...
+        return JsonResponse.create(StatusCode.OK, {"recent": submissionDetails})
+
     def listSubmissionsByCurrentUser(self):
         """ List all submission IDs associated with the current user ID """
         userId = LoginSession.getName(flaskSession)
         submissions = self.interfaces.jobDb.getSubmissionsByUserId(userId)
         submissionIdList = []
         for submission in submissions:
             submissionIdList.append(submission.submission_id)
+        print "DEBUG: Submission Count => " + str(len(submissionIdList))
```
bsweger: whoops, leftover DEBUG :)

dataactbroker/handlers/accountHandler.py
```diff
@@ -479,8 +514,8 @@ def isAccountExpired(self, user):
         """
         today = parse(time.strftime("%c"))
         daysActive = (today-user.last_login_date).days
-        secondsActive = (today-user.last_login_date).seconds
-        if daysActive > self.INACTIVITY_THRESHOLD or (daysActive == self.INACTIVITY_THRESHOLD and secondsActive > 0):
+        # secondsActive = (today-user.last_login_date).seconds
```
bsweger: We should probably remove this comment so future team members don't get confused by the glory days of when we tracked active accounts by the second.

dataactbroker/handlers/fileHandler.py
```diff
@@ -108,6 +109,9 @@ def submit(self,name,CreateCredentials):
                 # If filetype not included in request, and this is an update to an existing submission, skip it
                 if not safeDictionary.exists(fileType):
                     if existingSubmission:
+                        submission = self.jobManager.getSubmissionById(submissionId)
```
bsweger: Great idea to make sure the submission table timestamp is reset for a submission update. What do you think about doing that outside of the filetype loop to prevent potential multiple submission table updates?

dataactbroker/handlers/jobHandler.py
```diff
@@ -24,6 +24,10 @@ def getSubmissionById(self,submissionId):
         result = self.runUniqueQuery(query,"No submission with that ID","Multiple submissions with that ID")
         return result
 
+    def getSubmissionsByUserAgency(self,user,limit=5):
```
bsweger: I honestly don't know the best practice here, so just throwing out this question...if the front-end is currently wireframed to show 5 submissions, is it better to send back exactly 5 and let the web app make additional calls if/when we see a pagination story, or is it better to send back more than 5 and let the web app handle pruning down the list to 5?

dataactbroker/permissions.py
```diff
@@ -20,6 +20,10 @@ def decorated_function(*args, **kwargs):
                     for permission in permissionList :
                         if(not userDb.hasPermission(user, permission)) :
                             validUser = False
+                        else:
```
bsweger: If I read this correctly, it looks like this code changes the way this decorator works when we're checking for multiple permissions....from "user needs to have all permissions in the list to be considered valid" to "user is valid if they have at least one permission in the list." Is this the intended effect?

dataactbroker/handlers/accountHandler.py
```diff
@@ -2,6 +2,7 @@
 from threading import Thread
 import re
 import time
+from flask import session
```
bsweger: Is this the same info as line 1 (`from flask import session as flaskSession`)? Reading through the code, it seemed like `session` and `flaskSession` are one in the same. If that's true, I think it would be better to stick with one to avoid future confusion.

dataactbroker/handlers/accountHandler.py
```diff
@@ -172,10 +173,18 @@ def ThreadedFunction (from_email="",username="",title="",agency="",userEmail=""
             """
             threadedDatabase =  UserHandler()
             try:
-                for user in threadedDatabase.getUsersByType("website_admin") :
+                for user in threadedDatabase.getUsersByType("website_admin"):
```
bsweger: 👍 for getting rid of extraneous spaces!

dataactcore/models/jobTrackerInterface.py
```diff
@@ -6,7 +6,8 @@
 from dataactcore.utils.responseException import ResponseException
 from dataactcore.utils.cloudLogger import CloudLogger
 from dataactcore.utils.jobQueue import JobQueue
-from dataactcore.config import CONFIG_DB
+from dataactcore.config import CONFIG_DB, CONFIG_SERVICES
```
bsweger: It doesn't look like CONFIG_SERVICES is being used.

bsweger:
@nmonga91 finished reviewing--apologies for taking so long. My overall read is that this syncs well with the acceptance criteria in DB-33. Are you planning to submit follow-up PRs with tests and updates for the API route documentation?

dataactbroker/handlers/fileHandler.py
nmonga91: @bsweger I did that because that loop has a check if this is a valid file type before checking for the existing submission. Just made sense to put that update right before the continue within that if block.

dataactbroker/handlers/jobHandler.py
nmonga91: @bsweger Either way would work just fine. I added the limit just as a parameter in anticipation of a pagination story where they could enter a specific number or get all of them at once. We could make the web app filter itself. I'm good with either way, it's a simple update in the backend. What would you suggest?

dataactbroker/permissions.py
nmonga91: @bsweger Yes you are correct. The user should have at least one permission in the list to utilize the route.

dataactbroker/handlers/accountHandler.py
nmonga91: You're right. That needs to be removed.

dataactcore/models/jobTrackerInterface.py
nmonga91: That will be removed!

dataactbroker/handlers/fileHandler.py
bsweger: Got -- I see the nuance now. Thanks for clarifying! 👍

dataactbroker/handlers/jobHandler.py
bsweger: I don't feel strongly either way....just wanted to check-in and see if there's a preferred method. I think your anticipation of a future pagination story makes sense, since the submission volume will pile up over time.

nmonga91:
@bsweger Unit test was updated and API Route docs will be updated with the next PR for DB-502

dataactbroker/handlers/accountHandler.py
```diff
@@ -408,7 +408,7 @@ def listSubmissionsByCurrentUserAgency(self):
             error_count = self.interfaces.errorDb.sumNumberOfErrorsForJobList(jobIds)
             submissionDetails.append({"submission_id": submission.submission_id, "last_modified": submission.updated_at.strftime('%m/%d/%Y'),
                                       "size": total_size, "status": status, "error": error_count})
-        return JsonResponse.create(StatusCode.OK, {"recent": submissionDetails})
+        return JsonResponse.create(StatusCode.OK, {"submissions": submissionDetails})
```
bsweger: Since the doc updates are coming in a future PR, just checking to make sure the front-end team is cool with this response (I agree that _submissions_ is much better than _recent_).

dataactbroker/handlers/userHandler.py
```diff
@@ -390,7 +390,7 @@ def getPermssionList(self):
         queryResult = self.session.query(PermissionType).all()
         return queryResult
 
-    def createUserWithPassword(self,email,password,bcrypt,admin=False):
+    def createUserWithPassword(self,email,password,bcrypt,permission=1):
```
bsweger: Does the value of `permission` represent `permission_type_id` in the permission_type_table? I'm still a bit unclear on how the broker sets permissions, but this smells like a magic number. We have a few of those in the code base already (e.g., in unit tests), but we should avoid introducing more.

dataactbroker/scripts/initialize.py
```diff
@@ -60,7 +60,7 @@ def createAdmin():
 
         if type(e.wrappedException) is NoResultFound:
             userDb.createUserWithPassword(
-                adminEmail, adminPass, Bcrypt(), admin=True)
+                adminEmail, adminPass, Bcrypt(), permission=2)
```
bsweger: See above comment re: magic numbers.

dataactbroker/handlers/userHandler.py
jworcestBAH: This can be avoided by calling UserHandler.getPermissionId("agency_user").  The one below would be "website_admin".  If desired, you could refactor this function to take a list of permission strings rather than a single number.  Alternatively, just set the default value here to None, and then in the code:
`if permission is None:
    permission = self.getPermissionId("agency_user")`

jworcestBAH: Actually you would need to do 2^getPermissionId("agency_user") to do it this way, in general it's preferred to use grantPermission and removePermission to avoid this.

tests/userTests.py
```diff
@@ -158,12 +166,24 @@ def test_list_submissions(self):
...
+        self.login_agency_user()
+        response = self.app.get("/v1/list_submissions/", headers={"x-session-id": self.session_id})
+        self.check_response(response, StatusCode.OK)
+        self.assertIn("submission_id_list", response.json)
+        self.assertEqual(len(response.json["submission_id_list"]), 6)
+
+        response = self.app.get("/v1/list_submissions/?filter_by=agency", headers={"x-session-id": self.session_id})
+        self.check_response(response, StatusCode.OK)
+        self.assertIn("submissions", response.json)
+        self.assertEqual(len(response.json["submissions"]), 5)
```
bsweger: @nmonga91 Sorry for not noticing this yesterday, but I'm curious about why the format of the submission info being returned varies based on whether or not the  `filter_by_agency` is used? Would it make more sense for the route to return the submission info consistently?

bsweger:
@nmonga91 Ok, finished reviewing the new changes. Made a note on [DB-502](https://federal-spending-transparency.atlassian.net/browse/DB-502) that we'll address the permission setting there.

My lingering question (and apologies for not asking sooner) is about the different formats for the submission data being returned when the new param is used.

bsweger:
I like how @klrBAH's monster avatar appears when the tests fail

jworcestBAH:
That was very confusing when it started out with Jenkins sending an automated message "Can an admin please check this?"

bsweger:
We're ready to stick a fork in this one. Logged a [JIRA task](https://federal-spending-transparency.atlassian.net/browse/DB-566) to clean up some of the duplicate code the next time we're in `accountHandler.py`

PR #43, opened 2016-05-20
fix bug in alembic metadata setup

bsweger:
This PR fixes a bug that was, in some cases, preventing data model metadata from being passed to the Alembic context. This problem was only occurring in cases where the database names in the broker's config.yml don't match the original database names that were hard-coded into the app back in the day.


jworcestBAH:
This is definitely better than dynamically changing all the function names in every migration to match the names in the config, merging.

PR #44, opened 2016-05-20
Added skip_guide parameter and handling of it

jworcestBAH:
This PR adds the skip_guide parameter which users can set to skip viewing the submission guide on subsequent visits, and adds a route to set it.


bsweger:
This looks good...thanks for updating the docs as part of the PR!

PR #45, opened 2016-05-20
Db 493 reset error

jworcestBAH:
This checks that users are both approved and active before allowing them to reset passwords.


klrBAH:
[run tests]

bsweger:
Looks good--merging!

PR #46, opened 2016-05-20
Db 511 calculations

jworcestBAH:
This includes all record level calculation rules for files A, B, and C, as well as a new type of conditional requirement rule for file C.


dataactvalidator/validation_handlers/validator.py
```diff
@@ -421,17 +424,39 @@ def evaluateMultiFieldRule(rule, record, interfaces, fileType):
...
+            return not cls.evaluateMultiFieldRule(conditionalRule,record,interfaces,fileType)
+        elif(ruleType == "REQUIRED_CONDITIONAL"):
+            # Field in rule_text_1 is required if rule in rule_text_2 is satisfied
+            conditionalRule = interfaces.validationDb.getMultiFieldRuleByLabel(rule.rule_text_2)
+            if cls.evaluateMultiFieldRule(conditionalRule,record,interfaces,fileType):
+                # Return True if field populated
+                field = FieldCleaner.cleanName(rule.rule_text_1)
+                if field in record and cls.isFieldPopulated(record[field]):
+                    return True
+                else:
```
nmonga91: The else clause for the default return of "False" isn't necessary. You can just have it do the if block then return False outside the if afterwards.
if x:
    return True
return False

dataactvalidator/validation_handlers/validator.py
```diff
@@ -421,17 +424,39 @@ def evaluateMultiFieldRule(rule, record, interfaces, fileType):
...
+                    return False
+            else:
```
nmonga91: Same as above comment

dataactvalidator/validation_handlers/validator.py
```diff
@@ -449,29 +474,35 @@ def cleanSplit(string, toLower = True):
...
-
+            interfaces: InterfaceHolder for DBs
         Returns:
             One string including all values involved in this rule check
         """
         ruletext1 = Validator.cleanSplit(rule.rule_text_1, True)
         ruletext2 = Validator.cleanSplit(rule.rule_text_2, True)
-        fields = ruletext1 + ruletext2 #TODO try catch
+        ruleType = interfaces.validationDb.getMultiFieldRuleTypeById(rule.multi_field_rule_type_id)
+        if ruleType == "CAR_MATCH" or ruleType == "REQUIRED_CONDITIONAL" or ruleType == "REQUIRE_ONE_OF_SET":
```
nmonga91: We should probably come up with a better way do things like get the fields for a specific rule type here instead of hard-coding the if blocks. This is obviously a future enhancement, but this could be checked against the database and if the ruleType doesn't exist there, then throw an exception otherwise utilize the table to get what the fields variable should correspond to.

dataactcore/models/validationModels.py
```diff
@@ -31,7 +31,6 @@ class RuleType(Base):
     description = Column(Text)
 
     TYPE_DICT = None
-    TYPE_LIST = ["TYPE", "EQUAL","NOT EQUAL","LESS","GREATER","LENGTH","IN_SET","SUM"]
```
nmonga91: Why is this set to None now vs the original array it contained?

dataactcore/models/validationModels.py
```diff
@@ -42,7 +41,6 @@ class MultiFieldRuleType(Base):
 
     session = None
     TYPE_DICT = None
-    TYPE_LIST = ["CAR_MATCH", "SUM_TO_VALUE"]
```
nmonga91: Same question as above comment

dataactvalidator/validation_handlers/validator.py
jworcestBAH: I agree these aren't necessary but visually it feels cleaner this way to me, makes it a bit more clear what the behavior is.

jworcestBAH: Also if the behavior in the if clause ever changed to not immediately return it would not break the behavior of returning False if the field is not present.

dataactcore/models/validationModels.py
jworcestBAH: These were not being used and the information they held was obsolete anyway.  We cleaned up some of these a while ago, but these two were missed, so I'm removing them here.  Just found another one in JobType, I'm pushing a change to delete that as well.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Hmm, this could be added as an extra column to the rule_type table that defines what fields to show values for.  I'll see if I can include that in my current refactor of the Rule tables

nmonga91:
Review completed. Looks good, merging.

PR #48, opened 2016-05-23
Db 502 user management

nmonga91:
@bsweger or @jworcestBAH


dataactbroker/README.md
```diff
@@ -382,6 +406,27 @@ Example output:
...
+}
+```
+
+Example output:
+
+```json
+{
+    "message": "User successfully updated"
+}
+```
```
jworcestBAH: Has this change been coordinated with Kevin to not break the existing admin page that uses change_status?

dataactbroker/handlers/accountHandler.py
```diff
@@ -113,8 +113,9 @@ def login(self,session):
                 LoginSession.logout(session)
                 raise ve
             except Exception as e:
-                    LoginSession.logout(session)
-                    raise ValueError("Invalid username and/or password")
+                print e
```
jworcestBAH: Was this for debugging?

dataactbroker/handlers/accountHandler.py
```diff
@@ -324,6 +325,71 @@ def checkPasswordToken(self,session):
...
+                    emailTemplate = {}
+                    newEmail = sesEmail(user.email, system_email,templateType="account_rejected",parameters=emailTemplate,database=self.interfaces.userDb)
+                    newEmail.send()
+            # Change user's status
+            self.interfaces.userDb.changeStatus(user,requestDict.getValue("status"))
+
+        if requestDict.exists("permissions"):
+            permissions_list = requestDict.getValue("permissions").split(',')
+
+            # Remove all existing permissions for user
```
jworcestBAH: Are we sure that the way of changing permissions should remove all other permissions?  Seems like we'd want a way to add or remove a permission without wiping everything?

dataactbroker/handlers/accountHandler.py
```diff
@@ -366,6 +432,27 @@ def changeStatus(self,system_email):
...
 
+    def listUsers(self):
+        """ List all users ordered by status. Associated request body must have key 'filter_by' """
+        user = self.interfaces.userDb.getUserByUID(LoginSession.getName(flaskSession))
+        isAgencyAdmin = True if self.interfaces.userDb.hasPermission(user, "agency_admin") else False
+        try:
+            if isAgencyAdmin:
+                users = self.interfaces.userDb.getUsers(agency=user.agency)
+            else:
+                users = self.interfaces.userDb.getUsers()
```
jworcestBAH: I'm confused by the purpose here.  What is the set of users someone who is not an admin is allowed to see?  If this is empty or just themselves, shouldn't we just require agency_admin in order to access the route?

dataactbroker/handlers/userHandler.py
```diff
@@ -58,6 +58,13 @@ def deleteToken(self,token):
         self.session.delete(oldToken)
         self.session.commit()
 
+    def getUsers(self, agency=None):
+        """ Return all users in the database """
+        query = self.session.query(User)
+        if agency is not None:
+            return query.filter(User.agency == agency)
+        return query.order_by(User.user_status_id).all()
```
jworcestBAH: This feels like they should be two separate functions, these don't really have any code in common, and the purpose is rather different.

dataactbroker/handlers/accountHandler.py
jworcestBAH: Assuming the second case is for the /v1/current_user/, are agency admins not allowed to use the current_user route?

dataactbroker/handlers/accountHandler.py
```diff
@@ -366,6 +432,27 @@ def changeStatus(self,system_email):
...
+        except ValueError as e:
+            # Client provided a bad status
+            exc = ResponseException(str(e),StatusCode.CLIENT_ERROR,ValueError)
+            return JsonResponse.error(exc,exc.status)
+        userInfo = []
+        for user in users:
+            thisInfo = {"name":user.name, "title":user.title,  "agency":user.agency, "email":user.email, "id":user.user_id,
+                        "is_active":user.is_active, "permissions": ", ".join(self.interfaces.userDb.getUserPermissions(user)),
+                        "status": user.user_status.name}
```
jworcestBAH: We could pull this part out into a separate function and have the current_user route call that function directly.  That seems cleaner than putting that route through this function just to get this one line and the getUserByUID line.

jworcestBAH:
Finished reviewing.

dataactbroker/README.md
nmonga91: I've notified Kevin that the change_status route will no longer be used but instead use the update_user route.

dataactbroker/handlers/accountHandler.py
nmonga91: Yes, I will remove the print statement.

dataactbroker/handlers/accountHandler.py
nmonga91: The way this would work is that all selected permissions (from a multi-select input box) would sent back. So if the user is already an "agency_user" it would be selected. If an admin adds "agency_admin" as a permission, then it basically removes "agency_user" and applies both "agency_user" AND "agency_admin". It's simpler to do it this way vs adding logic to see which of the selected permissions they already have and then grant/remove accordingly.

dataactbroker/handlers/accountHandler.py
nmonga91: non-admins can't use the list_users route. The second case is not for the "current_user". The first case handles if the agency admin is calling the route which returns only users in their agency. The second case is for the website admin which returns all users.

dataactbroker/handlers/userHandler.py
nmonga91: the if block should actually contain "query = query.filter(User.agency == agency"). But they code is common the only difference is adding a filter vs returning all. If we were to separate them out into 2 functions, then the only difference is that one is parameterized with a filtery_by option while the other isn't. Is that the desired way to organize these functions?

dataactbroker/handlers/accountHandler.py
nmonga91: No sure what the correlation to the current_user route is. This is a generalization of the "list_users_with_status" route which provides a larger scope of users as the result.

dataactbroker/handlers/accountHandler.py
jworcestBAH: On the backend side it's definitely simpler to just grant or remove specified permissions without wiping the rest.  Granting and removing permissions is idempotent, you don't need to care whether they already have the permission.  That said, if the frontend is showing the current permissions and those get maintained unless the user actively unchecks them, I guess this works.

dataactbroker/handlers/userHandler.py
jworcestBAH: If the line inside changes to query = query.filter(...), suddenly this function makes a lot more sense.

jworcestBAH:
Waiting on this merge until the frontend is ready for it.

jworcestBAH:
Okay, merging.

PR #49, opened 2016-05-23
Db 429 add target file col

bsweger:
This is the first part of the work for DB-429, which calls for adding a new sum_by_tas rule. These commits represent the scaffolding for this story:

* add the new rule type and put one of the new rules into the crossFileRules.csv
* add a new column to hold target file info
* add stub for the `sum_by_tas` rule into the validator
* add unit test data and stub


tests/cross_file_A.csv
```diff
@@ -0,0 +1,4 @@
+AllocationTransferAgencyIdentifier,AgencyIdentifier,BeginningPeriodOfAvailability,EndingPeriodOfAvailability,AvailabilityTypeCode,MainAccountCode,SubAccountCode,BudgetAuthorityUnobligatedBalanceBroughtForward_FYB,AdjustmentsToUnobligatedBalanceBroughtForward_CPE,BudgetAuthorityAppropriatedAmount_CPE,BorrowingAuthorityAmountTotal_CPE,ContractAuthorityAmountTotal_CPE,SpendingAuthorityfromOffsettingCollectionsAmountTotal_CPE,OtherBudgetaryResourcesAmount_CPE,BudgetAuthorityAvailableAmountTotal_CPE,GrossOutlayAmountByTAS_CPE,ObligationsIncurredTotalByTAS_CPE,DeobligationsRecoveriesRefundsByTAS_CPE,UnobligatedBalance_CPE,StatusOfBudgetaryResourcesTotal_CPE
+19,72,0,0,X,306,0,89000,4444444,76555,234434,123,0,333,111,3000000,9000000,33444,30009.12,99999
+28,28,0,0,X,8007,0,89000,4444444,76555,234434,123,0,333,111,24000005.9,1234567,33444,30009.12,99999
+69,21,0,0,X,2065,5,89000,4444444,76555,234434,123,0,333,111,99999,7654321.55,33444,30009.12,99999
```
jworcestBAH: Incidentally, these files are going to run into lots of calculation errors from DB-511.  That said, we can probably just ignore those errors in the unit test for this.

dataactvalidator/interfaces/validatorValidationInterface.py
```diff
@@ -272,13 +272,16 @@ def addMultiFieldRule(self,fileId, ruleTypeText, ruleTextOne, ruleTextTwo, descr
             ruleTextOne: definition of rule
             ruleTextTwo: definition of rule
             description: readable explanation of rule
+            ruleLabel: description of rule
```
jworcestBAH: Thanks for adding these.  Minor point, the rule label doesn't provide any description, it's a string like "A23" that provides a label we can use to refer to that rule.

jworcestBAH:
Done reviewing.

tests/cross_file_A.csv
bsweger: I can remove cross_file_A and cross_file_B from this PR to avoid unit test errors. These aren't necessary for the db changes.

bsweger:
@jworcestBAH Thanks for the review! Pushed some changes as a result of the feedback.

tests/cross_file_A.csv
jworcestBAH: They shouldn't cause errors in the unit test, just in the error report.  I just wanted you to be aware of them because they will prevent the rows from being added to the staging DB, which presumably will break the cross file rules you're implementing.

jworcestBAH:
Looks good, merging.

PR #52, opened 2016-05-24
Code cleanup for list_users response and ordering permissions list in…

nmonga91:
… response alphabetically


klrBAH:
tests passed, looks good.

PR #56, opened 2016-05-25
Dev stable

klrBAH:


bsweger:
Closing the book on another sprint. Woo!

bsweger:
@klrBAH Not sure what happened, but it looked like all checks where complete, but something fired again when I left the comment above? Or when I hit merge?

PR #57, opened 2016-05-26
Changing KeyError catch to AttributeError to handle closing a connect…

nmonga91:
…ion when it doesn't exist.

### DIFF
--- a/dataactcore/models/baseInterface.py
+++ b/dataactcore/models/baseInterface.py
@@ -47,7 +47,7 @@ class BaseInterface(object):
             self.Session.remove()
             self.connection.close()
             self.engine.dispose()
-        except KeyError:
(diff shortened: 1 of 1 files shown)


dataactcore/models/baseInterface.py
```diff
@@ -47,7 +47,7 @@ def __del__(self):
             self.Session.remove()
             self.connection.close()
             self.engine.dispose()
-        except KeyError:
+        except AttributeError:
             # KeyError will occur in Python 3 on engine dispose
             pass
```
jworcestBAH: Actually I meant add AttributeError, not replace, I suspect KeyError can still occur.

jworcestBAH:
And fine here as well.

PR #58, opened 2016-05-26
Changing KeyError catch to AttributeError to handle closing a connect…

nmonga91:
…ion when it doesn't exist.


jworcestBAH:
Okay, looks good now.

PR #59, opened 2016-05-26
Function that filters rules for a field now first checks if file_colu…

jworcestBAH:
…mn exists to avoid an error on multi-field rules


nmonga91:
Looks good

PR #60, opened 2016-05-26
Fixes cross file rule loading's use of common addRule function

jworcestBAH:
File ID was being provided as a field ID when cross file rules were loaded, which caused a foreign key bug


nmonga91:
All tests pass, looks good. Merging.

PR #61, opened 2016-05-26
Adding logic to send email upon account unlock. Updated reset email l…

nmonga91:
…ogic to be separate method to be able to have local method call.

@bsweger or @jworcestBAH

### DIFF
--- a/dataactbroker/handlers/accountHandler.py
+++ b/dataactbroker/handlers/accountHandler.py
@@ -386,6 +386,8 @@ class AccountHandler:
         # Activate/deactivate user
         if requestDict.exists("is_active"):
             is_active = bool(requestDict.getValue("is_active"))
+            if not self.isUserActive(user) and is_active:
+                self.sendResetPasswordEmail(user, system_email, inactive_user_override=True)
             self.interfaces.userDb.setUserActive(user, is_active)
 
         return JsonResponse.create(StatusCode.OK, {"message": "User successfully updated"})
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/accountHandler.py
```diff
@@ -558,23 +560,33 @@ def resetPassword(self,system_email,session):
...
+
+    def sendResetPasswordEmail(self, user, system_email, email=None, inactive_user_override=False):
+        if email is None:
+            email = user.email
+
         # User must be approved and active to reset password
         if user.user_status_id != self.interfaces.userDb.getUserStatusId("approved"):
             raise ResponseException("User must be approved before resetting password", StatusCode.CLIENT_ERROR)
-        elif not user.is_active:
+        elif not inactive_user_override and not user.is_active:
```
nmonga91: Intentionally left it with 2 nots as opposed to doing not (a or b) for readability

dataactbroker/handlers/accountHandler.py
```diff
@@ -558,23 +560,33 @@ def resetPassword(self,system_email,session):
...
             return JsonResponse.error(exc,exc.status)
+
+        email = requestDict.getValue("email")
+        LoginSession.logout(session)
+        self.sendResetPasswordEmail(user, system_email, email)
+
+        # Return success message
+        return JsonResponse.create(StatusCode.OK,{"message":"Password reset"})
+
+    def sendResetPasswordEmail(self, user, system_email, email=None, inactive_user_override=False):
```
jworcestBAH: Minor nitpick, I feel like parts of this function would be clearer if the last flag were named something like unlock_user rather than inactive_user_override, especially where we use it to pick the email template for unlocking a user

jworcestBAH:
Okay, looks good

PR #63, opened 2016-05-27
Resetting fields that cause a user's account to be locked when they a…

nmonga91:
…re being unlocked by an admin

@jworcestBAH or @bsweger

### DIFF
--- a/dataactbroker/handlers/accountHandler.py
+++ b/dataactbroker/handlers/accountHandler.py
@@ -387,6 +387,10 @@ class AccountHandler:
         if requestDict.exists("is_active"):
             is_active = bool(requestDict.getValue("is_active"))
             if not self.isUserActive(user) and is_active:
+                # Reset password count to 0
+                self.resetPasswordCount(user)
+                # Reset last login date so the account isn't expired
+                self.interfaces.userDb.updateLastLogin(user, unlock_user=True)
                 self.sendResetPasswordEmail(user, system_email, unlock_user=True)
             self.interfaces.userDb.setUserActive(user, is_active)
 
--- a/dataactbroker/handlers/userHandler.py
+++ b/dataactbroker/handlers/userHandler.py
@@ -443,10 +443,10 @@ class UserHandler(UserInterface):
         self.session.merge(template)
         self.session.commit()
 
-    def updateLastLogin(self, user):
+    def updateLastLogin(self, user, unlock_user=False):
         """ This updates the last login date to today's datetime for the user to the current date upon successful login.
         """
-        user.last_login_date = time.strftime("%c")
+        user.last_login_date = time.strftime("%c") if not unlock_user else None
         self.session.commit()
 
     def setUserActive(self, user, is_active):


dataactbroker/handlers/accountHandler.py
```diff
@@ -387,6 +387,10 @@ def updateUser(self, system_email):
         if requestDict.exists("is_active"):
             is_active = bool(requestDict.getValue("is_active"))
             if not self.isUserActive(user) and is_active:
+                # Reset password count to 0
+                self.resetPasswordCount(user)
+                # Reset last login date so the account isn't expired
+                self.interfaces.userDb.updateLastLogin(user, unlock_user=True)
```
jworcestBAH: Adding the password count reset is a good idea, but I'm less certain of setting the last login date, as we're pretty much putting in a false piece of information here.  Is there any way we can do this that doesn't involve cleverly deceiving ourselves?

nmonga91: I put the unlock_user param for that reason. That flags it to be set to None so it's empty in the database. That way it doesn't look like they've logged in recently. With a blank last_login_date the only reasons are that they've never logged in before or they've just been unlocked.

klrBAH:
[run tests]

dataactbroker/handlers/accountHandler.py
jworcestBAH: Oh right understanding that now, but we're still putting a false piece of info in here, this user has logged in before, what happens if we then want to know when they did last log in?

nmonga91: That becomes a much trickier scenario that I think we can handle later. If we store their last login date to be accurate, their account expiration logic needs to happen differently as a result.

jworcestBAH: Is there anything currently present we can use to check if they've been unlocked?  Maybe add an unlock_date that we can check as an alternative before locking due to inactivity?

nmonga91: No nothing like that in place but we do have the updated_at field at our disposal

bsweger: Is the last user login date a piece of information needed to meet our security requirements?

nmonga91: No, it's just used to see when the user should have their account expire. It's not used for anything more than that for now

bsweger: Thanks--that's good to know. @kaitlin predicted this would happen as we implement a custom user/login handling...all of these scenarios. Are we moving forward with the MAX login for sure? If so, will it handle this scenario?

nmonga91: from what I understood, MAX wasn't on the table for the time being until we're told otherwise.

bsweger: Ok. I agree with @jworcestBAH that it would better not to blow away the last login info as a way trick the password reset trigger. Unfortunately, I think the extra work to implement this correctly is a function of our choosing to custom-code our own user and login handling.

nmonga91: Would you both be opposed to having this is as a separate story? I'm definitely in agreement that is shouldn't be handled appropriately but we may not have the bandwidth to do it this sprint.

bsweger: Ok, let's put it as a technical debt task and use your method of the unlock_user parameter as a stop-gap.

kaitlin: Just FYI, the main blockers to MAX right now are the MAX team bothering to
respond to me in an email, and getting .gov emails for the dev team
(required for MAX log in).

On Fri, May 27, 2016 at 2:28 PM, Becky Sweger <notifications@github.com>
wrote:

> In dataactbroker/handlers/accountHandler.py
> <https://github.com/fedspendingtransparency/data-act-broker-backend/pull/63#discussion_r64948162>
> :
>
> > @@ -387,6 +387,10 @@ def updateUser(self, system_email):
> >          if requestDict.exists("is_active"):
> >              is_active = bool(requestDict.getValue("is_active"))
> >              if not self.isUserActive(user) and is_active:
> > +                # Reset password count to 0
> > +                self.resetPasswordCount(user)
> > +                # Reset last login date so the account isn't expired
> > +                self.interfaces.userDb.updateLastLogin(user, unlock_user=True)
>
> Ok, let's put it as a technical debt task and use your method of the
> unlock_user parameter as a stop-gap.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/fedspendingtransparency/data-act-broker-backend/pull/63/files/e81d5afe61c941c6cf59b6eaceefd00e48eae014#r64948162>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAEnIzZB7v9qezEu-eMJoztlvwpiB0Mzks5qFze_gaJpZM4IotEa>
> .
>



-- 
Kaitlin Devine
Treasury/18F

Need to schedule a meeting? Check out my Calendly page
<https://calendly.com/kaitlin-devine>

bsweger: Thanks for the update, @kaitlin. The idea that we might someday be able to migrate to MAX is an even better reason not to rewrite the account expiration logic at this time.

Gonna merge.

PR #64, opened 2016-05-27
Resetting fields that cause a user's account to be locked when they a…

nmonga91:
…re being unlocked by an admin

@jworcestBAH or @bsweger


jworcestBAH:
Merging, we have a task in Jira to look for a better way that doesn't claim the user has never logged in, so we'll address that then.

PR #65, opened 2016-05-31
Removing obsolete change_status route that has been replaced by updat…

jworcestBAH:
…e_user route

Frontend no longer uses this, we left it in during the PR last sprint to still work while frontend was transitioning.


bsweger:
+100 for deleted unused code. Merging!

PR #66, opened 2016-05-31
Db 429 add staging models

bsweger:
This is the first step towards getting away from the dynamically-created staging tables used by the broker.

This PR contains four staging models (one for each file being submitted by agencies) along with their corresponding migrations.

No other application logic is included. Submitting this now becayse @afrasier needs to know what the staging tables will look like for his data store story.

A few notes:
* There are no FK relationships between models as I originally envisioned, b/c that would require setting up unique constraints. We still don't know if we're supposed to be enforcing those, and we def. don't know if we'd want the db to enforce it.
* The models for files A, B, and C contain a `tas` field that will have a concatenated version of the TAS codes for easier joins between tables.
* The models for files A, B, and C specify index on tas, tas/object class/program activity code as appropriate.
* Model field names that exceed 63 characters are aliases. Those columns contain a "key" attribute that will allow application code to access the models using the original element names (_i.e._, those in the submitted .csv files)


klrBAH:
[run tests]

dataactcore/models/stagingModels.py
```diff
@@ -1,15 +1,261 @@
...
+    __tablename__ = "object_class_program_activity"
+
+    object_class_program_activity_id = Column(Integer, primary_key=True)
+    agencyidentifier = Column(Text, nullable=False)
+    allocationtransferagencyidentifier = Column(Text)
+    availabilitytypecode = Column(Text)
+    beginningperiodofavailability = Column(Text)
+    bydirectreimbursablefundingsource = Column(Text, nullable=False)
+    deobligationsrecoveriesrefundsprioryrbyprogobjectclass_cpe = Column(
+        Numeric, nullable=False, key="deobligationsrecoveriesrefundsdofprioryearbyprogramobjectclass_cpe")
```
jworcestBAH: So can I access this by either ObjectClassProgramActivity.deobligationsrecoveriesrefundsprioryrbyprogobjectclass_cpe or ObjectClassProgramActivity.deobligationsrecoveriesrefundsdofprioryearbyprogramobjectclass_cpe?  And getattr will work with both?  Or are there limits on how the key can be used?

dataactcore/models/stagingModels.py
```diff
@@ -1,15 +1,261 @@
...
+    ussgl497100_downadjsprioryrunpaiddelivordersobligrec_cpe = Column(
+        Numeric, key="ussgl497100_downwardadjustmentsofprioryearunpaiddeliveredordersobligationsrecoveries_cpe")
+    ussgl497200_downadjsprioryrpaiddelivordersobligrefclt_cpe = Column(
+        Numeric, key="ussgl497200_downwardadjustmentsofprioryearpaiddeliveredordersobligationsrefundscollected_cpe")
+    ussgl498100_upadjsprioryeardeliveredordersobligunpaid_cpe = Column(
+        Numeric, key="USSGL498100_UpwardAdjustmentsOfPriorYearDeliveredOrdersObligationsUnpaid_CPE")
+    ussgl498200_upadjsprioryrdelivordersobligpaid_cpe = Column(
+        Numeric, key="ussgl498200_upwardadjustmentsofprioryeardeliveredordersobligationspaid_cpe")
+    tas = Column(Text, nullable=False, default=concatTas, onupdate=concatTas)
+    Index("ix_award_financial_tas_oc_pa", "tas", "objectclass", "programactivitycode", unique=False)
```
jworcestBAH: Just to be sure, this would still be used for queries just on TAS, right?  Is there an efficiency benefit to having a separate index just on TAS?

jworcestBAH: Did a bit of research, this seems fine now for TAS, though this will not help if we have cases where we are querying for object class or program activity alone, or a case where we query by TAS + PA without OC.  I'll look at the few cross-file rules we have right now and see if we're breaking this pattern yet.

jworcestBAH: It seems the current rules only need TAS so far.

dataactcore/models/stagingModels.py
```diff
@@ -1,15 +1,261 @@
...
+    primaryplaceofperformancecountrycode = Column(Text)
+    primaryplaceofperformancecountyname = Column(Text)
+    primaryplaceofperformanceforeignlocationdescription = Column(Text)
+    primaryplaceofperformancestatename = Column(Text)
+    primaryplaceofperformancezipplus4 = Column(Text)
+    recordtype = Column(Integer)
+    sai_number = Column(Text)
+    totalfundingamount = Column(Numeric)
+    uri = Column(Text)
+
```
jworcestBAH: This and file C (award financial) could use an index on FAIN and URI.  File C will also eventually need one on PIID and ParentAwardId.

jworcestBAH:
Finished reviewing.

dataactcore/models/stagingModels.py
bsweger: Good question. It works with both. Attaching a screenshot of messing around with the ObjectClassProgramActivity model in the REPL.

screen shot 2016-06-01 at 11 32 37 am

dataactcore/models/stagingModels.py
bsweger: I included object class and program activity in this index because we have at least one rule that will need to group by that combination and because it's conceivable that the optimizer could use this as a covering index in some cases. But you're right that there's only so much predicting we can do at the moment. We can add individual indices on program activity an object class if we need to.

dataactcore/models/stagingModels.py
jworcestBAH: And just because that screenshot makes me nervous, they are the same attribute yes?  So if you follow those lines by getting the long name you'll get 333?

bsweger: Good call @jworcestBAH. You're right...the `getattr` is returning different values. I had the key logic working as intended last week when I was messing around with dynamically-created models, so I'm doing more research and will post back with better specifics on how this works.

bsweger: So, the technical explanation for your findings is that I was doing it wrong. I found the fix..better screenshot tomorrow morning.

bsweger: @jworcestBAH Hopefully this is a better idea of what's going on when the aliases are implemented correctly (working on making the fixes to the rest of the columns now).

The list of properties after a new ObjectClassProgramActivity is created...the short name doesn't appear and can't be accessed via `getattr`

screen shot 2016-06-02 at 9 25 46 am

Nothing stops you from manually creating a property on the object that matches the shorter column name, but the ORM ignores it.

screen shot 2016-06-02 at 9 19 42 am

Essentially, we should continue to reference the column names as used on the incoming .csvs and ignore the short names. They're only there for use by the physical db.

Reference: http://docs.sqlalchemy.org/en/latest/orm/mapping_columns.html

bsweger:
@jworcestBAH Ok, pushed some changes to: 
* fix up the column aliasing behavior.
* add the indices you suggest (also threw in one on `piid` on `award_financial`)
* change the way of defining the multi-field indices in the models so that alembic behaves properly (previously, it was generating code to drop those on subsequent autogens).

To keep things tidy, I updated the existing migration file instead of creating a new one. Sorry--I know that's annoying on your end. You'll have to drop the 4 new tables on your db and run `scripts/resetAlembicVersion.py 849e1ac35007` to reset the migration.

*Note* I had to change `resetAlembicVersion.py` because it didn't include the staging db, so pull these updates before running.

jworcestBAH:
Okay, looks good now, we just need to make sure moving forward that we continue to use the long names in places like the file_columns table if the short names don't work on the ORM models.

PR #67, opened 2016-06-01
Hotfix for alembic backwards migrations

bsweger:
Some of the downgrade text that alembic presumably auto-generated was breaking things when running downgrades.


jworcestBAH:
This probably occurred because these tables were originally created referring to an explicit sequence, which they no longer are.  Merging now.

PR #70, opened 2016-06-02
Db 557 validations

jworcestBAH:
This includes rules A20, B11, B12, C18 and the first half of A16


dataactvalidator/config/awardFinancialFields.csv
```diff
@@ -12,7 +12,7 @@ fain,FALSE,str,30,"C8, C9, C14"
 uri,FALSE,str,70,"C8, C9, C14"
 programactivityname,FALSE,str,164,B9
 programactivitycode,FALSE,str,4,B10
-objectclass,TRUE,int,4,B11
+objectclass,TRUE,str,4,B11
```
bsweger: 💯 💯 💯

dataactvalidator/filestreaming/loadCgac.py
```diff
@@ -1,9 +0,0 @@
-from dataactvalidator.filestreaming.loaderUtils import LoaderUtils
-from dataactvalidator.interfaces.validatorValidationInterface import ValidatorValidationInterface
-from dataactcore.models.domainModels import CGAC
-
-def loadCgac(filename):
-    LoaderUtils.loadCsv(filename,CGAC,ValidatorValidationInterface(),{"cgac":"cgac_code","agency":"agency_name"},{"cgac_code":{"pad_to_length":3,"skip_duplicates":True}})
-
-if __name__ == '__main__':
-    loadCgac("../config/cgac.csv")
```
bsweger: It would be great to reference the config folder by using the info we have in config.py. I run the application initialize.py scripts from the main dataactbroker folder, and this line failed.

tests/runTests.py
```diff
@@ -28,11 +28,11 @@ def runTests(argv=''):
     # Create test suite
     suite = unittest.TestSuite()
 
-    suite.addTests(unittest.makeSuite(LoginTests))
-    suite.addTests(unittest.makeSuite(FileTests))
-    suite.addTests(unittest.makeSuite(UserTests))
-    suite.addTests(unittest.makeSuite(ValidatorTests))
-    suite.addTests(unittest.makeSuite(JobTests))
+    #suite.addTests(unittest.makeSuite(LoginTests))
```
bsweger: Whoops, will need to uncomment the other tests before we merge.

bsweger:
@jworcestBAH done reviewing.

The only things that I think need to be addressed before merging are in the line comments above.

I have a question about our instructions to apply the CGAC check to the agency identifiers that are in the TAS, since we're already validating against a TAS list (guessing it might be for better error messages?). Also wondering why that rule doesn't apply to award-level fields like `FundingAgencyCode` in file D2. I'll ask the schema team directly.

Thanks for always being so good about writing unittests!

dataactvalidator/filestreaming/loadCgac.py
jworcestBAH: Good catch, fixed this here and in unit tests.

tests/runTests.py
jworcestBAH: Fixed

jworcestBAH:
Yes, my assumption behind this type of rule is that it gives more specific errors than just a generic TAS error.  It definitely would make sense to apply it to other fields that should be an agency ID.

jworcestBAH:
[run tests] better Jenkins...

jworcestBAH:
Okay alembic is behaving now.

bsweger:
Looks good--merging!

PR #71, opened 2016-06-03
Db 575 agency code

nmonga91:
@jworcestBAH or @bsweger 

This was a pretty hefty change. Please let me know if anything was overlooked. Thanks!


dataactcore/models/jobModels.py
```diff
@@ -29,7 +29,7 @@ class Submission(Base):
     submission_id = Column(Integer, primary_key=True)
     datetime_utc = Column(DateTime)
     user_id = Column(Integer, nullable=False) # This refers to the users table in the User DB
-    agency_name = Column(Text)
+    cgac_code = Column(Text)
```
bsweger: I see that above we're using a surrogate key to reference back to a submission's user. Does it make sense to follow that model here and use the cgac id instead of the code?

dataactcore/models/userModel.py
```diff
@@ -13,7 +13,7 @@ class User(Base):
     username = Column(Text)
     email = Column(Text)
     name = Column(Text)
-    agency = Column(Text)
+    cgac_code = Column(Text)
```
bsweger: Same question here about using a surrogate key instead of a code.

dataactcore/scripts/updateCGAC.py
```diff
@@ -0,0 +1,46 @@
+from dataactcore.models.jobModels import Submission
+from dataactcore.models.userModel import User
+from dataactcore.models.validationInterface import ValidationInterface
+from dataactcore.models.userInterface import UserInterface
+from dataactcore.models.jobTrackerInterface import JobTrackerInterface
+
+def updateCGAC():
```
bsweger: This is a one-time script to convert our current users/jobs from agency to cgac, yes? What's the standard procedure for something like this? Do we need it in the repo, do we just remove it once the data is converted?

bsweger:
Hi @nmonga9, still working on the review. My biggest question so far is the use of CGAC code being stored with the user and submission instead of the surrogate key (`cgac_id`).

I come from the `trust no one` school of data programming so have always used a surrogate key in these kinds of situations. I recognize, though, that doing so here is complicated by the fact that we have stuff spread out across three different database, making it hard to use the ORM for easy relationship mapping.

What do you think?

Pinging @jworcestBAH on this too. We should decide as a group how we want to work with our domain values as we implement more of them.

dataactbroker/handlers/domainHandler.py
```diff
@@ -0,0 +1,28 @@
...
+            interfaces: InterfaceHolder object to DBs
+        """
+        self.interfaces = interfaces
+        self.validationManager = interfaces.validationDb
+
+    def listAgencies(self):
+        agencies = self.validationManager.getAllAgencies()
+        agency_list = []
+
+        for agency in agencies:
```
bsweger: Relevant to the comments re: surrogate keys, I'm used to seeing a pattern of sending an id/name to the front end, and then the front-end returns the id for us to stick in the db. Doe you think we should send the cgac_id too?

bsweger: (though I realize you all mocked this up already...am more curious about getting the overall surrogate key question answered)

jworcestBAH:
I will throw in that we are indexing on cgac_code in that table.  The biggest issue for me is that without an explicit foreign key (it's in a different DB), we are not protected from cgac_ids changing when the table is reloaded, which actually happens automatically on deploy at the moment.  We may want to rethink some aspect of domain value table handling, this applies to TAS numbers as well.  Currently we're reloading those tables on each deploy, and unless the latest staging refactor changed this we do not have anything with foreign keys to these tables.

jworcestBAH:
Possibly we should refactor domain table loading to overwrite existing records rather than the current wipe and reload, but if so we still need to drop old records that are not in the new file, or we need every domain value table to have an inactive flag.

dataactbroker/handlers/accountHandler.py
```diff
@@ -443,7 +444,7 @@ def listUsers(self):
         isAgencyAdmin = True if self.interfaces.userDb.hasPermission(user, "agency_admin") else False
         try:
             if isAgencyAdmin:
-                users = self.interfaces.userDb.getUsers(agency=user.agency)
+                users = self.interfaces.userDb.getUsers(cgac_code=user.cgac_code)
```
bsweger: see other questions re: surrogate key

bsweger: see other questions re: surrogate key

dataactbroker/handlers/accountHandler.py
```diff
@@ -100,7 +101,7 @@ def login(self,session):
                         if(self.interfaces.userDb.hasPermission(user, permission.name)):
                             permissionList.append(permission.permission_type_id)
                     self.interfaces.userDb.updateLastLogin(user)
-                    return JsonResponse.create(StatusCode.OK,{"message":"Login successful","user_id": int(user.user_id),"name":user.name,"title":user.title ,"agency":user.agency, "permissions" : permissionList})
+                    return JsonResponse.create(StatusCode.OK,{"message":"Login successful","user_id": int(user.user_id),"name":user.name,"title":user.title ,"cgac_code":user.cgac_code, "permissions" : permissionList})
```
bsweger: Would it be useful to whoever calls this function to also have the CGAC agency name included in the returned user info?

dataactbroker/handlers/accountHandler.py
```diff
@@ -173,12 +174,12 @@ def ThreadedFunction (from_email="",username="",title="",agency="",userEmail=""
             threadedDatabase =  UserHandler()
             try:
                 for user in threadedDatabase.getUsersByType("website_admin"):
-                    emailTemplate = {'[REG_NAME]': username, '[REG_TITLE]':title, '[REG_AGENCY]':agency,'[REG_EMAIL]' : userEmail,'[URL]':link}
+                    emailTemplate = {'[REG_NAME]': username, '[REG_TITLE]':title, '[REG_AGENCY]':cgac_code,'[REG_EMAIL]' : userEmail,'[URL]':link}
```
bsweger: Does it make sense to use the CGAC description instead of the code in these e-mail templates (or some combination of both?). I suspect the CGAC codes aren't meaningful for most people.

dataactbroker/handlers/accountHandler.py
```diff
@@ -479,7 +480,7 @@ def listUsersWithStatus(self):
             return JsonResponse.error(exc,exc.status)
         userInfo = []
         for user in users:
-            thisInfo = {"name":user.name, "title":user.title,  "agency":user.agency, "email":user.email, "id":user.user_id }
+            thisInfo = {"name":user.name, "title":user.title,  "cgac_code":user.cgac_code, "email":user.email, "id":user.user_id }
```
bsweger: Same question as above re: is it useful for the caller of this function to also get the agency name? Especially since that's what was being returned previously.

dataactbroker/handlers/accountHandler.py
```diff
@@ -614,7 +615,7 @@ def getCurrentUser(self,session):
         for permission in self.interfaces.userDb.getPermissionList():
             if(self.interfaces.userDb.hasPermission(user, permission.name)):
                 permissionList.append(permission.permission_type_id)
-        return JsonResponse.create(StatusCode.OK,{"user_id": int(uid),"name":user.name,"agency":user.agency,"title":user.title, "permissions" : permissionList, "skip_guide":user.skip_guide})
+        return JsonResponse.create(StatusCode.OK,{"user_id": int(uid),"name":user.name,"cgac_code":user.cgac_code,"title":user.title, "permissions" : permissionList, "skip_guide":user.skip_guide})
```
bsweger: Same question re: returning agency name.

dataactbroker/handlers/fileHandler.py
```diff
@@ -222,7 +222,7 @@ def getStatus(self):
             # Build dictionary of submission info with info about each job
             submissionInfo = {}
             submissionInfo["jobs"] = []
-            submissionInfo["agency_name"] = submission.agency_name
+            submissionInfo["cgac_code"] = submission.cgac_code
```
bsweger: Same question re: surrogate keys for CGAC.

dataactbroker/handlers/jobHandler.py
```diff
@@ -27,7 +27,7 @@ def getSubmissionById(self,submissionId):
 
     def getSubmissionsByUserAgency(self,user,limit=5):
         """ Returns all submissions associated with the specified user's agency """
-        return self.session.query(Submission).filter(Submission.agency_name == user.agency).order_by(Submission.updated_at.desc()).limit(limit).all()
+        return self.session.query(Submission).filter(Submission.cgac_code == user.cgac_code).order_by(Submission.updated_at.desc()).limit(limit).all()
```
bsweger: Same question re: surrogate key

dataactbroker/handlers/userHandler.py
```diff
@@ -58,12 +58,12 @@ def deleteToken(self,token):
         self.session.delete(oldToken)
         self.session.commit()
 
-    def getUsers(self, agency=None):
+    def getUsers(self, cgac_code=None):
```
bsweger: Same question throughout the userHandler functions re: surrogate key vs using the CGAC code

tests/baseTestAPI.py
```diff
@@ -117,7 +120,7 @@ def setUpClass(cls):
 
         # Set the Agency for the agency user
         agencyUser = userDb.getUserByEmail(test_users['agency_user'])
-        agencyUser.agency = "testAgency"
+        agencyUser.cgac_code = "testAgency"
```
bsweger: Since real-life CGAC codes are 4 digit numbers, does it make more sense to use something like that for our unit test data for better parallels to what we'd see in actual use?

tests/baseTestAPI.py
```diff
@@ -145,7 +148,7 @@ def setUpClass(cls):
         admin = userDb.getUserByEmail(test_users['admin_email'])
         userDb.setPassword(admin, admin_password, Bcrypt())
         admin.name = "Mr. Manager"
-        admin.agency = "Unknown"
+        admin.cgac_code = "Unknown"
```
bsweger: Same question re: more realistic CGAC codes.

tests/fileTests.py
```diff
@@ -32,7 +32,7 @@ def setUpClass(cls):
 
         # setup submission/jobs data for test_check_status
         cls.status_check_submission_id = cls.insertSubmission(
-            cls.jobTracker, cls.submission_user_id, agency = "Department of the Treasury", startDate = "10/2015", endDate = "06/2016", is_quarter = True)
+            cls.jobTracker, cls.submission_user_id, cgac_code = "Department of the Treasury", startDate = "10/2015", endDate = "06/2016", is_quarter = True)
```
bsweger: If we want to mimic Dept. of Treasury, the CGAC code would be '020'

tests/userTests.py
```diff
@@ -53,7 +53,7 @@ def setUpToken(self,email):
     def test_registration_no_token(self):
         """Test without token."""
         self.logout()
-        postJson = {"email": "user@agency.gov", "name": "user", "agency": "agency", "title": "title", "password": "userPass"}
+        postJson = {"email": "user@agency.gov", "name": "user", "cgac_code": "agency", "title": "title", "password": "userPass"}
```
bsweger: Very minor nit: changing the word "agency" to "000", "999" or similar in this fake JSON would make it more explicit that CGAC code doesn't represent an agency name.

tests/userTests.py
```diff
@@ -63,7 +63,7 @@ def test_registration(self):
         self.logout()
         email = self.test_users["change_user_email"]
         self.setUpToken(email)
-        postJson = {"email": email, "name": "user", "agency": "agency", "title": "title", "password": self.user_password}
+        postJson = {"email": email, "name": "user", "cgac_code": "agency", "title": "title", "password": self.user_password}
```
bsweger: Same comment as above re: making the fake JSON explicit about what type of info is being represented.

tests/userTests.py
```diff
@@ -83,14 +83,14 @@ def test_registration_empty(self):
...
-            "Request body must include email, name, agency, title, and password")
+            "Request body must include email, name, cgac_code, title, and password")
 
     def test_registration_bad_email(self):
         """Test user registration with invalid email."""
         self.logout()
         self.setUpToken("user@agency.gov")
         postJson = {"email": "fake@notreal.faux",
-                "name": "user", "agency": "agency",
+                "name": "user", "cgac_code": "agency",
```
bsweger: See above comment re: JSON

tests/userTests.py
```diff
@@ -293,7 +293,7 @@ def test_current_user(self):
         response = self.app.get("/v1/current_user/", headers={"x-session-id":self.session_id})
         self.check_response(response, StatusCode.OK)
         self.assertEqual(response.json["name"], "Mr. Manager")
-        self.assertEqual(response.json["agency"], "Unknown")
+        self.assertEqual(response.json["cgac_code"], "Unknown")
```
bsweger: See above comment re: JSON

dataactbroker/interfaces/validationBrokerInterface.py
```diff
@@ -0,0 +1,16 @@
+from dataactcore.models.validationInterface import ValidationInterface
+from dataactcore.models.domainModels import CGAC
+
+
+class ValidationBrokerInterface(ValidationInterface):
```
bsweger: I'd suggest moving this function to an interface in the core, since it will likely have use beyond the broker.

That would also allow us to continue avoiding injecting validator dependencies into the broker and vice-versa. It doesn't seem terrible to do that if we absolutely need to, but it doesn't seem like we do here.

bsweger:
@nmonga91 Done reviewing. Whew--what a lot of work...thanks for the heavy lift on this!

My other main comment in addition to the surrogate key question is to note that we're switching from a descriptive name to a code, so there might be a few places that need an extra translation step. I tried to note those inline, but for example:
* the e-mail templates expect an agency name, and we're now passing back a CGAC code
* some of the user functions that used to send an agency name are now sending a CGAC code...not sure of the downstream effects.

Both of these themes lead back to a larger question as we start adding domain values to the broker. Some domain values (like CGAC) are needed by the validator, staging, broker, and job databases. The fact that the domain tables can only live in one db means we can't use SQL Alchemy to describe these relationships and handle the work of automatically returning descriptions and codes that accompany a surrogate key, for example. (At least I don't think so).

For this PR, we need to make a decision about the surrogate key question.

Longer-term, we need to consider the trade-offs of having separate databases.

dataactcore/scripts/updateCGAC.py
nmonga91: I added it in the repo since it isn't hasn't been run yet. Once it's been successfully PR'd and merged, we'll run the script then remove it from the repo. That was my thinking of the process.

tests/baseTestAPI.py
nmonga91: I would agree it's more useful although not necessary since it's stored as a string in the database. For the unit tests, we need to use some code (something like we discussed as the default one we'd use internally) that is realistic but doesn't rely on something that's actually in the database.

bsweger:
Update on the surrogate key question, for posterity. As @jworcestBAH points out, we re-load the domain value tables (including CGAC) whenever the app is deployed, so we can't use surrogate keys as things stand now. Given that imminent CGAC changes are 1) unlikely and 2) wouldn't really affect the broker's code logic, we'll keep referencing the codes.

In JIRA, we'll put a tech debt task to consider the larger question of whether or not we should to a more nuanced loading of the domain tables.

dataactbroker/userRoutes.py
```diff
@@ -19,7 +19,7 @@ def add_user_routes(app,system_email,bcrypt):
     #check the session to make sure register is set to prevent any one from using route
     @permissions_check(permissionList=["check_email_token"])
     def register_user():
-        """ Expects request to have keys 'email', 'name', 'agency', and 'title' """
+        """ Expects request to have keys 'email', 'name', 'cgac_code', and 'title' """
```
bsweger: Thanks for catching the docstring changes!

bsweger:
Thanks for the updates, @nmonga91! Merging...

PR #72, opened 2016-06-06
DB-429: in rules files, use rule timing description instead of surrogate key

bsweger:
This PRs shores up the rule load process by replacing the rule timing surrogate key in the rules .csv files with the corresponding description. The load process itself now uses the description to query for the corresponding surrogate key prior to inserting the rules.


jworcestBAH:
Looks good, merging

PR #73, opened 2016-06-06
Added last_updated to check_status route

jworcestBAH:
Check status route now includes last_updated

### DIFF
--- a/tests/fileTests.py
+++ b/tests/fileTests.py
@@ -18,7 +18,6 @@ class FileTests(BaseTestAPI):
     updateSubmissionId = None
     filesSubmitted = False
     submitFilesResponse = None
-    status_check_time = {"year":2016,"month":01,"day":02,"hour":03,"minute":04,"second":05}
 
     @classmethod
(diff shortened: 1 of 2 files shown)


tests/fileTests.py
```diff
@@ -18,6 +18,7 @@ class FileTests(BaseTestAPI):
     updateSubmissionId = None
     filesSubmitted = False
     submitFilesResponse = None
+    status_check_time = {"year":2016,"month":01,"day":02,"hour":03,"minute":04,"second":05}
```
bsweger: Is this line being used in the tests anywhere (couldn't find it, but looked pretty quickly)?

bsweger:
@jworcestBAH done reviewing--one small question inline above...

tests/fileTests.py
jworcestBAH: Good catch, this was part of my attempt to manually set the last updated time, which didn't work out.  I've removed it now.

bsweger:
Looks good--merging!

PR #74, opened 2016-06-06
Updated tas

jworcestBAH:
Updated to new all_tas_betc file


bsweger:
I won't pretend that I've reviewed this new TAS file but trust that everything is in order here.

PR #75, opened 2016-06-06
Db 559 program activity

jworcestBAH:
Implements rules B9 and B10 to validate program activity


dataactcore/models/domainModels.py
```diff
@@ -25,4 +25,14 @@ class ObjectClass(Base):
...
     object_class_code = Column(Text,nullable=False,index=True,unique=True)
-    object_class_name = Column(Text)
\ No newline at end of file
+    object_class_name = Column(Text)
+
+class ProgramActivity(Base):
+    __tablename__ = "program_activity"
+    program_activity_id = Column(Integer, primary_key=True)
+    budget_year = Column(Text,nullable=False)
+    agency_id = Column(Text,nullable=False)
```
bsweger: Is the agency_id here the same as a CGAC code? If so, should we consider naming this column cgac_code to be consistent with the CGAC model? I'm honestly a bit torn on how to refer to agencies throughout the broker, since CGAC doesn't seen to be common parlance. That said, using it consistently would remove any ambiguity around the department/agency/bureau discussion.

bsweger: Or is this agency code being used as part of a TAS-like component in this table? In those cases, it seems like we use the `agency` moniker.

bsweger: This is such helpful feedback, I know.

dataactcore/models/domainModels.py
```diff
@@ -25,4 +25,14 @@ class ObjectClass(Base):
...
     object_class_id = Column(Integer, primary_key=True)
     object_class_code = Column(Text,nullable=False,index=True,unique=True)
-    object_class_name = Column(Text)
\ No newline at end of file
+    object_class_name = Column(Text)
+
+class ProgramActivity(Base):
+    __tablename__ = "program_activity"
+    program_activity_id = Column(Integer, primary_key=True)
+    budget_year = Column(Text,nullable=False)
```
bsweger: Would we gain anything by having this field be an integer instead of a string, since this table reflects domain values and not user-submitted data?

dataactcore/models/domainModels.py
```diff
@@ -25,4 +25,14 @@ class ObjectClass(Base):
     __tablename__ = "object_class"
     object_class_id = Column(Integer, primary_key=True)
     object_class_code = Column(Text,nullable=False,index=True,unique=True)
-    object_class_name = Column(Text)
\ No newline at end of file
+    object_class_name = Column(Text)
+
+class ProgramActivity(Base):
```
bsweger: Is there a combination of fields in this table that would qualify for a unique index?

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -10,9 +10,15 @@ def loadCgac(filename):
 def loadObjectClass(filename):
     LoaderUtils.loadCsv(filename,ObjectClass,ValidatorValidationInterface(),{"max_oc_code":"object_class_code","max_object_class_name":"object_class_name"},{"object_class_code":{"skip_duplicates":True}})
 
+def loadProgramActivity(filename):
+    LoaderUtils.loadCsv(filename, ProgramActivity, ValidatorValidationInterface(), {"year":"budget_year","agency_id":"agency_id",
+        "alloc_id":"allocation_transfer_id","account":"account_number","pa_code":"program_activity_code","pa_name":"program_activity_name"},
+        {"program_activity_code":{"pad_to_length":4},"agency_id":{"pad_to_length":3},"account_number":{"pad_to_length":4},"allocation_transfer_id":{"pad_to_length":3}})
```
bsweger: I'm curious about the padToLength function (technically out of scope for this review). I see that it also functions as check on the length of incoming data, but couldn't it use Python's zfill string function to do the actual 0 padding?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -72,7 +73,7 @@ def loadCsv(cls,filename,model,interface,fieldMap,fieldOptions):
                     options = fieldOptions[field]
                     if "pad_to_length" in options:
                         padLength = options["pad_to_length"]
-                        row[field] = cls.padToLength(row[field],padLength)
+                        row[field] = Validator.padToLength(row[field],padLength)
```
bsweger: I'm curious about the padToLength function. I see that it also functions as check on the length of incoming data, but couldn't it use Python's zfill string function to do the actual 0 padding?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -453,6 +453,83 @@ def rule_exists_in_table(cls, data, value, rule, datatype, interfaces, record):
...
+
+        Returns:
+            padded string of length padLength
+        """
+        if data is None:
+            # Convert None to empty string so it can be padded with zeros
+            data = ""
+        data = data.strip()
+
+        if len(data) < padLength:
```
bsweger: Doing a length check and then using Python's `zfill` string function might be a more readable way to do this.

Something like:
```
if len(data) < padLength:
        # if data is already the correct length, zfill leaves it alone
        data = data.strip().zfill(padLength)
else:
        too long!

**Comment by bsweger** [comment]
@jworcestBAH Done reviewing.

To be fully transparent: 
* The code for the new data model/load is fully reviewed
* James and I had a back-channel convo about the specifics of implementing the program activity rule. The corresponding logic is reviewed for security/project standards. However, I haven't dug deeply into the logic of how `validator.py` is doing the implementation.

File: dataactcore/models/domainModels.py (hunk 1 above)
Comment (jworcestBAH) [inline dataactcore/models/domainModels.py]: This is here as part of a TAS, the next field is allocation_transfer_id.  Both should be CGAC codes, but we need to make the distinction between these two as they are both parts of a TAS.

File: dataactcore/models/domainModels.py (hunk 2 above)
Comment (jworcestBAH) [inline dataactcore/models/domainModels.py]: We are doing comparisons between this and beginningperiodofavailability in files B and C, which we are representing as text.  Since we're filtering by values from that, I'd prefer that they be the same type to avoid complications.

File: dataactcore/models/domainModels.py (hunk 3 above)
Comment (jworcestBAH) [inline dataactcore/models/domainModels.py]: Good point, I've added an index over the six fields we're querying against, which should be a unique combination.

File: dataactvalidator/validation_handlers/validator.py (hunk 6 above)
Comment (jworcestBAH) [inline dataactvalidator/validation_handlers/validator.py]: Agreed, this change was moving an existing function from before we started doing zfill for this, I've changed it to use that.

File: dataactvalidator/filestreaming/loaderUtils.py [hunk 7]
```diff
@@ -83,7 +84,14 @@ def loadCsv(cls,filename,model,interface,fieldMap,fieldOptions):
                             valuePresent[field][row[field]] = True
                     record = model(**row)
                 if not skipInsert:
-                    interface.session.merge(record)
+                    try:
+                        interface.session.merge(record)
+                    except IntegrityError as e:
+                        # Hit a duplicate value that violates index, skip this one
+                        print("".join(["Warning: Skipping this row: ",str(row)]))
```
bsweger: Nice, I like the soft landing here instead of just letting the db blow up if we get duplicates.

bsweger:
Looks good--merging!

PR #76, opened 2016-06-06
Db 567 numeric field comma

nmonga91:


dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -120,6 +122,32 @@ def cleanLength(length):
             raise ValueError("Length must be positive")
         return length
 
+    @staticmethod
+    def cleanRow(row):
+        for key in row.keys():
+            value = row[key]
+            if value is not None:
```
jworcestBAH: Do we want to consider what the type of the field is supposed to be, we should be able to get that out of file_columns.field_type_id.  Possibly the answer is no, and if someone puts "9,4" into a random string we want to convert it to "94".  I just want to raise the point that as written this will apply to all strings that can be converted to ints.

nmonga91: I wouldn't think it matters. But with that in mind, is there any possibility of any field containing a list of numbers? That would throw this implementation off. It may just be safer to check the field type. What do you think?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -231,8 +231,7 @@ def validate(record,rules,csvSchema,fileType,interfaces):
         for rule in multiFieldRules:
             if not Validator.evaluateRule(record,rule,None,interfaces,record):
                 recordFailed = True
-                failedRules.append(["MultiField", "".join(["Failed rule: ",str(rule.description)]), Validator.getMultiValues(rule, record, interfaces)])
-
+                failedRules.append(["MultiField", "".join(["Failed rule: ",str(rule.rule_type.name)]), Validator.getMultiValues(rule, record, interfaces)])
```
jworcestBAH: I don't agree with this change, the description is intended to be a human-readable explanation of the rule that failed, replacing that with just the type of rule that failed will not give users enough information in the error report to fix their files.

nmonga91: This change should not have been committed. Thanks for catching that.

dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -120,6 +122,32 @@ def cleanLength(length):
...
+        return row
+
+    @staticmethod
+    def isNumeric(data):
+        try:
+            int(data)
+            return True
+        except:
+            try:
+                Decimal(data)
```
nmonga91: I don't like the nested values either. Decimal was used in Validation.checkType but float would handle all scenarios. I will change it to only check using float.

jworcestBAH: Are there cases that would pass long but would not pass float?  We may need both of those.

dataactvalidator/filestreaming/fieldCleaner.py
jworcestBAH: Not being a financial expert, I can't rule out ever seeing that.  Checking the field type does seem safer, but there are also a number of fields that are labeled as strings when we know it's really an int (e.g. object class codes).  So if someone gives us a subaccount number of 1,234 or a year of 2,016, do we try to help them?  I can see arguments for both sides here.

jworcestBAH:
Okay, everything looks good now, merging

PR #77, opened 2016-06-06
Db 568 case sensitivity

nmonga91:


dataactvalidator/validation_handlers/validator.py
```diff
@@ -375,12 +374,12 @@ def rule_greater(cls, data, value, rule, datatype, interfaces, record):
     @classmethod
     def rule_equal(cls, data, value, rule, datatype, interfaces, record):
         """Checks that data is equal to specified value"""
-        return Validator.getType(data,datatype) == Validator.getType(value,datatype)
+        return Validator.getType(data.lower(),datatype) == Validator.getType(value.lower(),datatype)
```
jworcestBAH: Are we completely confident that data and value here are always strings?

nmonga91: No, not always. But `data` is a string until it's converted using getType so calling `lower()` on that should be fine.

jworcestBAH:
Everything looks good once we add a bit of checking that data and value are actually strings.  I think there is at least one rule where data ends up being a list.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Multi-field rules get passed the entire record as data, which is a dict.  So it does not always start as a string, and you'll get an error trying to .lower() a dict.  At a minimum if you don't want to do any extra checking, you could explicitly cast as a str with (str(data).lower()) to avoid those errors.

nmonga91: I'll add the check in to see if it's a string and then do lower if so.

jworcestBAH:
Looks like this needs to have dev merged in

jworcestBAH:
Okay, merging

PR #78, opened 2016-06-07
Db 429 use consolidated staging

bsweger:
This PR adds the logic for writing to the new, consolidated staging tables instead of the per-submission, dynamically-created tables.

* All code related to creating, getting the name of, and reflecting the per-submission tables is gone
* The column mapping table and related code is gone
* Submission-related data is retrieved through new methods in the validation staging interfaces that query by file type and submission id
* Valid records are written to the new submission tables, along with row #, job id, and submission id
* Unit tests are updated and old, redundant tests are removed. Coverage is 83% in the development branch and is 83% here.


dataactvalidator/interfaces/validatorStagingInterface.py
```diff
@@ -1,77 +1,52 @@
...
+        if fileType == "award":
+            rec = AwardFinancialAssistance(**record)
+        elif fileType == "award_financial":
+            rec = AwardFinancial(**record)
+        elif fileType == "appropriations":
+            rec = Appropriation(**record)
+        elif fileType == "program_activity":
+            rec = ObjectClassProgramActivity(**record)
+        self.session.add(rec)
+        self.session.commit()
```
jworcestBAH: Not doing batch inserting any more will make longer files quite a bit slower to load.  Could we recover some of the performance simply by not committing here, and waiting until some number of rows has been added before committing?  This needs some research and maybe some performance testing on files with more records.  We did this when we added batch inserting, but that was one of the tests that didn't use actual field names, we could just copy the rows in one of the valid files a few thousand times.

dataactvalidator/interfaces/validatorStagingInterface.py
```diff
@@ -1,77 +1,52 @@
...
-    def countRows(self,table):
-        """ Returns number of rows in the specified table """
-        metadata = MetaData()
-        try:
-            stagingTable = Table(table, metadata, autoload_with=self.engine)
-        except NoSuchTableError:
-            return 0
-        rows = self.session.query(stagingTable).count()
-        self.session.close()
-        return rows
```
jworcestBAH: Don't we still want to be able to get a count of rows in the staging tables?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -229,13 +233,19 @@ def runValidation(self, jobId, interfaces):
...
                     valid, failures = Validator.validate(record,rules,csvSchema,fileType,interfaces)
-                    if(valid) :
+                    if valid:
                         try:
-                            tableObject.insert(record,fileType)
+                            record["job_id"] = jobId
+                            record["submission_id"] = submissionId
+                            # temporary fix b/c we can't use '+4' as a column alias :(
+                            if "primaryplaceofperformancezip+4" in record:
+                                record["primaryplaceofperformancezipplus4"] = record["primaryplaceofperformancezip+4"]
```
jworcestBAH: I hope that this is temporary, do we expect the next version of the schema to not have the '+'?  Also, didn't we include replacing '+' with 'plus' in FieldCleaner?  That should make this unnecessary.

bsweger: This is very temporary, since the schema team has already provided a machine-friendly list of field names, which will replace the aliases currently used in the models. Knowing this list was imminent, I put this line in as the quickest patch.

It is not my proudest coding moment, to be sure.

dataactvalidator/interfaces/validatorStagingInterface.py
bsweger: The only place this was being used was in the unit tests. To replace it, I used one of the new methods that get staging records by submission and got the count on it (see https://github.com/fedspendingtransparency/data-act-broker-backend/pull/78/files).

dataactvalidator/interfaces/validatorStagingInterface.py
bsweger: That's a good point. Putting the `commit` here essentially breaks the recommended practice of [keeping session lifecycles separate from functions that access/manipulate data](http://docs.sqlalchemy.org/en/latest/orm/session_basics.html).

Can you say more about what you mean by batch inserting? Does that mean committing a transaction after some number of inserts and then starting a new one? Trying to map my mental model of databases to the validator parlance.

One easy thing to do would be to move the `session.commit()` to `threadedValidateJob`:
```
if jobType == csv_record_validation:
    runValidation(job, interface)
    session.commit()
elif jobType = validation:
    runCrossValidation(job, interface)
    session.commit()
```

Alternately, we could do shorter transactions by doing a `session.commit()` down in the validation routines every _n_ records.

I'm inclined to try the first option, since nothing else (in theory) should be trying to update a submission's records in a concurrent transaction.

No matter what, we definitely need to run some large files through the validator (will put in a story for that). At this point, we're just speculating what the bottlenecks might be. The indexes will also slow down our insert performance.

bsweger: Ok, I just went to make the change to wrap things in transactions and realized that the model interfaces use their own sessions, which makes it hard to control transactions/logical units for work in the application.

At this point, I'm hesitant to add application code for something that database session handling would provide for free. I could pass the db session to this insert function, ensuring that everything is associated with the validator's staging db session. However, I realize that's not how we usually use the interfaces.

jworcestBAH: That is what I mean by batch inserting, in the code being replaced I believe we were doing an insert after every 100 lines.  I assume sqlalchemy is smart enough to put multiple adds into a single insert statement at commit time, but I'm not certain of this, we should check on that.  I believe we were doing it in raw sql before.  My concern about waiting until the end of the file is that breaks our current pattern of never trying to hold the entire file in memory.  If we do n lines at a time we can control how much memory we're using.  For the story for large files, DB-35 is a story for that that's been in there since pretty much the beginning of the project.

jworcestBAH: I'm confused by the session comments because this function is in the staging interface, so it has access to the session.  Model objects have not had independent sessions for quite a while now, unless we messed up somewhere.  Can you clarify where you're seeing a session for a model?

dataactvalidator/interfaces/validatorStagingInterface.py
```diff
@@ -1,77 +1,52 @@
...
+        elif fileType == "award_financial":
+            rec = AwardFinancial(**record)
+        elif fileType == "appropriations":
+            rec = Appropriation(**record)
+        elif fileType == "program_activity":
+            rec = ObjectClassProgramActivity(**record)
+        self.session.add(rec)
+        self.session.commit()
+
+    def getSubmissionRecordsByFileType(self, submissionId, fileType):
```
jworcestBAH: Could we change the name of this function to be clearer that it's returning a query?  I was surprised in the validator code to see that this wasn't actually returning the records.  Maybe "getQueryBySubmissionAndFileType", or something better?

jworcestBAH:
Finished initial review, have not yet looked through responses.

dataactvalidator/interfaces/validatorStagingInterface.py
bsweger: @jworcestBAH You are right about the staging interface and the session handling, and I should stop doing this stuff late at night. I had tested my commit changes in the threaded validation manager, which wasn't being hit by the unit tests.

Just pushed a change to move the commit() out of the interface and up to the validation manager. I currently have it set to commit once at the end of the file.

To your point about chunking, I think we have fundamentally differing philosophies. I'm inclined not complicate the code base by putting in our own chunking logic until it becomes apparent that our AWS and database resources need us to. If we need to code it ourselves, I'd prefer to address when we hit DB-35.

I hope this doesn't come across as me being lazy. Your original point about burying the commit in the interface update function was spot on, and I'm glad you raised it. My goal is to keep the code base as simple as possible.

dataactvalidator/interfaces/validatorStagingInterface.py
bsweger: Would simply calling it `getSubmissionsByFileType` address your concern (_i.e._, just remove the word `Records`?

bsweger:
@jworcestBAH thanks for thorough review, as always. I'm done with the responses.

dataactvalidator/interfaces/validatorStagingInterface.py
bsweger: Now that we've had the SQL validation convo, what do you think about moving the commit here, so we commit to staging before the final batch of error records gets written? Then we could slot in the call to the SQL validator between these two lines.

If this is going to require more thought, I'll leave things as is, and we can change when working on the SQL stories.

dataactvalidator/interfaces/validatorStagingInterface.py
jworcestBAH: Yes not having 'records' in the name would definitely help.

jworcestBAH: This seems to be the only thing remaining this PR is waiting on, so once this change is made I can merge the PR.

bsweger:
@jworcestBAH Thanks for the ping--I made that last change and fixed up the alembic chain. Since this PR has been a WIP for so long, I'll do the staging tables move in a separate PR related to the SQL validation work. Thanks again for the thorough review!

jworcestBAH:
Okay, merging now

PR #79, opened 2016-06-07
Db 584 sf 133

jworcestBAH:
Loading table for SF-133.  Questions still remain about the fields being provided, this is just the code to load the table.


dataactcore/config_example.yml
```diff
@@ -68,6 +68,10 @@ broker:
     # AWS credentials for uploading files. Ignored if use_aws is false.
     aws_create_temp_credentials: true
 
+    # S3 filenames for SF-133 file, only required if planning to load SF-133 table
+    sf_133_folder: config
+    sf_133_file_one: sf_133_part1.csv
+    sf_133_file_two: sf_133_part2.csv
```
bsweger: Apologies for not providing any context re: the two SF-133 files. They sent two .xls files b/c the data exceeded the row limit. Now that they're in .csv form, should we just combine them into a single file for simplicity?

This method of passing around SF-133 data is temporary anyway, so I don't anticipate that we'd have to regularly combine two files to get the entire dataset.

dataactcore/migrations/versions/177d90a225dc_sf_133_amount_text.py
```diff
@@ -0,0 +1,88 @@
+"""sf_133_amount_text
```
bsweger: Is this script doing anything? How odd, Alembic.

dataactcore/models/domainModels.py
```diff
@@ -27,6 +28,27 @@ class ObjectClass(Base):
...
+    availabilitytypecode = Column(Text)
+    beginningperiodofavailability = Column(Text)
+    endingperiodofavailability = Column(Text)
+    mainaccountcode = Column(Text, nullable=False)
+    subaccountcode = Column(Text, nullable=False)
+    tas = Column(Text, nullable=False, default=concatTas, onupdate=concatTas, index=True)
+    fiscal_year = Column(Text)
+    period = Column(Text)
+    line = Column(Integer,nullable=False)
+    amount = Column(Text,nullable=False,default=0,server_default="0")
```
bsweger: should amount be a decimal/numeric field?

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -10,15 +11,39 @@ def loadCgac(filename):
...
+
 def loadProgramActivity(filename):
     LoaderUtils.loadCsv(filename, ProgramActivity, ValidatorValidationInterface(), {"year":"budget_year","agency_id":"agency_id",
         "alloc_id":"allocation_transfer_id","account":"account_number","pa_code":"program_activity_code","pa_name":"program_activity_name"},
         {"program_activity_code":{"pad_to_length":4},"agency_id":{"pad_to_length":3},"account_number":{"pad_to_length":4},"allocation_transfer_id":{"pad_to_length":3}})
 
-def loadDomainValues(basePath):
+
+def loadDomainValues(basePath, localSFPath = None):
+    """ Load all domain value files, localSFPath is used to point to a SF-133 file, if not provided it will be downloaded from S3  """
```
bsweger: What do you think about simplifying this logic to:

1. put a small SF-133 sample into the config folder with the other domain values
2. add that filename to .gitignore so one of us doesn't accidentally commit a full version of it
3. when loading, just use the S3 bucket if the AWS flag is true; otherwise use what's in the config folder.

None of this SF-133 business is ideal, but would love to keep the domain stuff grouped together.

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -10,15 +11,39 @@ def loadCgac(filename):
...
+    # SF 133 is kept on S3, so need to download that
+    reader = CsvS3Reader()
+
+    if localSFPath is not None:
+        # Load SF 133 from same path
+        loadSF133(localSFPath)
+    else:
+        # Download files if using aws, if not they will need to already be in broker_files location
+        if(CONFIG_BROKER["use_aws"]):
+            print("Pulling values from config: " + str(CONFIG_BROKER["aws_region"]) + " " + str(CONFIG_BROKER["aws_bucket"]) + " " + str(CONFIG_BROKER["sf_133_folder"]) + " " + str(CONFIG_BROKER["sf_133_file_one"]))
```
bsweger: Did you want to leave this print statement in (not sure how terrible it is to print out the AWS info from the config file)?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -24,7 +24,7 @@ def compareRecords (recordA,recordB, fields) :
         return True
 
     @classmethod
-    def loadCsv(cls,filename,model,interface,fieldMap,fieldOptions):
+    def loadCsv(cls,filename,model,interface,fieldMap,fieldOptions,skipClear = False):
```
bsweger: If `skipClear` is here just for the multiple files, then +1 to the earlier comment about pre-combining them and not building special logic into the file load process.

bsweger:
@jworcestBAH Done reviewing--thanks for going through this process and surfacing some important questions about the data.

dataactcore/migrations/versions/177d90a225dc_sf_133_amount_text.py
jworcestBAH: Whoops, thought I had removed this.  Alembic autogenerate didn't notice the change in column type, so I just changed the migration that creates the table to create as text.

dataactvalidator/filestreaming/loadFile.py
jworcestBAH: I would still like the unit tests to use the small sample file regardless rather than having to download the whole thing to run unit tests.  With the current structure, I can use localSFPath when calling from the unit test, but initialize can still load the full one whether it's coming from S3 or the broker_files path.  To clarify, what currently happens if the use_aws flag is false is that it tries to load the SF-133 file from the broker_files location.

dataactvalidator/filestreaming/loadFile.py
jworcestBAH: Thanks, that was a debug statement I missed.  It's been removed now.

dataactcore/config_example.yml
jworcestBAH: I thought about combining them, but don't many people (myself included) generally open csv files in Excel?  I'll try combining and see if that still breaks Excel in csv.

bsweger: Oh, it could certainly break Excel. But since the intent is to machine-load into the database, I don't think we should optimize for the use case that a human could open large .csv files in our S3 bucket.

dataactvalidator/filestreaming/loadFile.py
bsweger: I guess that last sentence is what I'm questioning. Our current model is to load domain values from the validator's config file, which we can't do for the full version because S3. From my perspective as a local user, I'd rather the broker look in the config file for a local version because that's where the other domain files are located.

Does that make sense?

dataactcore/config_example.yml
jworcestBAH: Okay, merged into single csv and removed skipClear option.

dataactvalidator/filestreaming/loaderUtils.py
jworcestBAH: Removed this

dataactcore/models/domainModels.py
jworcestBAH: Changed back to Numeric, and removing commas and converting empty to 0 to fit data into Numeric.

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -73,8 +70,17 @@ def loadCsv(cls,filename,model,interface,fieldMap,fieldOptions,skipClear = False
...
                 for field in fieldOptions:
+                    if row[field] is None:
+                        # If field is empty set to an empty string
+                        row[field] = ""
                     # For each field with options present, modify according to those options
                     options = fieldOptions[field]
+                    if "strip_commas" in options:
+                        # Remove commas from numeric fields
+                        row[field] = row[field].replace(",","")
+                        if row[field] == "":
```
bsweger: Not sure about the SF-133 data, but in many cases it's important to distinguish between a 0 value and a NULL value. If an SF-133 dollar amount comes in as empty, do we know the appropriate way to log that?

dataactvalidator/filestreaming/loadFile.py
bsweger: Been thinking about this...to put it another way, as a local user, I routinely clearly out the files in my broker_files folder because that location is used to store outputs. Since the dummy SF-133 is an input, it seems more logical to keep it in the validator/config folder with the other inputs.

dataactvalidator/filestreaming/loaderUtils.py
jworcestBAH: We've been told by schema team to interpret missing values as 0.

jworcestBAH: If this causes an error in calculations, it will prompt the agency to update their data with GTAS/SAM/whatever it's called, which is kind of the point.

bsweger: Thanks for confirming that the schema team gave clarification on this. I'll add it to our questions list, too, so we have that info documented.

dataactvalidator/filestreaming/loadFile.py
jworcestBAH: Okay, we are now storing the normal file in config, and it's in the gitignore.  The unit test is still overriding that with a small sample file from the tests folder, which is the same approach we're using for TAS files.

jworcestBAH:
I believe I've addressed everything now, please let me know if I missed something.

bsweger:
Thanks, @jworcestBAH for the back-and-forth on this. Although the requirement that the SF-133 data not be included in the repo is not ideal for our workflow or for users who want to run the broker locally, I think we landed in a reasonable place.

Glad we tackling the SF-133 file sooner rather than later!

PR #81, opened 2016-06-08
Removing permissions check for list_agencies route

nmonga91:


bsweger:
Looks good, @nmonga91. Will merge once Jenkins gives the 👍

PR #82, opened 2016-06-08
Users can now view/edit submissions if they're within that agency OR …

nmonga91:
…they are the owner


jworcestBAH:
Looks good, merging.

PR #84, opened 2016-06-09
add sql rule table

bsweger:
This PR is for a new table to house SQL-based validation rules.

Wanted to get this into the code base so @jworcestBAH can use it for testing.


bsweger:
@jworcestBAH based on the convo we're having about cross-file errors on the front end, does it make sense to also include a cross file boolean field in this table?

jworcestBAH:
Yes, that would definitely be helpful, alternatively we could have a file id for "cross file", that could be problematic somewhere or it could simplify several places, I'm not really sure.  Probably safer to have the separate flag.

anthonygarvan:
Hey one thing that might be helpful is (if you haven't done this already), write out some examples of error messages and make sure the static format I think you're envisioning here would work. In my previous work it was very helpful to pull out data from the error rows themselves, so that you could format the error message with data from the error rows, like `Error in row 1167: value "sp" in column "org_code" is not a valid org code.` Probably doesn't belong in this PR but just something to keep in mind for the future.

bsweger:
@jworcestBAH I added the cross file flag and renamed the warning flag for a consistent naming convension.

@anthonygarvan Thanks for weighing in here. The error messages that go back to users are made up of several parts, one of which is the generic message stored with each rule in this table. When we're processing the failed records returned from the db, we augment the error report with record-level info like row number, offending column names and values, etc.

Is that similar to what you've done in the past?

bsweger:
@jworcestBAH just pushed a few changes:
* change `rule_number` to `rule_label`
* change `rule_crossfile_flag` to `rule_cross_file_flag`
* add `rule_severity` table and an FK reference to it `rule_sql` (replaces the critical rule flag)

When I submit the follow-up PR to populate the rules table, I'll also include a script that does the initial load of `rule_severity`. Wanted to get the skeleton tables out there so you can test with them.

Thanks for the feedback of the table structure--very helpful to have it at this initial stage!

jworcestBAH:
Looks good, merging.

PR #85, opened 2016-06-10
Hotfix tas

jworcestBAH:
This adds a unique index to the TAS table and checking for conflicts with that, should fix the current error.  Will be heading to DC and back for the next couple hours, if there are simple changes feel free to make them before merging, as this is holding up the deploy process.


bsweger:
Pushed a fix to the migration script, which didn't reflect the new index.

nmonga91:
looks good!

bsweger:
@nmonga91 gave the 👍. merging! Thanks for tracking this down, @jworcestBAH

PR #86, opened 2016-06-10
Removing a debug step

jworcestBAH:
Removing the temporary step that was breaking dev


jworcestBAH:
Apparently the TODO flag that makes the comment bright blue was not salient enough for me to remember to remove this.  This was a temporary thing for debugging some weird TAS filtering issues we were having last sprint, and somehow got left in and pushed to dev.

PR #88, opened 2016-06-13
Check status cross file

jworcestBAH:
Check status route now includes info on cross-file errors

### DIFF
--- a/dataactbroker/check_status.yml
+++ b/dataactbroker/check_status.yml
@@ -1,0 +1,127 @@
+swagger: '2.0'
+info:
+  title: Check Status API Route
+  description: check_status route, updated for sprint 15
+  version: '15'
+schemes:
+  - https
+# will be prefixed to all paths
+basePath: /v1
+produces:
+  - application/json
+paths:
+  /check_status:
+    post:
+      description: |
+        A call to this route will provide status information on all jobs associated with the specified submission.
+      tags:
+        - DB-587
+      parameters:
+        - name: X-Session-ID
+          in: header
+          required: true
+          type: string
+          description: Session identifier
+      responses:
+        200:
+          description: List of validation jobs
+          schema:
+            type: object
+            properties:
+              jobs:
+                type: array
+                items:
+                  $ref: '#/definitions/Job'
+              cgac_code:
+                type: string
+              reporting_period_start_date:
+                type: string
+                description: "MM/YYYY or Q#/YYYY"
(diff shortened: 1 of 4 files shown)


dataactbroker/check_status.yml
```diff
@@ -0,0 +1,127 @@
+swagger: '2.0'
```
bsweger: Was the intent to check this into the repo? At some point, it would be great to use Swagger to replace our non-machine-readable API docs, but this looks like the file used to verify the change with the front-end team?

bsweger:
@jworcestBAH done reviewing--just one question about the inclusion of the Swagger YAML.

dataactbroker/check_status.yml
jworcestBAH: I don't think we have a process around swagger yet.  This was what we created to document this route, if there's somewhere else we should put these, I can put it there instead.  I put it here because this is where the current route documentation is in the readme.  Paging @nmonga91, what was done with the swagger from last sprint?

nmonga91: The swagger file from last sprint wasn't stored anywhere. Wasn't sure if we wanted to hold on to it or not or if we just were going to update the README

jworcestBAH: Seems kind of pointless to do the swagger if we're just throwing it out, especially if we can't get it to do examples.

nmonga91: Agreed, swagger is helpful for auto-generation for documentation purposes. But what we needed it for was collaboration during planning. I've found some open source tools that may be a helpful alternative. I can dive into those a little more and see if they're a viable solution.

bsweger: @nmonga91 since we don't yet have a built-in place for exposing the swagger, what do you recommend we do with these files in the meantime? keep them in a working folder somewhere in the repo? stash them somewhere until [DB-395](https://federal-spending-transparency.atlassian.net/browse/DB-395) gets moved into a sprint?

nmonga91: We can leave them as a part of the repo, but they will be overwritten once we auto-generate them using Python's flask. But until then, it's probably a good idea to hold on to them in a top-level directory in the repo

bsweger: Sounds good--we can always move them into a folder if things start to get cluttered.

bsweger:
Looks good--merging!

PR #89, opened 2016-06-15
Db 578 email users route

nmonga91:
@jworcestBAH or @bsweger


dataactbroker/handlers/accountHandler.py
```diff
@@ -447,7 +450,7 @@ def changeStatus(self,system_email):
     def listUsers(self):
         """ List all users ordered by status. Associated request body must have key 'filter_by' """
         user = self.interfaces.userDb.getUserByUID(LoginSession.getName(flaskSession))
-        isAgencyAdmin = True if self.interfaces.userDb.hasPermission(user, "agency_admin") else False
+        isAgencyAdmin = self.userManager.hasPermission(user, "agency_admin") and not self.userManager.hasPermission(user, "website_admin")
```
bsweger: Just to clarify: do we know for sure that an agency admin will never also be a website admin?

dataactbroker/handlers/accountHandler.py
```diff
@@ -717,4 +720,33 @@ def setSkipGuide(self, session):
...
+        if not (requestDict.exists("users") and requestDict.exists("submission_id")):
+            exc = ResponseException("Email users route requires 'users' and 'submission_id'", StatusCode.CLIENT_ERROR)
+            return JsonResponse.error(exc, exc.status)
+
+        uid = session["name"]
+        current_user = self.interfaces.userDb.getUserByUID(uid)
+
+        user_ids = requestDict.getValue("users")
+        submission_id = requestDict.getValue("submission_id")
+        self.jobManager.getSubmissionById(submission_id)
```
bsweger: Good idea to check for a submission id here. If the submission is invalid, does the job handler send the appropriate message back to the front end, and is the front end expecting such a message? (it's hard to imagine the front end sending a bad id to begin with, but want to make sure everything is hooked up)

dataactbroker/handlers/accountHandler.py
```diff
@@ -717,4 +720,33 @@ def setSkipGuide(self, session):
...
+
+        users = []
+
+        link = "".join([AccountHandler.FRONT_END, '#/reviewData/', str(submission_id)])
+        emailTemplate = {'[REV_USER_NAME]': current_user.name, '[REV_URL]': link}
+
+        for user_id in user_ids:
+            users.append(self.userManager.getUserByUID(user_id))
```
bsweger: What happens if this is not a valid user id? Will we send back a reasonable error?

dataactbroker/handlers/accountHandler.py
nmonga91: This accounts for the situation where they are both. If they are both, the website admin permissions should have higher priority and therefore return all users and not just users within their agency.

dataactbroker/handlers/accountHandler.py
nmonga91: Yeah the appropriate message comes back as something like, "No submission with that id"

dataactbroker/handlers/accountHandler.py
nmonga91: getUserByUID will raise an exception if a user isn't found, "No users with that uid"

dataactbroker/scripts/setupEmails.py
```diff
@@ -68,6 +69,10 @@ def setupEmails():
     template = "Your account has been unlocked and requires your password to be reset. Please click the following link <a href='[URL]'>here</a> to start the processs. The link will expire in 24 hours. <br/> <br/> If you did not request your account to be unlocked, please notify the DATA Act Broker Helpdesk (DATABroker@fiscal.treasury.gov) <br /><br />DATA Act Broker Helpdesk<br /><br />DATABroker@fiscal.treasury.gov"
     userDb.loadEmailTemplate("DATA Act Broker - Password Reset", template, "unlock_account")
 
+    #Submission Review
+    template = "[REV_USER_NAME] has shared a DATA Act broker submission with you. Click <a href='[REV_URL]'>here</a> to review their submission. For questions or comments, please email the DATA Act Broker Helpdesk (DATABroker@fiscal.treasury.gov)."
+    userDb.loadEmailTemplate("DATA Act Broker - Notification", template, "review_submission")
```
bsweger: Does it make sense to have a less generic subject line here? Something like "DATA Act Broker - Submission Ready for Review"? Better yet, maybe our content folks have an idea. Have they looked at the text for this new notification?

nmonga91: I got the actual content from @kaitlin, I just did the subject on my own. We can definitely make it more specific if that's what we decide

bsweger: Cool--thanks for clarifying! If @kaitlin provided the text, perhaps she could weigh in on the subject line before we merge.

dataactbroker/handlers/accountHandler.py
bsweger: Yep, I see that when testing the method in Python. Just want to make sure that the front-end can handle that error in this context. Does that make sense?

dataactbroker/handlers/accountHandler.py
bsweger: Does the front-end know how to handle that error in this context?

dataactbroker/handlers/accountHandler.py
bsweger: Thanks for clarifying that!

tests/userTests.py
```diff
@@ -322,4 +322,22 @@ def test_update_user(self):
...
+                 "email_template": "review_submission"}
+        response = self.app.post_json("/v1/email_users/", input, headers={"x-session-id": self.session_id})
+        self.check_response(response, StatusCode.OK, "Emails successfully sent")
+
+        badInput = {"users": [self.agency_user_id]}
+        response = self.app.post_json("/v1/email_users/", badInput, expect_errors=True,
+                                      headers={"x-session-id": self.session_id})
+        self.check_response(response, StatusCode.CLIENT_ERROR)
+
+        badInput = {"users": [self.agency_user_id], "submission_id": -1}
```
bsweger: Thanks for adding this thorough unit test. Do you think we should also test bad input with an invalid e-mail template name?

dataactbroker/handlers/accountHandler.py
```diff
@@ -717,4 +720,39 @@ def setSkipGuide(self, session):
...
+        self.userManager.getEmailTemplate(template_type)
+
+        users = []
+
+        link = "".join([AccountHandler.FRONT_END, '#/reviewData/', str(submission_id)])
+        emailTemplate = {'[REV_USER_NAME]': current_user.name, '[REV_URL]': link}
+
+        for user_id in user_ids:
+            # Check if user id is valid, if so add User object to array
+            users.append(self.userManager.getUserByUID(user_id))
```
bsweger: Looking at [the wireframe](http://pueyiu.axshare.com/#g=1&p=review__certify___publish_submission), I see that a user can free text any e-mail addy in the submission notification box. This code will obviously not send an e-mail to a person who doesn't already have a broker account. Do we need to additional checks before sending the e-mail (_e.g._, user we're e-mailing has an active account, belongs to the same agency, is an agency admin)? 

Or is the plan to rely on the login permissions to handle all of this after the e-mail recipient clicks the link? In other words, the e-mail recipient may receive a link to look at things they aren't authorized to see. Just want to be explicit about how we decide to handle this (or maybe I missed it elsewhere in the code).

bsweger:
@nmonga91 done reviewing!

dataactbroker/handlers/accountHandler.py
nmonga91: The front-end only allows selecting from a list of approved users. I didn't put that check in here in case we wanted to ever override that in any way. If not, we should definitely add that check in there so no one can do a post request to users that aren't actually approved users.

tests/userTests.py
nmonga91: Since we tested for an invalid submission_id, we should also test for invalid users and invalid email templates too

dataactbroker/handlers/accountHandler.py
bsweger: Confirmed with @kevinli-work that the front-end is populated via the `list_users` route, which only returns the appropriate users. Thanks!

bsweger:
Looks good--merging!

PR #90, opened 2016-06-15
Db 667 sql validations

jworcestBAH:
Includes changes to staging tables and basic architecture for running sql-based validations.


jworcestBAH:
Submitting this before rest of sql validation components are complete to make the changes to staging tables available.

jworcestBAH:
Now marking records as valid and having unit tests check number of valid records.

dataactvalidator/interfaces/validatorStagingInterface.py
```diff
@@ -13,17 +15,17 @@ def insertSubmissionRecordByFileType(self, record, fileType):
...
-        if fileType == "award":
-            rec = AwardFinancialAssistance(**record)
-        elif fileType == "award_financial":
-            rec = AwardFinancial(**record)
-        elif fileType == "appropriations":
-            rec = Appropriation(**record)
-        elif fileType == "program_activity":
-            rec = ObjectClassProgramActivity(**record)
+
+        rec = self.getModel(fileType)(**record)
```
bsweger: Thanks for streamlining this!

dataactvalidator/validation_handlers/validator.py
```diff
@@ -782,3 +789,27 @@ def validateTAS(fieldsToCheck, tasFields, record, interfaces, fileType):
         else:
             # Multiple instances of same TAS, something is going wrong
             raise ResponseException("TAS check is malfunctioning",StatusCode.INTERNAL_ERROR)
+
+    @staticmethod
+    def validateFileBySql(submissionId,fileType,interfaces):
+        # Pull all SQL rules for this file type
+        fileId = interfaces.validationDb.getFileId(fileType)
+        rules = interfaces.validationDb.session.query(RuleSql).filter(RuleSql.file_id == fileId).all()
```
bsweger: We probably also want to filter for cross_file_flag = False here. In my initial stab at the sql validation table, I added fileType in the cross file rules A18 and A19 to indicate the target file being validated, and we don't want those rules to be returned in this query.

dataactvalidator/validation_handlers/validator.py
```diff
@@ -143,9 +144,13 @@ def validate(record,rules,csvSchema,fileType,interfaces):
         fileType -- name of file type to check against
 
         Returns:
-        True if validation passed, False if failed, and list of failed rules, each with field, description of failure, and value that failed
+        Tuple of three values:
```
bsweger: Thanks for updating the docstring...so easy to forget that.

dataactvalidator/validation_handlers/validator.py
```diff
@@ -782,3 +789,27 @@ def validateTAS(fieldsToCheck, tasFields, record, interfaces, fileType):
...
+        fileId = interfaces.validationDb.getFileId(fileType)
+        rules = interfaces.validationDb.session.query(RuleSql).filter(RuleSql.file_id == fileId).all()
+        errors = []
+        # For each rule, execute sql for rule
+        for rule in rules:
+            failures = interfaces.validationDb.connection.execute(rule.rule_sql.format(submissionId))
+            # Build error list
+            for failure in failures:
+                row = failure["row"]
+                errorMsg = rule.rule_description
```
bsweger: The sql rule table has a separate `rule_error_message` column that I had envisioned as the error message. This would be to handle cases where we've split a rule into multiple SQL statements (B11 for example...one error message would about an incorrect objectclass, and one message would be about an invalid objectclass prefix).

What do you think about that approach?

bsweger:
@jworcestBAH Done reviewing.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Good point, now filtering out cross file rules.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: That makes sense, now using rule_error_msg.  Would rule_description still be used for anything?  Even if not, I guess it might still make sense to keep that just so the original description is present somewhere.

bsweger:
Looks good! Merging...

PR #91, opened 2016-06-16
load sql rules and rule severity

bsweger:
This PR:

* adds function to populate the new rule severity when initializing the validator
* populates the new sql rule table from a spreadsheet in the validator config file

*Note*:
Coming next is a PR to move the staging tables to the validation database. As a result, this round of SQL rules doesn't contain any that involve cross-database joins.

This is the MVP for getting something out there for the SQL rule tests to proceed. Going forward, we might want to also think about a better place to store the sql code that we're loading (i.e., not in a csv file)


dataactvalidator/config/sqlRules.csv
```diff
@@ -0,0 +1,4 @@
...
... ussgl498100_upadjsprioryeardeliveredordersobligunpaid_cpe IS NOT NULL OR ussgl480200_undeliveredordersobligationsprepaidadvanced_fyb IS NOT NULL OR ussgl480200_undeliveredordersobligationsprepaidadv_cpe IS NOT NULL OR ussgl483200_undeliveredordersobligtransferredppdadv_cpe IS NOT NULL OR ussgl488200_upadjsprioryrundelivordersobligprepaidadv_cpe IS NOT NULL OR ussgl490200_deliveredordersobligationspaid_cpe IS NOT NULL OR ussgl490800_authorityoutlayednotyetdisbursed_fyb IS NOT NULL OR ussgl490800_authorityoutlayednotyetdisbursed_cpe IS NOT NULL OR ussgl498200_upadjsprioryrdelivordersobligpaid_cpe IS NOT NULL) AND COALESCE(bydirectreimbursablefundingsource,'') NOT IN ('R', 'D')",B12,"Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".","Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".",,award_financial,fatal
```
jworcestBAH: I thought we had decided to just use the submission ID in these queries so we have a consistent way of using format().   We're now removing records that match submission ID and filetype at the start of the job to avoid conflicts from this.

jworcestBAH: For the cross file flag, is a blank entry the same thing as false?  Would it be better to put that in explicitly rather than interpret a blank entry?

dataactvalidator/config/sqlRules.csv
```diff
@@ -0,0 +1,4 @@
...
+"SELECT approp.row, approp.tas, approp.grossoutlayamountbytas_cpe, SUM(op.grossoutlayamountbyprogramobjectclass_cpe) as grossoutlayamountbyprogramobjectclass_cpe_total FROM appropriation approp JOIN object_class_program_activity op ON approp.tas = op.tas WHERE approp.submission_id = {} GROUP BY approp.row, approp.tas, approp.grossoutlayamountbytas_cpe HAVING approp.grossoutlayamountbytas_cpe <> SUM(op.grossoutlayamountbyprogramobjectclass_cpe)",A18,"GrossOutlayAmountByTAS_CPE (file A) = sum of all GrossOutlayAmountByProgramObjectClass (file B) {When Program Activity and Object Class are not blank, this value is the sum of all Gross Outlay Amounts reported in file B, by TAS/Subaccount}",The GrossOutlayAmountByTAS_CPE amount in the appropriation file (A) does not equal the sum of the corresponding GrossOutlayAmountByProgramObjectClass_CPE values in the award financial file (B).,y,appropriations,warning
```
jworcestBAH: It may take me a while to understand this one, my first question is what is op?  Is that supposed to represent file B?

dataactvalidator/config/sqlRules.csv
bsweger: Ah, ok--I misunderstood. Yeah, as long as pulling everything by submission ID won't cause any issues, I'll switch that back.

dataactvalidator/config/sqlRules.csv
bsweger: @jworcestBAH Yes, that's my shorthand for File B--you can see it being defined in the `FROM` clause: `JOIN object_class_program_activity op` 

You raise a good point--we should decide on table aliases when writing queries and use those consistently. (We need to alias the tables in the SQL queries to disambiguate columns with the same name).

jworcestBAH: Yes those are fine.

bsweger: @kaitlin showed me a good website that could help with the SQL code review in the short term: http://sqlformat.org/. I'm a little hesitant to embed comments right in the sql string b/c I haven't tested how those behave when passed to the db engine via the ORM.

bsweger:
```sql
SELECT af.row_number,
       af.bydirectreimbursablefundingsource
FROM award_financial AS af
WHERE af.submission_id = {}
-- check to see if any there are any reports of obligation or outlay USSGL balances (excluding downward adjustments)
  AND (ussgl480100_undeliveredordersobligationsunpaid_fyb IS NOT NULL
       OR ussgl480100_undeliveredordersobligationsunpaid_cpe IS NOT NULL
       OR ussgl483100_undeliveredordersobligtransferredunpaid_cpe IS NOT NULL
       OR ussgl488100_upadjsprioryearundelivordersobligunpaid_cpe IS NOT NULL
       OR ussgl490100_deliveredordersobligationsunpaid_fyb IS NOT NULL
       OR ussgl490100_deliveredordersobligationsunpaid_cpe IS NOT NULL
       OR ussgl493100_deliveredordersobligstransferredunpaid_cpe IS NOT NULL
       OR ussgl498100_upadjsprioryeardeliveredordersobligunpaid_cpe IS NOT NULL
       OR ussgl480200_undeliveredordersobligationsprepaidadvanced_fyb IS NOT NULL
       OR ussgl480200_undeliveredordersobligationsprepaidadv_cpe IS NOT NULL
       OR ussgl483200_undeliveredordersobligtransferredppdadv_cpe IS NOT NULL
       OR ussgl488200_upadjsprioryrundelivordersobligprepaidadv_cpe IS NOT NULL
       OR ussgl490200_deliveredordersobligationspaid_cpe IS NOT NULL
       OR ussgl490800_authorityoutlayednotyetdisbursed_fyb IS NOT NULL
       OR ussgl490800_authorityoutlayednotyetdisbursed_cpe IS NOT NULL
       OR ussgl498200_upadjsprioryrdelivordersobligpaid_cpe IS NOT NULL)
-- if there is a qualifying balance being reported, make sure direct reimbursable funding source is D or R
  AND COALESCE(bydirectreimbursablefundingsource,'') NOT IN ('R',
                                                             'D')
```

bsweger:
```sql
-- rule A18
SELECT approp.row_number,
       approp.tas,
       approp.grossoutlayamountbytas_cpe,
      -- next line is summing values in the object_class_program_activity_table 
      -- so we can return that value in the query result
       SUM(op.grossoutlayamountbyprogramobjectclass_cpe) AS grossoutlayamountbyprogramobjectclass_cpe_total
-- next 2 lines join appropriations and object_class_program_activity by TAS
FROM appropriation approp
JOIN object_class_program_activity op ON approp.tas = op.tas
WHERE approp.submission_id = {}
GROUP BY approp.row,
         approp.tas,
         approp.grossoutlayamountbytas_cpe
-- HAVING clause compares the summed value from the object_class_program_activity table 
-- to its counterpart in the appropriations table. 
-- We're interested in the cases where those two are not equal.
HAVING approp.grossoutlayamountbytas_cpe <> SUM(op.grossoutlayamountbyprogramobjectclass_cpe)
```

bsweger:
```SQL
-- rule A19
SELECT approp.row_number,
       approp.tas,
       approp.obligationsincurredtotalbytas_cpe,
       -- next line is summing values in the object_class_program_activity_table 
      -- so we can return that value in the query result
       SUM(op.obligationsincurredbyprogramobjectclass_cpe) AS obligationsincurredbyprogramobjectclass_cpe_sum
FROM appropriation approp
-- next 2 lines join appropriations and object_class_program_activity by TAS
JOIN object_class_program_activity op ON approp.tas = op.tas
WHERE approp.submission_id = {}
GROUP BY approp.row,
         approp.tas,
         approp.obligationsincurredtotalbytas_cpe
-- HAVING clause compares the summed value from the object_class_program_activity table 
-- to its counterpart in the appropriations table. 
-- We're interested in the cases where those two are not equal.
HAVING approp.obligationsincurredtotalbytas_cpe <> SUM(op.obligationsincurredbyprogramobjectclass_cpe)
```

jworcestBAH: For rule A18, grouping by TAS is clear since that's what the rule is checking, but I don't understand why we're grouping on row number and grossoutlayamount.  Since these are both coming from approp and there is only one line per TAS in the approp file, it seems like these would have no effect, but I'd like to understand the intent behind them.  I also still have some confusion over the sum that I'd like to discuss in slack.

jworcestBAH:
Looks good, merging

dataactvalidator/config/sqlRules.csv
bsweger: @jworcestBAH Sorry, this question got lost in my github notifications when you posted it! 

For posterity, we did discuss the query via Slack. In short, the we're including non-TAS fields from the appropriation file in the `SELECT` and `GROUP BY` statements b/c we need them for the error report. Because the schema team has told us to expect only one record per tas in the appropriation file (which we'll validate during individual file validations), we can safely pull back any columns and aggregate on them.

PR #92, opened 2016-06-16
Program activity rule now skipped if both fields are blank

jworcestBAH:
Program activity rule is only applied if at least one of the two fields is populated


bsweger:
Looks good, merging!

PR #94, opened 2016-06-17
DB 657: move staging tables to validation database

bsweger:
This PR lays the foundation for being able to write SQL validations that check domain values. To do that w/o dblink or other methods for cross database joins, we'll store the staging tables in the same physical database (i.e., validation) as the rest of the validation data.

I followed the pattern of the domain value tables, where we maintain the models in a separate file that's imported into the validation models. For the interfaces, I kept the `validationStagingInterface.py` intact since conceptually, the staging tables are a discrete thing we work with. That interface now inherits from `validationInterface` instead of `stagingInterface`, which has been deleted.

Alembic migrations now ignore the staging database entirely, so old migration code that references it doesn't impact upgrades and downgrades.

The PR also renames the `row` column of the staging tables to `row_number` since `row` is a reserved word in many databases (and `rownum` is an Oracle reserved word).


dataactcore/models/base.py
```diff
@@ -0,0 +1,4 @@
+from sqlalchemy.ext.declarative import declarative_base
+from dataactcore.utils.timeStampMixin import TimeStampBase
+
+Base = declarative_base(cls=TimeStampBase)
```
jworcestBAH: What is the purpose of this file?  I believe we had different base objects for different database to keep metadata objects separate for sqlalchemy, combining them could be problematic.  If the goal here is just to use this for the files in the validation database, could we give it a more specific name like validationBase.py?

dataactvalidator/README.md
```diff
@@ -35,7 +35,7 @@ The validation process begins with a call to one of the above routes from either
...
 
-Each record is checked against the set of rules for the fields present in the file.  If a record passes all rules, it goes into the staging database.  Otherwise, each failed rule goes into the error report, and a running sum is kept of error occurrences for each rule, which are written to the error database after all rows have been checked.  Finally, the job is marked as finished in the job tracker, the file is marked completed in the error database.
+Each record is checked against the set of rules for the fields present in the file.  With the exception of records that fail a data type check, each record goes into a staging table in the validation database. Data that fails any validation rule is written to the error report, and a running sum is kept of error occurrences for each rule, which are written to the error database after all rows have been checked.  Finally, the job is marked as finished in the job tracker, the file is marked completed in the error database.
```
jworcestBAH: Thanks for cleaning up the documentation for this, it's good to keep that up to date.

dataactvalidator/interfaces/validatorJobTrackerInterface.py
```diff
@@ -63,22 +63,6 @@ def checkPrerequisites(self, jobId):
...
-        Returns:
-            True if successful
-        """
-        query = self.session.query(Job).filter(Job.job_id == jobId)
-        result = self.checkJobUnique(query)
-        result.staging_table = stagingTable
-        self.session.commit()
-        return True
-
     def checkJobType(self,jobId):
```
jworcestBAH: Good catch removing this, I just pushed a commit to remove the staging_table field from the job table as that is no longer needed.

dataactcore/models/base.py
bsweger: Yep, it's specifically for the validation database--I needed to extract out the Base stuff now that we're importing more than one file into validationModels (stagingModels now, in addition to domainModels). 

Good call on being more specific in the name--just pushed a change for that.

dataactvalidator/interfaces/validatorJobTrackerInterface.py
bsweger: Thanks!

jworcestBAH:
Okay, merging now.

PR #96, opened 2016-06-17
Fixed check of empty fields in program activity rule

jworcestBAH:
Also updated sample file to have whitespace present


bsweger:
Looks good-merging!

bsweger:
For real merging this time.

PR #97, opened 2016-06-17
Fixed check of empty fields in program activity rule

jworcestBAH:
Applying parenthetical correction to dev.


bsweger:
Looks good--merging!

PR #100, opened 2016-06-20
Db 664 original label

jworcestBAH:
Original rule labels now included in error reports and error metadata


dataactvalidator/config/awardFinancialRules.csv
```diff
@@ -1,27 +1,27 @@
...
+rule_type,is_single_field,field_name,rule_text_one,rule_text_two,description,rule_label,rule_timing,original_label
+car_match,FALSE,,"allocationtransferagencyidentifier, agencyidentifier, beginningperiodofavailability, endingperiodofavailability, availabilitytypecode, mainaccountcode, subaccountcode","allocation_transfer_agency, agency_identifier, beginning_period_of_availability, ending_period_of_availability, availability_type_code, main_account_code, sub_account_code",TAS fields must match CARS file,A1_C,multi_field,A1
+REQUIRE_ONE_OF_SET,FALSE,,"piid, fain, uri",,"Must have either a piid, fain, or uri",NONE,multi_field,
```
bsweger: is there a rule number for this one?

dataactbroker/README.md
```diff
@@ -616,14 +618,16 @@ Example output:
           "error_name": "rule_failed",
           "field_name": "award_financial",
           "occurrences": "11",
-          "rule_failed": "fain, uri"
+          "rule_failed": "fain, uri",
+          "original_label":""
```
bsweger: Is this sample output for the cross file validation errors? Would we expect there to be a value for `original_label` here?

bsweger:
@jworcestBAH done reviewing!

dataactbroker/README.md
jworcestBAH: That particular rule does not have a label in the current rule definition, as it originally came in through a JIRA ticket and not through the practices and procedures document.  That said, C14 is now present and is similar but not quite the same rule since it doesn't include piid.  That said, since we've fixed the bug on cross-file rule_failed, I'll fix the text on that.

bsweger: Thanks for the context. Do we have a canonical list of validations we've implemented that aren't explicitly expressed in the practices and procedures doc? Since we've started the weekly check-ins, I've been trying to log things there. 

Unrelated to this particular PR, but I want to make sure we're not doing validations that aren't (or won't be) explicitly written out by the schema team, since those docs are what agencies will use to prep their data.

jworcestBAH:
Added a blank label to length rules to avoid inconsistency in error reports with 'None' showing up just for length checks.

dataactvalidator/config/awardFinancialRules.csv
jworcestBAH: Labeling this as C14, will add the difference here to the spreadsheet for discussion with schema team.

jworcestBAH: Just talked with the schema team, they're agreed that piid should be part of this rule, we'll discuss in the next meeting if it should be "at least one" or "exactly one".

dataactbroker/README.md
jworcestBAH: This was the only rule we had that didn't correspond to a rule in prac & proc, and it's now linked to C14.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -325,6 +325,7 @@ def runCrossValidation(self, jobId, interfaces):
                             self.crossFileReportHeaders) as writer:
             for failure in failures:
                 writer.write(failure)
+                print("failure: " + str(failure) + " length: " + str(len(failure)))
```
bsweger: Is this a debug statement we should take out before merging?

jworcestBAH: Good catch, removed this.

jworcestBAH:
Okay, normal rule_label is now being passed through into the error reporting, and the discussed query_name has been added.

bsweger:
@jworcestBAH changes look good--thanks! Looks like Jenkins is upset, though.

jworcestBAH:
Jenkins is passing now, just needed another merge from dev

bsweger:
Thanks for taking a look--merging!

PR #101, opened 2016-06-20
Skipping "external_validation" jobs when looking up submission status…

nmonga91:
…es. Fixing mistyped reference.

@jworcestBAH or @bsweger


bsweger:
@nmonga91 Good catch--merging!

PR #102, opened 2016-06-20
add sql rules for B11 (object class checks)

bsweger:
Two new entries in the SQL rule table to implement RSS rule B11:
1. check to make sure `objectclass` in award_financial file is valid
2. if `objectclass` is 4 digits, do some additional checking on the first digit

Since we don't have the SQL loader written yet, will follow-up with comments that show formatted versions of the SQL code w/ comments.


bsweger:
First check (valid object class):
```sql
SELECT af.row_number, af.objectclass
FROM award_financial AS af
WHERE af.submission_id = {}
-- missing (i.e., NULL) object class codes will not be returned here
-- which should be ok because a previous process already checks
-- for required fields that don't have values
AND af.objectclass NOT IN (
	SELECT object_class_code
	FROM object_class)
```

2nd check, if the objectclass is 4 digits
```sql
SELECT af.row_number, af.objectclass, af.bydirectreimbursablefundingsource
FROM award_financial af
WHERE af.submission_id = {}
AND LENGTH(af.objectclass) = 4
-- note lower() function on bydirectreimbursablefundingsource since we
-- are doing case-insensitive validations
AND (
    (LEFT(af.objectclass,1) = '1'
    AND lower(bydirectreimbursablefundingsource) IS DISTINCT FROM 'd')
OR 
    (LEFT(af.objectclass,1) = '2' 
    AND lower(af.bydirectreimbursablefundingsource) IS DISTINCT FROM 'r')
)
```

jworcestBAH:
I'd like to get the fixes to sql validation merged into this branch so we can make sure this is passing unit tests, that PR is just waiting on changes to B12.

dataactvalidator/config/sqlRules.csv
```diff
@@ -1,7 +1,14 @@
...
+"SELECT af.row_number, af.objectclass, af.bydirectreimbursablefundingsource FROM award_financial af WHERE af.submission_id = {} AND LENGTH(af.objectclass) = 4 AND ((LEFT(af.objectclass,1) = '1' AND lower(bydirectreimbursablefundingsource) IS DISTINCT FROM 'd') OR (LEFT(af.objectclass,1) = '2' AND lower(af.bydirectreimbursablefundingsource) IS DISTINCT FROM 'r'))",B11,"Must be valid 3-digit object class as defined in OMB Circular A-11 Section 83.6, or a 4-digit code which includes a 1-digit prefix that distinguishes direct, reimbursable, and allocation obligations. Do not include decimal points when reporting in the Schema.","Failed rule: If object class is 4 digits, check that 1 digit prefix is consistent with value for ByDirectReimbursableFundingSource",n,award_financial,warning
```
jworcestBAH: I'm guessing one of the pairs for B11 is supposed to be for program activity instead of both being award financial?  This should cut the 8 extra errors in half, leaving 4 unexpected.  Given that this doesn't seem to include a removal of the existing B11 rule, removing those 6 errors will actually leave us two short.  One of those is because the prefix rule is passing anything other than '1' or '2', so 'ABCD' passes the prefix rule.  The other is because this rule is exactly 4 whereas the old implementation was > 3, so '25412' doesn't get checked.  Only checking for exactly 4 makes sense, so we can drop the expected errors by one, I'm not sure what the behavior should be for unexpected prefixes, do we show both errors or just the error of not being a valid object class (until there's some weird exception).

bsweger: Yep, thanks for catching that--it was a cut-and-paste-merge-conflict-resolution gone awry. Also, I had a commit somewhere that removed the existing B11. Will find out what happened to that and put it back.

Also, thank you very much for digging into the unittest differences. Re: the prefix question, the rule as written assumes that 4 digit objectclasses with a prefix other than 1 or 2 will generate errors in the SQL rule that validates against the objectclass domain value list. That assumption holds true for the current set of domain values (the only 4 digit objectclasses on that list begin w/ 1 and 2).

So the question is...if we get an `ABCD` objectclass, is it enough to return an invalid object class message, or do we also need to tell users about an invalid prefix?

bsweger:
After fixing up the rule file dupes, removing the old B11 rule entries, and

`test_award_fin_mixed`:
* `ABCD` objectclass wasn't showing up as a B11 error at all because that record failed type checks in other fields and never got loaded to the staging table (hence, didn't get processed by the query). Added `ABCD` to another test record to make sure we hit this condition and updated the expected valid staging records accordingly.
* Revised the number of expected errors down by 2, per previous convo (I realize we don't have full resolution, but I feel safe b/c our current list of 4-digit object classes all begin with 1 and 2)

`test_program_mixed` had 3 fewer errors than expected
* There is one less invalid objectclass being reported (`X`) because the row it's has type errors and doesn't get loaded to the staging table for SQL rule processing
* There are two fewer prefix errors being reported. One (`25.401`) is due to a length > 4, so the SQL isn't checking that. The other `3255` is the scenario we discussed earlier--it's not reported as an invalid prefix but does show up on the error report as an invalid object class.

I revised that test to expect 3 fewer errors.

bsweger:
@jworcestBAH one last comment about the error reports--I noticed that some fieldnames previously reported as `MultiField` now contain a list of specific column names (e.g., `objectclass, bydirectreimbursablefundingsource`). I actually like this newfound specificity, but wanted to check in and make sure that won't cause any issues downstream.

bsweger:
@jworcestBAH I had an aha moment about the SQL rules for A18 and A19. Although those are technically unrelated to this PR, I pushed a fix for those here to avoid a future merge conflict in sqlRules.csv. 

The fix is this: when joining between two staging tables, we need to join not only on tas (or whatever the fk is) but also on submission id to ensure that we're not pulling back extra records. This problem didn't manifest in unittests because we're working in an isolated database.

Here's what the revised queries look like:

*A18*
```sql
SELECT approp.row_number,
       approp.tas,
       approp.grossoutlayamountbytas_cpe,
       SUM(op.grossoutlayamountbyprogramobjectclass_cpe) AS grossoutlayamountbyprogramobjectclass_cpe_sum
FROM appropriation approp
JOIN object_class_program_activity op ON approp.tas = op.tas
-- line below is the change: submsission_id needs to be in the join
AND approp.submission_id = op.submission_id
WHERE approp.submission_id = {}
GROUP BY approp.row_number,
         approp.tas,
         approp.grossoutlayamountbytas_cpe
HAVING approp.grossoutlayamountbytas_cpe <> SUM(op.grossoutlayamountbyprogramobjectclass_cpe)
```

*A19*
```sql
SELECT approp.row_number,
       approp.tas,
       approp.obligationsincurredtotalbytas_cpe,
       SUM(op.obligationsincurredbyprogramobjectclass_cpe) AS obligationsincurredbyprogramobjectclass_cpe_sum
FROM appropriation approp
JOIN object_class_program_activity op ON approp.tas = op.tas
-- line below is the change: submsission_id needs to be in the join
AND approp.submission_id = op.submission_id
WHERE approp.submission_id = {}
GROUP BY approp.row_number,
         approp.tas,
         approp.obligationsincurredtotalbytas_cpe
HAVING approp.obligationsincurredtotalbytas_cpe <> SUM(op.obligationsincurredbyprogramobjectclass_cpe)
```

jworcestBAH:
The specificity of the field names in place of multi-field is an intended feature, since we had that info and nothing is dependent on "multifield" I put it in.  For the errors, I'm fine with not catching prefix errors for invalid object classes, but I'd like the bad value check to still happen.  Could you put the 'X' or other bad value into one of the rows that does get checked?  This would involve the unit test expecting one fewer row to pass and one extra error.

bsweger:
@jworcestBAH Thanks for confirming that the specificity won't break anything. Glad we're able to show more info now.

For the bad `ABCD` objectclass, I did add that to the test data for `test_award_fin_mixed` and adjusted the stagingRows number down by 1 (since this change made a previously valid row invalid). If you want to check it out (which would be great0 you should be able to the result in the error report if you run the tests locally.

`test_program_mixed` already had row that tests this condition: an 4-digit objectclass that begins with '3'. That shows up as a bad objectclass but no longer shows up on the list of invalid prefixes.

jworcestBAH:
Looks good, merging.

PR #106, opened 2016-06-21
update b12 sql rules

bsweger:
@jworcestBAH This is the PR for the B12 SQL changes we talked about:

* also implement the rules for program_activity
* add a separate check for blank, R, or D when reimbursable flag isn't required
* make sure the check is case insensitive


dataactvalidator/config/sqlRules.csv
```diff
@@ -1,4 +1,7 @@
...
...8100_upadjsprioryeardeliveredordersobligunpaid_cpe IS NOT NULL OR ussgl480200_undeliveredordersobligationsprepaidadvanced_fyb IS NOT NULL OR ussgl480200_undeliveredordersobligationsprepaidadv_cpe IS NOT NULL OR ussgl483200_undeliveredordersobligtransferredppdadv_cpe IS NOT NULL OR ussgl488200_upadjsprioryrundelivordersobligprepaidadv_cpe IS NOT NULL OR ussgl490200_deliveredordersobligationspaid_cpe IS NOT NULL OR ussgl490800_authorityoutlayednotyetdisbursed_fyb IS NOT NULL OR ussgl490800_authorityoutlayednotyetdisbursed_cpe IS NOT NULL OR ussgl498200_upadjsprioryrdelivordersobligpaid_cpe IS NOT NULL) AND COALESCE(lower(bydirectreimbursablefundingsource),'') NOT IN ('r', 'd')",B12,"Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".","Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".",n,award_financial,fatal
```
jworcestBAH: Does it make sense for this to still be checking against 'r' and 'd' when that's done by the rule below?  I would think we might want this one to just check that it is not empty (null or '').  In this implementation if a user puts in 'Q', they'll get both errors, this one telling them it's required and the next one telling them it's a bad value.

bsweger: Good point, let's make that separation clear. Just pushed some revised SQL.

jworcestBAH:
Looks good, merging

PR #107, opened 2016-06-21
Db 667 unit tests

jworcestBAH:
This includes some fixes to sql validation executions, and updates to sample files.


dataactvalidator/validation_handlers/validator.py
```diff
@@ -846,26 +847,41 @@ def validateFileBySql(submissionId,fileType,interfaces):
...
+                    if values is None:
+                        values = "{}: {}".format(field,failure[field])
+                    else:
+                        values = ", ".join([values, "{}: {}".format(field,failure[field])])
+                    if fieldNames is None:
+                        fieldNames = field
+                    else:
+                        fieldNames = ", ".join([fieldNames,field])
                 errors.append([fieldNames,errorMsg,values,row])
+            # Pull where clause out of rule
```
bsweger: I put a note to our future selves about this: https://federal-spending-transparency.atlassian.net/browse/DB-708
(no action necessary, was just reminded again that we shouldn't be doing this long-term)

tests/awardFinancialMixed.csv
```diff
@@ -2,11 +2,11 @@ allocationtransferagencyidentifier,agencyidentifier,beginningperiodofavailabilit
...
 0,19,2016,2016,,113,0,ABCD,ABCD,123456803,12-34-56-78-90-AS-DF-AB-XZ-YW,Overseas Programs,2,1255,D,6500,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,13000,52000,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,,6500,6500,6500,6500,26000
-0,19,2013,2016,B,999,0,ABCD,ABCD,123456804,12-34-56-78-90-AS-DF-AB-XZ-YW,Overseas Programs,2,1255,R,,,,,,,,,,,,,,,,,,,,,,,,,,,,6500,,,,,
-0,19,2016,2016,,113,0,ABCD,ABCD,123456805,12-34-56-78-90-AS-DF-AB-XZ-YW,Overseas Programs,2,1255,D,6500,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,13000,52000,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,6500,6500,6500,6500,6500,26000
+0,19,2013,2016,B,999,0,ABCD,ABCD,123456804,12-34-56-78-90-AS-DF-AB-XZ-YW,Overseas Programs,2,1255,RR,,,,,,,,,,,,,,,,,,,,,,,,,,,,6500,,,,,
+0,19,2016,2016,,113,0,ABCD,ABCD,123456805,12-34-56-78-90-AS-DF-AB-XZ-YW,Overseas Programs,2,2255,R,6500,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,13000,52000,6500,26000,6500,6500,6500,6500,6500,26000,6500,6500,6500,6500,6500,6500,6500,6500,6500,26000
```
bsweger: Should this file also include the scenario where we're expecting a reimbursable funding flag and the value is something other than empty, R, or D? (to make sure the two rules interact as expected)?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -846,26 +847,41 @@ def validateFileBySql(submissionId,fileType,interfaces):
...
                 row = failure["row_number"]
-                errorMsg = rule.rule_error_msg
-                values = ""
-                fieldNames = ""
-                for field in failure:
+                errorMsg = rule.rule_error_message
+                values = None
+                fieldNames = None
+                # failure is a RowProxy, which does not quite act the same as a dict
+                for field in failure.keys():
```
bsweger: This is more about coding style and syntax preference, but it would be possible to avoid explicitly looping through the columns in the query result and having to do different formatting for the first value of the `values` and `fieldNames` variables. 

The cross-file processing does this by using a few list comprehension: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/db-429-cross-file-sql/dataactvalidator/validation_handlers/validator.py#L47

(I'm a Python list comprehension fan)

bsweger:
@jworcestBAH done reviewing. A few minor questions, and also a query to make sure we're covering the full range of possibilities in the unittests. I left a comment above re: `awardFinancialMixed.csv` and have a similar question on `programActivityMixed.csv`: should that include the case where we're checking for an invalid `bydirectreimbursablefundingsource` when that field is option (i.e., none of the triggering USSGL fields are present).

Or is the point of this solely to run the SQL rules against the existing tests? If so, that's fair. We can update the test cases next sprint.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Good idea, eventually we do want a more robust way of doing this.

tests/awardFinancialMixed.csv
jworcestBAH: The program activity file has 'Q' and 'RR', and has '1255' paired with 'R', along with just '255' paired with both 'D' and 'R'.  All of those are records that have the ussgl numbers present.  Does that cover the cases you're looking for here, or do you have a different test in mind?

jworcestBAH: Reading the final comment, it seems like you wanted the invalid value where all the USSGL numbers are missing, which is what line 8 here does.  The only one populated here is "transactionobligatedamount".

bsweger: Heh, that was a confusing note I left. I was thinking that we're checking the following for the B12 implementation:

1. reimbursable flag not required and is set to blank/Null, R, or D (passes)
2. reimbursable flag not required and is set to an invalid value (fails)
3. reimbursable flag is required and is set to R or D (passes)
4. reimbursable flag is required and is set to blank (fails)
5. reimbursable flag is required and is set to an invalid value (fails...one time only)

In `awardFinancialMixed`, I didn't see scenario 1.

bsweger: Continuing the above, in `programActivityMixed` I didn't see scenario 1 or 2.

Since there are separate SQL queries for each file, and because the SQL null/blank field handling can be tricky, my instinct is to cover as many bases as possible in the test files.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Changed to be more similar to that

tests/awardFinancialMixed.csv
bsweger: After chatting with @jworcestBAH, we decided to leave this, since the point of this changeset is to make sure the SQL-based implementation of B12 behaves the same way as the prior implementation.

bsweger:
Looks good--merging!

PR #108, opened 2016-06-22
Rule doc updates

bsweger:
* Mark A18 and A19 as complete in the rule doc
* Update the .zip of test files


dataactvalidator/config/rule_documentation.csv
```diff
@@ -14,8 +14,8 @@ A,A14,GrossOutlayAmountByTAS_CPE= CPE value for GTAS SF 133 line #3020,Calculati
...
+"A, B",A19,ObligationsIncurredTotalByTAS_CPE (File A) = sum of ObligationsIncurredByProgramObjectClass_CPE (File B),Calculation,Cross File Calculation,Warning,,"Obligations Incurred Total By TAS in File A must match the sum of Obligations Incurred By Program Object Class in File B, by TAS.","For the TAS in each submission row in File A, match to same TAS in File B, sum ObligationsIncurredByProgramObjectClass_CPE (including blank prog actv & obj cls), and validate match to ObligationsIncurredTotalByTAS_CPE in File A.",No,Ready
 A,A20,Must be valid Agency ID in the Common Government-wide Accounting Classification (CGAC),Requirement,External CGAC,Fatal Error,,The Agency ID is defined in CGAC,Match value to CGAC Agency Codes tab in Domain Values file.,Yes,Ready
```
jworcestBAH: Can we change the second to last column ("Implemented") to Yes, since these rules are implemented now?

dataactvalidator/config/rule_documentation.csv
```diff
@@ -47,4 +47,4 @@ C/D2,C11,"Unique PIID, ParentAwardId from file C exists in file D1",Linkage,Cros
...
 C/D1,C14,If FAIN is not provided then provide URI ,Element Conditionally Required,Conditional requirement,Warning,,,JW: ready?,Yes,"Including PIID in this rule, one of the three must be present "
 C,C17,Not Required on rows that include values for USSGL accounts balances or subtotals,Element Conditionally Required,Conditional requirement,Warning,,This is for transactions not for USSGL amounts from the financial system.,JW: ready,Yes,Ready
-C,C18,DeobligationsRecoveriesRefundsOfPriorYearByAward_CPE = USSGL(4871+ 4872 + 4971 + 4972). This applies to the award level.,Requirement,Record calculation,Warning,,The first element is a total of the other elements,,Yes,
+C,C18,DeobligationsRecoveriesRefundsOfPriorYearByAward_CPE = USSGL(4871+ 4872 + 4971 + 4972). This applies to the award level.,Requirement,Record calculation,Warning,,The first element is a total of the other elements,,Yes,
```
jworcestBAH: I'm not seeing what the change is here, is there something different between these lines?

jworcestBAH:
Thanks for updating the zip, I keep forgetting to do that.  Done reviewing, just a couple questions.

dataactvalidator/config/rule_documentation.csv
bsweger: It looks like there's no longer a line break at the end of the file. Could that be the reason? Definitely wasn't my intent to change this line.

bsweger:
Thanks @jworcestBAH. Updated the implemented column and added back the ending line break, which seems to get rid of that diff at the end of the file.

jworcestBAH:
Looks good, merging as soon as Jenkins gives the okay

PR #110, opened 2016-06-24
Hotfix blank year

jworcestBAH:
Initial implementation of BPOA check fails the record if year cannot be cast to an int, but fails to make an exception for blank values


jworcestBAH:
So the schema team ran into an issue where their files have no BPOA reported, but are getting program activity errors.  We're supposed to only check program activity if BPOA is 2016 or later, since that's what we have data for.

bsweger:
Looks good--merging!

PR #111, opened 2016-06-24
Hotfix stable blank year

jworcestBAH:
Same as dev PR, but with lower file sizes on unit tests since 664 is not included.


bsweger:
Looks good--merging!

PR #113, opened 2016-06-24
prep for loading short col names to db

bsweger:
First step for incorporating short column names into the broker. This PR adds a `fieldname_short` column to the files that load the column/.csv schema info into the validation database. 

That piece isn't updated here--it just gets the new column names out there so people can see them.


jworcestBAH:
Looks good, merging.

PR #114, opened 2016-06-27
Db 705 sql loader

nmonga91:
@bsweger or @jworcestBAH


dataactvalidator/filestreaming/sqlLoader.py
```diff
@@ -0,0 +1,55 @@
...
+                raise KeyError("".join(["Found unexpected fields: ", str(list(unknownFields))]))
+
+            missingFields = set(SQLLoader.headers)-set(fieldNames)
+            if len(missingFields) != 0:
+                raise ValueError("".join(["Missing required fields: ", str(list(missingFields))]))
+
+            reader = csv.DictReader(csvfile, fieldnames=fieldNames)
+            for row in reader:
+                sql_filename = "".join([row['query_name'], ".sql"])
+                sql_file = open(os.path.join(SQLLoader.sql_rules_path, sql_filename), 'rU')
```
bsweger: Do we need to have some error handling here in case the expected .sql file isn't present?

dataactvalidator/filestreaming/sqlLoader.py
```diff
@@ -0,0 +1,55 @@
...
+                sql = " ".join(map(lambda line: FieldCleaner.cleanString(line.replace("\n", ""), removeSpaces=False), sql_file.readlines()))
```
bsweger: hooray for `map`/`lambda`!

bsweger:
@nmonga91 look good! my only suggestion is to add some error handling in case the expected .sql file for a rule isn't found. gonna merge, though, so I can fix up some merge conflicts on my branch and unblock the sql query stories

dataactvalidator/filestreaming/sqlLoader.py
nmonga91: `open` does some of that and throws an exception saying that the specified file doesn't exist.

bsweger: right-good point!

PR #115, opened 2016-06-27
Db 711 approved user list

nmonga91:
@bsweger or @jworcestBAH

### DIFF
--- a/dataactcore/utils/requestDictionary.py
+++ b/dataactcore/utils/requestDictionary.py
@@ -1,22 +1,30 @@
 import json
+from dataactcore.utils.responseException import ResponseException
+from werkzeug.exceptions import BadRequest
+
 class RequestDictionary() :
     """ Provides an interface to an http request """
 
-    def __init__(self,request):
+    def __init__(self,request,optionalRequest=False):
+        try:
+            self.requestDict = None
+            if(not ("Content-Type" in request.headers)):
+                raise ValueError("Must include Content-Type header")
 
-        self.requestDict = None
-        if(not ("Content-Type" in request.headers)):
-            raise ValueError("Must include Content-Type header")
-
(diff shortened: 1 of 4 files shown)


dataactcore/utils/requestDictionary.py
```diff
@@ -1,22 +1,30 @@
 import json
+from dataactcore.utils.responseException import ResponseException
+from werkzeug.exceptions import BadRequest
```
bsweger: great to see some of Flask's built-in handling being used!

tests/userTests.py
```diff
@@ -137,15 +137,18 @@ def test_status_change_bad_status(self):
...
         """Test getting user list by status."""
         postJson = {"status": "denied"}
-        response = self.app.post_json("/v1/list_users_with_status/", postJson, headers={"x-session-id":self.session_id})
+        response = self.app.post_json("/v1/list_users/", postJson, headers={"x-session-id":self.session_id})
         self.check_response(response, StatusCode.OK)
         users = response.json["users"]
         self.assertEqual(len(users), 1)
 
+        response = self.app.post_json("/v1/list_users/", headers={"x-session-id": self.session_id})
+        self.check_response(response, StatusCode.OK)
```
bsweger: Does it make sense to all check for the number of users returned here, to test the scenario where not sending a parameter results in a complete user list being returned?

bsweger:
@nmonga91 done reviewing--just a small question on one of the unit tests. Makes a lot of sense to consolidate these routes!

tests/userTests.py
nmonga91: yeah we can add that. since the goal is to check that all users are indeed being returned, i can put the total number of users in the check.

bsweger:
Thanks for updating the unit test @nmonga91. Looks good--merging!

PR #116, opened 2016-06-27
Db 702 new colnames

bsweger:
This PR completes updates the staging table models to use the new ["machine friendly" column names provided by the schema team](https://federal-spending-transparency.atlassian.net/secure/attachment/13500/AgencyLabel_to_TerseLabel.xlsx).

Note that the column aliases are still in effect, meaning that we're still only expecting data submissions in the long-name format. A later subtask will add in the logic to accept either the short or long names. For now, just getting the short column names in there so we can write sql queries against them.

This also updates the TAS components in the SF-133 table, since that model uses the composite TAS function, which has changed to use the new column names.

Re-wrote the migrations to alter the column names instead of dropping/adding columns, so existing data in the staging tables should be preserved.


tests/fileTypeTests.py
```diff
@@ -110,7 +110,7 @@ def test_program_mixed(self):
         """Test mixed job with some rows failing."""
         jobId = self.jobIdDict["programMixed"]
         self.passed = self.run_test(
-            jobId, 200, "finished", 21060, 4, "complete", 118, True)
+            jobId, 200, "finished", 20998, 4, "complete", 118, True)
```
jworcestBAH: Do these file size changes mean that we're reporting back short names now?  I'm not sure we want to be taking long names but reporting short names, that will be confusing.

jworcestBAH:
Looks good, just one question about the change in file size on unit tests.

tests/fileTypeTests.py
bsweger: Yes, good point! Because the SQL queries return the column names in the physical database, that's what is causing the new file sizes. One acceptance criteria on this story states that we should be returning the long names, so we won't close the book on this until a subsequent PR makes that change.

In the meantime, however, I thought it was more important to do the first step of getting the db columns updated so people can write queries using the new names.

jworcestBAH: Okay, so we'll merge this PR as is to make these names available, and the next PR on this should include a way to report long names.  I think long-term we had wanted to report either short or long names based on whichever was provided in that file, but that may be part of another story.

PR #117, opened 2016-06-28
Db 587 cross file errors

jworcestBAH:
Updates routes for source and target file in cross-file errors, and splits cross-file errors into multiple reports


jworcestBAH:
Working on conflicts currently

jworcestBAH:
Okay, ready for review now

dataactcore/models/errorModels.py
```diff
@@ -46,4 +46,7 @@ class ErrorMetadata(Base):
     occurrences = Column(Integer)
     first_row = Column(Integer)
     rule_failed = Column(Text, nullable=True)
+    file_type_id = Column(Integer)
```
bsweger: I'm curious about the original reasoning for having the error tables in a different table than the validation tables, since those two functions are so closely related. Now that we're referencing file types here, we might stand to gain from combining these databases to take advantage of foreign keys and sqlalchemy relationships.

dataactvalidator/interfaces/validatorValidationInterface.py
```diff
@@ -1,46 +1,15 @@
...
-from dataactcore.models.baseInterface import BaseInterface
+from dataactcore.models.validationInterface import ValidationInterface
 from dataactcore.models.validationModels import Rule, RuleType, FileColumn, FileType, FieldType, RuleTiming, RuleSeverity, RuleSql
 from dataactcore.models.domainModels import TASLookup
 from dataactvalidator.filestreaming.fieldCleaner import FieldCleaner
 from dataactcore.config import CONFIG_DB
 
 
-class ValidatorValidationInterface(BaseInterface):
+class ValidatorValidationInterface(ValidationInterface):
```
bsweger: Nice catch--thanks for simplifying this and removing the unecessary code repetition.

dataactcore/models/errorModels.py
jworcestBAH: The main original motivation is that the error DB is primarily used for communication between the validator and broker, recording errors to be sent back to the user, while the validation DB is primarily internal to the validator, used during the validation process.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -318,16 +320,45 @@ def runCrossValidation(self, jobId, interfaces):
...
+        fileTypes = interfaces.validationDb.getFileTypeList()
+        # Create dict mapping file types to empty lists
+        fileTypeDict = {targetFile:[] for targetFile in fileTypes}
+
+        # Sort failures by source and target file
+        failureDict = {sourceFile:copy.deepcopy(fileTypeDict) for sourceFile in fileTypes}
+
+        for failure in failures:
+            errorDb.recordRowError(jobId, "cross_file",
+                    failure[0], failure[2], failure[4], failure[5], failure[6], failure[7])
```
bsweger: Do you think it would be helpful to include the target file in the error reports as well? I know you can infer that from the file name, and it would be repetitive, but I'm wondering if we'd ever want a sure way to get the target file w/o having to parse the name of the error report. Also, though the front-end may not need that name, as someone who opened an error report to review it, I think it would be helpful to have the target file listed explicitly.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -318,16 +320,45 @@ def runCrossValidation(self, jobId, interfaces):
...
+            # Put failure into list for this source and target file
+            sourceType = interfaces.validationDb.getFileTypeById(failure[6])
+            targetType = interfaces.validationDb.getFileTypeById(failure[7])
+            if failure[7] < failure[6]:
```
bsweger: If the source/target ordering of files is important, we might consider adding an explicit ordering column to the `FileType` model and keep the surrogate key meaningless. I realize we control those keys when loading the table, but it doesn't feel right to add meaning to `file_id`.

If the order doesn't actually matter, and this `<` check is to ensure that we only get a single error file per source/target combo, then please disregard this comment!

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -318,16 +320,45 @@ def runCrossValidation(self, jobId, interfaces):
...
+                # Lower ID goes first
+                failureDict[targetType][sourceType].append(failure)
+            elif failure[7] == failure[6]:
+                raise ValueError("Both files in cross file are the same type")
+            else:
+                failureDict[sourceType][targetType].append(failure)
+
+        # For each pair of files create an error report
+        for sourceFile in fileTypes:
```
bsweger: Not sure if this was the intent, but when I ran this locally, the resulting `filesubmission_12_cross_award_financial_award.csv` had two rows, each with a different value in the `Source file` column (award_financial and award). Is that expected? This goes back to my confusion about not seeing the target file specified in the error report. I might be misunderstanding how this is supposed to work, though.

bsweger:
@jworcestBAH done reviewing. Some questions above, but mostly I want to throw this idea out there...if we're now chunking the cross file errors into separate files, the cross file validation code might be overall easier to understand and maintain if it follows that same pattern. I got a little lost in the various loops and dictionaries needed to keep track of things

What if instead of getting back all the cross file errors in a single list and then trying to map that list into the appropriate error report, we run sql validations individually for each error report?

I played with a little code to test the concept, but didn't fully flush it out...

```python
from dataactcore.models.validationModels import FileType, RuleSql
from dataactcore.models.validationInterface import ValidationInterface

# use db to get a list of the cross-file combinations
validation_db = ValidationInterface()
targetFiles = validation_db.session.query(FileType).subquery()
crossFileCombos = validation_db.session.query(
    FileType.name.label('source_file_name'),
    FileType.file_id.label('source_file_id'),
    targetFiles.c.name.label('target_file_name'),
    targetFiles.c.file_id.label('target_file_id')
).filter(FileType.file_id < targetFiles.c.file_id)
print('cross file error report source/target list:')
print(crossFileCombos.all())

# get all cross file rules from db
crossFileRules = validation_db.session.query(RuleSql).filter(RuleSql.rule_cross_file_flag==True)

# for each cross-file combo, run associated rules and create error report
for row in crossFileCombos:
    comboRules = crossFileRules.filter(
        RuleSql.file_id==row.source_file_id,
        RuleSql.target_file_id==row.target_file_id)
    print('{} rules found for source={} and target={}'.format(
        comboRules.count(),row.source_file_name,row.target_file_name))
    # send comboRules to validator.crossValidate sql
    # get error file name
    # loop through failures to create the error report
```

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: The fain/uri rule between C and D2 is applied in both directions (C8 and C9) .  So one error is that a fain/uri from file C was not found in D2, and the other was that a fain/uri in D2 was not found in C, this is why the source file is different for the two errors.  The displayed values come from the source file, and are being checked against the other file in the pair.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Order may actually matter only because the frontend will probably check for specific pairs.  I'll add an extra column for this.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Added target file to error reports

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Target file has now been added, hopefully this makes it a bit clearer that these two errors are happening in opposite directions.

jworcestBAH:
I suppose it would be possible to separate by file types before executing the rules rather than after, to minimize the rework needed I'll pass the file types into the cross-file validation function which will then only do rules for one pair.  This will involve checking the rules table 6 times rather than the current single query to get rules, but as that's not a very large table this shouldn't be too much of a performance impact.  I'll wait until you approve this method before starting the changes.

bsweger:
@jworcestBAH I think we can avoid querying the rules table 6 times by getting one query result that has all the cross-file rules and then filtering it as necessary (that's what my noodling code was doing). Haven't tested this in SQL Alchemy, but other ORMs will query the query and not hit the database again.

jworcestBAH:
Changed the approach for cross-file rules and removed C8 and C9.  Since that emptied the cross file rules of the old method, I've also removed functionality that was specific to the old implementation of cross file rules

dataactcore/models/errorModels.py
bsweger: Thanks for clarifying.

dataactvalidator/validation_handlers/validationManager.py
bsweger: Thanks--helpful to see that info in the file.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -310,55 +311,44 @@ def runValidation(self, jobId, interfaces):
...
+            targetFiles.c.name.label('target_file_name'),
+            targetFiles.c.file_id.label('target_file_id')
+        ).filter(FileType.file_id < targetFiles.c.file_id)
+
+        # get all cross file rules from db
+        crossFileRules = validationDb.session.query(RuleSql).filter(RuleSql.rule_cross_file_flag==True)
+
+        # for each cross-file combo, run associated rules and create error report
+        for row in crossFileCombos:
+            comboRules = crossFileRules.filter(
```
bsweger: Is this filter sufficient for making sure we include the ["two way rules" you explained](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/117#discussion_r69129088). I was playing around with this query before I understood that particular use case, so want to make sure we're covering all the bases.

bsweger:
@jworcestBAH thank you for going back and forth on this. All the new changes look good. I have one question, though (inline, above) re: the filtering that we do to grab the rules for each file. When playing around with that code, I wrote the filter w/o the "rule both ways" use case in mind. Given that, is that filter sufficient for grabbing the rules for each file? Do we need to add an or filter that checks the other direction?

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Added file_order to be checked for this.

jworcestBAH:
Good point, I'll add a filter that goes the other direction and flips the labels for source and target files, which I'll probably rename "first" and "second" to distinguish between order in the report name, which is what these are, and source/target as used within the report and the error metadata.

jworcestBAH:
When we filter for rules we'll also need to include both directions

jworcestBAH:
So actually the single filter in creating the combos is sufficient as long as we filter for rules in both directions, which I've just pushed an update to do.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: This filter now includes rules in both directions.

PR #119, opened 2016-06-28
Preprocessing now strips extra whitespace and replaces blank values w…

jworcestBAH:
…ith null


tests/fileTypeTests.py
```diff
@@ -123,6 +124,13 @@ def test_award_fin_mixed(self):
         jobId = self.jobIdDict["awardFinMixed"]
         self.passed = self.run_test(
             jobId, 200, "finished", 17206, 5, "complete", 87, True)
+        # Test that whitespace is converted to null
+        rowThree = self.interfaces.validationDb.session.query(AwardFinancial).filter(AwardFinancial.row_number == 3).filter(AwardFinancial.submission_id == self.interfaces.jobDb.getSubmissionId(jobId)).first()
```
bsweger: This is a great unittest -- glad we're checking for all of this. Is there any other way to query for the records you need besides depending on the row number? TAS? anything? Depending on things happening in a particular row number seems fragile for the long term.

dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -111,10 +111,17 @@ def cleanRow(row, fileType, validationInterface):
...
             value = row[key]
-            if value is not None and field_type in ["INT", "DECIMAL", "LONG"]:
-                tempValue = value.replace(",","")
-                if FieldCleaner.isNumeric(tempValue):
-                    row[key] = tempValue
+            if value is not None:
+                # Remove extra whitespace
+                value = value.strip()
+                if field_type in ["INT", "DECIMAL", "LONG"]:
+                    tempValue = value.replace(",","")
```
bsweger: Now that we're stripping commas here, does that mean the code @nmonga91 added in the validation routine to ignore commas is no longer needed?

bsweger:
@jworcestBAH done reviewing!

dataactvalidator/filestreaming/fieldCleaner.py
jworcestBAH: The comma removal for numeric values is not new (it's also in the deleted section here), I think this was the way we wound up ignoring commas, it had to be done before validation so the values could pass the type check.

tests/fileTypeTests.py
jworcestBAH: Just to be clear, this is the row number from the test file, not an autoincremented primary key.  Most of the rows in the test files are identical for many values.  We could give these rows specific values in some other field to query for, but that seems like it might be equally fragile.

dataactvalidator/filestreaming/fieldCleaner.py
bsweger: Ah, thanks for clarifying!

tests/fileTypeTests.py
bsweger: I get that. I'm wondering if counting on a specific test case being in a specific row in the test file is too fragile (e.g., if someone unknowningly inserts a row).

bsweger: @jworcestBAH Just to clarify, I don't think this point needs to hold up merging this. Was more fishing around to see if there's an easy way to replicate the unittest functionality without the row number. If there's not, there's not. Also, I anticipate that we'll have to take a pass through our unit tests in the near future anyway, as we move to real data.

jworcestBAH: We could switch from finding the row by row_number to finding it by some other specific value we put in for this purpose, I'll make this change after finishing 587

jworcestBAH: Now checking against parentawardid instead of row number, so inserting or deleting rows won't affect these tests.

bsweger:
Looks good--merging!

PR #122, opened 2016-06-29
Db 759 tas preprocessing

jworcestBAH:
This includes some of the code from the current DB-707 PR, so that should be completed before this PR.


dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -106,10 +106,10 @@ def cleanLength(length):
             raise ValueError("Length must be positive")
         return length
 
-    @staticmethod
-    def cleanRow(row, fileType, validationInterface):
+    @classmethod
+    def cleanRow(cls, row, fileType, interfaces):
         for key in row.keys():
-            field_type = validationInterface.getColumn(key, fileType).field_type.name
+            field_type = interfaces.validationDb.getColumn(key, fileType).field_type.name
```
nmonga91: I know it doesn't matter here, but is there a reason we're passing all the interfaces with the `interfaces` parameter vs just the `validationInterface`?

jworcestBAH: When I wrote the padField function I wrote it to take the full interfaceHolder object since that's what we typically pass around.  Then when I added it to the cleanRow function I needed the interfaceHolder object available to pass into it, so I made this change.  I'll change padField to just require the validation interface and put this back the way it was.

nmonga91: Ok sounds good, thanks!

nmonga91:
Everything looks good, merging!

PR #125, opened 2016-06-30
Db 697 sql file a

jworcestBAH:
Includes all 'A' rules as sql


jworcestBAH:
This includes material from PR 122: https://github.com/fedspendingtransparency/data-act-broker-backend/pull/122

dataactvalidator/config/sqlrules/a21_appropriations.sql
```diff
@@ -0,0 +1,7 @@
+SELECT
+    row_number,
+    availability_type_code
+FROM appropriation
+WHERE submission_id = {}
+AND LOWER(COALESCE(availability_type_code,'')) NOT IN ('X','x','')
```
bsweger: Now that we're using `LOWER`, we don't need `X` in the list of acceptable values.

bsweger:
@jworcestBAH This all looks good. One comment about `x` and `X` (see above) and also wanted to confirm that the reason for the file size differences in the unit tests are because we're still using the short names.

dataactvalidator/config/sqlrules/a21_appropriations.sql
jworcestBAH: Good catch, that got left in from experimenting with different forms, it's now been removed.

jworcestBAH:
Difference in number of errors and file size change is primarily because A21 had not been applied to files B and C in the original definitions.  Prac & proc specifies that it applies to all three files, so I added it as a sql rule for all three, which brings in the new errors.

bsweger:
Looks good--merging!

PR #126, opened 2016-07-01
Db 716 file b sql rules

nmonga91:
SQL Rules for File B


jworcestBAH:
The update CGAC script and associated changes don't seem to have much to do with file B, is this just a case of the branch you happened to be on when you wrote the script or is there a deeper connection I'm missing here?

dataactcore/models/jobTrackerInterface.py
```diff
@@ -275,3 +275,6 @@ def getSubmissionStatus(self,submissionId):
         if statuses["finished"] == len(jobIds)-skip_count: # need to account for the jobs that were skipped above
             return "finished"
         return "unknown"
+
+    def getSubmissionsByUserId(self, user_id):
+        return self.session.query(Submission).filter(Submission.user_id == user_id).all()
```
jworcestBAH: This is an existing function in dataactbroker.handlers.jobHandler.  I'm fine with moving it here for use in the cgac script, but we should delete the one in jobHandler to avoid the duplication.  Since jobHandler inherits this class and the function is identical except for user_id vs userId and the missing docstring, we shouldn't need to make any other changes after deleting the jobHandler one (and if you could move the docstring over, that would be helpful).

nmonga91: I can move it. We could also leave it and remove it here since the cgac script isn't a permanent solution.

nmonga91:
@jworcestBAH the CGAC scripts is completely unrelated. You're right in assuming I was just in this branch when fixing the references in Production

dataactcore/scripts/updateCGAC.py
```diff
@@ -20,21 +21,44 @@ def updateCGAC():
     all_users = userDb.session.query(User).all()
     all_submissions = jobDb.session.query(Submission).all()
 
+    users_to_update = []
+
     for user in all_users:
-        user.cgac_code = findCGACCode(all_cgac, user.cgac_code)
+        if user.cgac_code != None:
+            users_to_update.append(user)
```
jworcestBAH: Does it break any part of the broker to still have users with no CGAC?  Will things like the recent submissions route just return an empty list, or will they 500?

nmonga91: It'll just show an empty list and the recent activity table wouldn't show any results

dataactcore/scripts/updateCGAC.py
```diff
@@ -20,21 +21,44 @@ def updateCGAC():
...
 
-    for submission in all_submissions:
-        submission.cgac_code = findCGACCode(all_cgac, submission.cgac_code)
+    for user in users_to_update:
+        if user.user_id == 2:
```
jworcestBAH: Was this line when you were updating a specific user?  I'm guessing it should be removed if we want this script to update all users?

nmonga91: Yup, thanks for catching that. i think I'm going to just remove this script from the repo since it's not a permanent solution. it should've only happened for older users, and everyone except for Booz and Kearney people have been updated.

dataactcore/scripts/updateCGAC.py
```diff
@@ -20,21 +21,44 @@ def updateCGAC():
...
+            if validAgency(user.cgac_code, validationDb):
+                user.cgac_code = validationDb.getCGACCode(user.cgac_code)
```
jworcestBAH: As written this is calling getCGACCode twice, could we just have validAgency pass back the cgac code to be stored here if it is not None, so that we don't have to do each query twice?  Probably not that important since this script isn't something that needs to be really efficient.

dataactvalidator/config/programActivityRules.csv
```diff
@@ -1,17 +1,6 @@
...
+SET_EXISTS_IN_TABLE,FALSE,,ProgramActivity,"{""beginningperiodofavailability"":{""target_field"":""budget_year"",""skip_if_below"":2016}, ""agencyidentifier"":{""target_field"":""agency_id"",""pad_to_length"":3}, ""allocationtransferagencyidentifier"":{""target_field"":""allocation_transfer_id"",""pad_to_length"":3},""mainaccountcode"":{""target_field"":""account_number"",""pad_to_length"":4},""programactivitycode"":{""target_field"":""program_activity_code"",""pad_to_length"":4,""skip_if_blank"":true},""programactivityname"":{""target_field"":""program_activity_name"",""skip_if_blank"":true}}",Must be a valid program activity name and code for the corresponding TAS/TAFS as defined in Section 82 of OMB Circular A-11.,B9,multi_field,B9
```
jworcestBAH: Was there a reason that we didn't include B9/B10 in this story?  Was this just an oversight when we listed the rules?

nmonga91: Might have been oversight, I didn't see it listed, but assumed it was intentional

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Either way, as long as we don't leave two copies and the copy we keep gets a docstring I'm happy.

jworcestBAH:
Looks like there are currently merge conflicts, just need to resolve those and pick a location for getSubmissionsByUserId.

nmonga91:
Merge conflicts have been resolved

jworcestBAH:
Looks good, merging

PR #127, opened 2016-07-01
Updated file loaders to be consistent with new preprocessing

jworcestBAH:


dataactvalidator/validation_handlers/validator.py
```diff
@@ -750,14 +752,7 @@ def validateTAS(fieldsToCheck, tasFields, record, interfaces, fileType):
...
             data = record[str(fieldsToCheck[i])]
-            if(data == None):
-                # Set data to empty string so it can be padded with leading zeros
-                data = ""
-            field = fieldsToCheck[i].lower()
-            # Pad field with leading zeros
-            length = interfaces.validationDb.getColumnLength(field, fileType)
-            data = data.strip().zfill(length)
-            query = query.filter(TASLookup.__dict__[tasFields[i]] == str(data))
+            query = query.filter(TASLookup.__dict__[tasFields[i]] == data)
```
nmonga91: Why is the padding removed from this block?

nmonga91:
Everything else looks good, once my question is addressed, I will go ahead and merge

dataactvalidator/validation_handlers/validator.py
jworcestBAH: That padding is now done in the preprocessing before the data is put into staging.  When I added it there I did not think to remove it here, but this is now duplicate functionality.

nmonga91: Got it, thanks for clarifying

nmonga91:
Looks good, merging.

PR #130, opened 2016-07-05
Db 733 error updates

mbrumback:
Updated error messages within the rule_documentation CSV


klrBAH:

jworcestBAH:
It looks like this was branched off master instead of dev, I've taken the rule documentation file and put in into a new branch, I'll also move the updated rule test into the rule files.

PR #131, opened 2016-07-05
SQL Implementation of Validations C8, C9, and C14

mab6bf:


jworcestBAH:
Unit tests not passing, it looks like none of the rows in award financial valid are passing.

dataactvalidator/config/sqlrules/c14_award_financial.sql
```diff
@@ -0,0 +1,5 @@
+SELECT row_number, fain, uri
+FROM award_financial
+WHERE submission_id = {}
+	AND fain IS NULL
+	AND uri IS NULL;
```
jworcestBAH: So the current version of this rule includes piid as well as fain and uri.  This may be why the unit tests are failing.

dataactvalidator/config/sqlrules/c8_award_financial.sql
```diff
@@ -0,0 +1,13 @@
+SELECT af.row_number, af.fain, af.uri
+FROM award_financial AS af
+WHERE af.submission_id = {}
+	AND af.fain IS NOT NULL
+	AND af.uri IS NOT NULL
```
jworcestBAH: This will only catch errors where a row has both fain and uri.  If it only has one of the two, that still needs to match the other file.

klrBAH:
[run tests]

jworcestBAH:
Pushed a fix to mark C8 and C9 as cross-file, but award_financial_mixed is still missing an error and cross-file has two errors extra

jworcestBAH:
Okay, the extra cross-file errors are correct, we had reduced it to 3 when we removed C8 and C9, so 5 is correct

bsweger:
@jworcestBAH @mab6bf...fixed the merge conflicts in dev. In so doing, adjusted the expected errors in the award_financial unit tests again. @jworcestBAH would be great to get your eyes on this because this expected number of errors in the award financial file seems to be a moving target.

klrBAH:
[run tests]

klrBAH:
[run tests]

klrBAH:
[run tests]

jworcestBAH:
The extra two errors come from A21, which was not implemented originally for files B and C, but now is.  Merging now.

PR #133, opened 2016-07-05
Length errors no longer classified as rule_failed since they do not h…

jworcestBAH:
…ave original label


bsweger:
@jworcestBAH looks good--just want to confirm that the front-end is aware of these changes, if necessary.

jworcestBAH:
Yes, Kevin brought up the need for these changes.

bsweger:
Thanks for the clarification--merging!

PR #135, opened 2016-07-05
Handling missing source or target file types

jworcestBAH:
Missing file types returned as empty string rather than 500


bsweger:
Looks good--merging!

PR #136, opened 2016-07-05
Reading only fields that are part of TAS

jworcestBAH:


bsweger:
Looks good, and worked for me locally. Merging!

PR #138, opened 2016-07-07
Removing cross file errors at start of job, and fixed sample file for…

jworcestBAH:
… C to pass file-level validation


bsweger:
@jworcestBAH looks good. Can you also add a new version of test_files.zip? Then we should be good to go.

jworcestBAH:
Added the cross-file samples to the test_files.zip

bsweger:
Looks good--merging!

PR #139, opened 2016-07-07
hotfix for row formatting error

bsweger:
Hotfix for correctly reporting formatting errors when there is no row label.


jworcestBAH:
Looks good, merging

PR #140, opened 2016-07-08
Db 714 split run validation

jworcestBAH:
Split run validation into several parts for easier profiling


nmonga91:
Looks good, merging

PR #141, opened 2016-07-11
Populating is_first_quarter based on submission metadata

jworcestBAH:


bsweger:
@jworcestBAH Done reviewing (have not investigated what Jenkins' issue is).

Will merge this, since it's a blocker to finishing up the SQL validations. I added a clean-up ticket to combine validation and job databases so that it will be easier to grab the quarter info.

Ideally, the SQL validation queries should be able to access the data they need without us needing to duplicate info and add code to ignore the extra columns when validating.

Thanks for getting this done so quickly so we can get A16 into the mix.

jworcestBAH:
Test result is "no tests", so looks like Jenkins was busy or something, trying again: [run tests].

bsweger:
ok, seems like that did the trick--merging!

PR #142, opened 2016-07-12
Db 702 validate short

bsweger:
This PR allows the broker to validate against either long or short column names. 

* Incoming long column names are converted to their short counterparts for the purpose of validations
* Hard-coded column aliases have been removed from the models in favor of the long/short col info in the validation database
* The logic from a previous PR changes those short names back to their long names for error reporting, as required by the acceptance criteria


dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
-        for row in csv.reader([line],dialect='excel', delimiter=self.delimiter):
-            for cell in row :
-                headerValue = FieldCleaner.cleanString(cell)
-                if( not headerValue in possibleFields) :
+
+        # check to see if header contains long or short column names
+        # TODO: is there a better way to do this?
+        validation_db = ValidatorValidationInterface()
+        longNameDict = validation_db.getLongToShortColname()
+        headerNames = line.split(self.delimiter)
```
jworcestBAH: So going around the csv package may cause some problems, I'd assume there are a number of cases that package handles.  For one thing, specifying the dialect as 'excel' probably had some consequence, we should do some research on the differences here before deciding not to use the csv package for this.

jworcestBAH: One thing that occurs to me is we'd need to implement quote handling if we want to do this directly.  If the delimiter is within quotes it should not be split on.

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
+        colMatches = 0
+        for i, value in enumerate(headerNames):
```
jworcestBAH: The first index doesn't seem to be used here, is there a reason this isn't just "for value in headerNames"?

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: @jworcestBAH Thanks for these notes--I'm definitely in the market for a better way to determine for sure whether a submitted file is using the long or short column headers. I'm not sure that you mean, though, about "going around the csv package." Can you say more about that?

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
+            if value in longNameDict:
+                colMatches += 1
+        if colMatches > 15:
```
jworcestBAH: 15 seems a bit arbitrary, maybe we could do "if colMatches > 0.5 * len(headerNames)" so we do long names if a majority match?

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: No reason. I probably just enumerated by default--will change that.

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: Indeed, 15 was totally arbitrary. I really like your suggestion for doing a majority match...will put that in.

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
-                if( not headerValue in possibleFields) :
+
+        # check to see if header contains long or short column names
+        # TODO: is there a better way to do this?
+        validation_db = ValidatorValidationInterface()
+        longNameDict = validation_db.getLongToShortColname()
+        headerNames = line.split(self.delimiter)
+        colMatches = 0
+        for i, value in enumerate(headerNames):
+            if value in longNameDict:
```
jworcestBAH: Should we put value through FieldCleaner.cleanString() before doing this check?  If a user puts a space before each header name ("header 1, header 2, header 3" rather than "header 1,header 2,header 3") this would not find any after the first one.

dataactvalidator/filestreaming/csvAbstractReader.py
jworcestBAH: The deleted line 70 above, or line 91 below are examples of how we have been reading csv lines in other places.  By splitting "line" without reading it through csv.reader, we bypass the handling of excel dialect, possible quoting of fields, and whatever else the csv package checks for.  So ' "header 1, header 2", header 3' would be interpreted by csv.reader as 2 elements ("header 1, header 2") and (header 3), but going through split will give 3 elements ("header 1), ( header 2"), and (header 3).

dataactvalidator/filestreaming/csvAbstractReader.py
jworcestBAH: Actually I just realized a problem with the majority thing, if more than half of the fields present are these extra fields we're allowing but ignoring, this will choose short names.  That said, if a user submits a file with short names, but some of those extra fields are the long name, checking for presence of a set number could also make the wrong choice.  Possibly we could count both presence of short and long names and go with whichever is greater (or pick one in case of a tie).  I may also be overthinking this, we could just state that those extra fields we're not using are not allowed to be the other version of a field name (in which case we could flag that as a header error).

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
+        possibleFields = {}
+        for schema in csvSchema:
+            possibleFields[FieldCleaner.cleanString(schema.name_short)] = 0
+
+        for row in csv.reader([line], dialect='excel', delimiter=self.delimiter):
+            for cell in row:
+                submittedHeaderValue = FieldCleaner.cleanString(cell)
+                if longHeaders and submittedHeaderValue in longNameDict:
+                    headerValue = FieldCleaner.cleanString(longNameDict[submittedHeaderValue])
+                elif longHeaders:
```
jworcestBAH: If we want to flag presence of both long and short headers as a separate type of header error, we could precede this with an "elif longHeaders and submittedHeaderValue in possibleFields:" check here

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
+                if not headerValue in possibleFields:
                     # Allow unexpected headers, just mark the header as None so we skip it when reading
                     self.headerDictionary[(current)] = None
                     current += 1
-                elif(possibleFields[headerValue] == 1) :
-                    # Add to duplicated header list
-                    duplicatedHeaders.append(headerValue)
+                elif(possibleFields[headerValue] == 1):
+                    # Add header value (as submitted) to duplicated header list
+                    duplicatedHeaders.append(submittedHeaderValue)
```
jworcestBAH: This actually takes care of duplicates across long and short header names, so ignore my earlier comment on that.

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -67,25 +66,54 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
+                submittedHeaderValue = FieldCleaner.cleanString(cell)
+                if longHeaders and submittedHeaderValue in longNameDict:
+                    headerValue = FieldCleaner.cleanString(longNameDict[submittedHeaderValue])
+                elif longHeaders:
+                    headerValue = None
+                else:
+                    headerValue = submittedHeaderValue
+                if not headerValue in possibleFields:
                     # Allow unexpected headers, just mark the header as None so we skip it when reading
                     self.headerDictionary[(current)] = None
```
jworcestBAH: So this is probably a Python thing I'm not getting, and I realize you didn't write this line, but is this making the index actually a single-element tuple?  Is there a reason we're doing this instead of just [current]?

jworcestBAH: So after a bit of googling, this is not creating a tuple, but I'm still not sure that the parentheses do anything here except create confusion.

dataactvalidator/interfaces/validatorValidationInterface.py
```diff
@@ -304,19 +307,29 @@ def getColumnById(self, file_column_id):
...
         return self.session.query(FileColumn).filter(FileColumn.file_column_id == file_column_id).first()
 
-    def getColumnId(self, fieldName, fileType):
+    def getColumnId(self, fieldName, fileType, shortCols=True):
         """ Find file column given field name and file type
 
         Args:
             fieldName: Field to search for
             fileId: Which file this field is associated with
+            shortCols: If true, return the short column names instead of the long names
```
jworcestBAH: Maybe "search by short column names" here since all we're returning is the ID?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -150,19 +143,23 @@ def validate(record,rules,csvSchema,fileType,interfaces):
                 # If type failed, don't do value checks
                 continue
             #Field must pass all rules
+            # temporary fix for long/short colnames until legacy validation rules are removed:
+            # use long colnames in the record header
+            colMapping = interfaces.validationDb.getShortToLongColname()
+            recordLong = {colMapping.get(key,key):value for key, value in record.items()}
```
jworcestBAH: Should these two lines be run for every field?  Maybe they should happen before the loop starts on line 101?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -150,19 +143,23 @@ def validate(record,rules,csvSchema,fileType,interfaces):
...
                     recordFailed = True
                     failedRules.append([fieldName,"".join(["Failed rule: ",str(currentRule.description)]), currentData, str(currentRule.original_label)])
         # Check all multi field rules for this file type
         multiFieldRules = interfaces.validationDb.getMultiFieldRulesByFile(fileType)
         for rule in multiFieldRules:
-            if not Validator.evaluateRule(record,rule,None,interfaces,record):
+            if not Validator.evaluateRule(recordLong,rule,None,interfaces,recordLong):
                 recordFailed = True
-                failedRules.append(["MultiField", "".join(["Failed rule: ",str(rule.description)]), Validator.getMultiValues(rule, record, interfaces), str(rule.original_label)])
+                failedRules.append(["MultiField", "".join(["Failed rule: ",str(rule.description)]), Validator.getMultiValues(rule, recordLong, interfaces), str(rule.original_label)])
```
jworcestBAH: Should this be dependent on whether the user passed in long names or not, so that the report has whichever form they chose to use?  Given that this will be gone by the end of the sprint anyway, this probably doesn't matter.

tests/fileTypeTests.py
```diff
@@ -38,29 +38,40 @@ def setUpClass(cls):
...
+            "awardMixedDelimiter": [statusReady, jobTypeCsv, str(submissionIDs[10]), s3FileNameAwardMixedDelimiter, 1],
+            "crossApprop": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossApprop, 3],
+            "crossPgmAct": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossPgmAct, 4],
+            "crossAwardFin": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossAwardFin, 2],
+            "crossAward": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossAward, 1],
+            "crossFile": [statusReady, jobTypeValidation, str(submissionIDs[11]), None, None],
+            "appropValidShortcols": [statusReady, jobTypeCsv, str(submissionIDs[12]), s3FileNameAppropValidShortcols, 3],
+            "programMixedShortcols": [statusReady, jobTypeCsv, str(submissionIDs[13]), s3FileNameProgramMixedShortcols, 4],
+            "awardFinMixedShortcols": [statusReady, jobTypeCsv, str(submissionIDs[14]), s3FileNameAwardFinMixedShortcols, 2],
+            "awardValidShortcols": [statusReady, jobTypeCsv, str(submissionIDs[15]), s3FileNameAwardValid, 1]
```
jworcestBAH: Should this be s3FileNameAwardValidShortcols?

jworcestBAH:
Done reviewing

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: Had a Slack consult about this and collectively decided to use the "majority" method. If we somehow pick the wrong thing, we'd catch that via a downstream header error on the file.

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: Ah, I see. Thanks for clarifying. Will make that change.

dataactvalidator/filestreaming/csvAbstractReader.py
bsweger: 100% with you on 👎  meaningless extra characters. Will remove the offending `()`.

dataactvalidator/validation_handlers/validator.py
bsweger: No, they most certainly should not be run for every field. Good catch.

dataactvalidator/validation_handlers/validator.py
bsweger: I think what we had decided previously was to stick the 702's acceptance criteria of keeping the long fieldname in the error reports. I suspect when agencies start using the short names they will request error reports that reflect their submissions.

tests/fileTypeTests.py
bsweger: Yes, yes it should.

bsweger:
@jworcestBAH thanks for the review--great feedback. I pushed up some changes based on the comments, and those are ready for your eyes.

jworcestBAH:
Looks good now, merging.

PR #143, opened 2016-07-12
Db 724 tas validation

jworcestBAH:
This PR adds A1 as SQL, as well as populating the padded flags for TAS fields during preprocessing


jworcestBAH:
Finished removing padded flags and merged in dev again, ready for review

dataactvalidator/config/sqlrules/a1_appropriations.sql
```diff
@@ -0,0 +1,16 @@
...
+    row_number,
+    allocation_transfer_agency,
+    agency_identifier,
+    beginning_period_of_availa,
+    ending_period_of_availabil,
+    availability_type_code,
+    main_account_code,
+    sub_account_code
+FROM appropriation as approp
+WHERE submission_id IS NOT DISTINCT FROM {}
```
bsweger: Should this be `submission_id = {}`?

dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -146,10 +147,11 @@ def padField(field,value,fileType,validationInterface):
         if value is not None and validationInterface.isPadded(field,fileType):
             # If padded flag is true, get column length
             padLength = validationInterface.getColumnLength(field, fileType)
-            # Pad to specified length with leading zeros
-            return value.zfill(padLength)
-        else:
-            return value
+            if len(value) < padLength:
```
bsweger: Just checking my assumptions here...I _think_ `zfill` will just ignore your padding request if the length of the field in question is already >= the length you're padding to. Is there a case that requires us to have this explicit `if len(value) < padLength` check?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -104,8 +106,8 @@ def validate(record,rules,csvSchema,fileType,interfaces):
         recordLong = {colMapping.get(key, key): value for key, value in record.items()}
 
         for fieldName in record :
-            if fieldName == "row_number" or fieldName == "is_first_quarter":
-                # Skip row number and quarter flag, nothing to validate on that
+            if fieldName in cls.META_FIELDS:
```
bsweger: Good idea to add this. Should the comment below reflect the change?

tests/fileTypeTests.py
```diff
@@ -119,7 +119,7 @@ def test_approp_mixed(self):
         """Test mixed job with some rows failing."""
         jobId = self.jobIdDict["mixed"]
         self.passed = self.run_test(
-            jobId, 200, "finished", 6330, 4, "complete", 46, True)
+            jobId, 200, "finished", 8847, 4, "complete", 46, True)
```
bsweger: Are these files larger now b/c the error message contains more info re: the specific fields and their values?

dataactvalidator/filestreaming/fieldCleaner.py
bsweger: Also, just want to make sure I understand what this padding function is affecting--is it the TAS load (or perhaps other domain value loads with TAS like program activity) or the user submissions themselves as they get loaded to the staging tables? Or both?

bsweger:
@jworcestBAH Done reviewing. Just a few minor comments and a few questions. Thanks!

dataactvalidator/config/sqlrules/a1_appropriations.sql
jworcestBAH: Whoops, got a bit carried away with replace all, will fix that after my tests on a different branch finish running.

dataactvalidator/filestreaming/fieldCleaner.py
jworcestBAH: All of the above, both user submissions and domain value loading gets run through this function.  For the above, you're right that was part of the changes for populating the padding flag, I'll simplify this function back to the earlier form.

tests/fileTypeTests.py
jworcestBAH: In moving to sqlRules.csv, I put in the complete description from the prac & proc document, my old version had a much simpler message about not finding TAS in the CARS file.  The much longer error message led to a larger file size.

dataactvalidator/validation_handlers/validator.py
jworcestBAH: Good catch, fixed.

bsweger:
Looks good--merging!

PR #145, opened 2016-07-13
Staging

nmonga91:


bsweger:
Looks good. Thanks all around @jworcestBAH and @nmonga91!

PR #146, opened 2016-07-14
SQL Implementation of Validations A16 and B9/B10

mab6bf:
B9 and B10 are combined, as they perform the same checks.


dataactvalidator/config/sqlrules/b9_award_financial.sql
```diff
@@ -0,0 +1,25 @@
+SELECT af.row_number,
+	af.beginning_period_of_availa,
+	af.agency_identifier,
+	af.allocation_transfer_agency,
+	af.main_account_code,
+	af.program_activity_name,
+	af.program_activity_code
+FROM award_financial as af
+WHERE af.submission_id = {0}
+	AND CAST(COALESCE(SUBSTR(af.beginning_period_of_availa,1,4),'0') AS integer) >= 2016
```
nmonga91: Could you explain B9/B10 again? That would clear up why it has to be greater than 2016 explicitly in the check.

nmonga91: Cleared up in a discussion via Slack

PR #150, opened 2016-07-15
Rule B4 hotfix

nmonga91:


bsweger:
Merging with the understanding that we'll talk more about how we use the `dev-stable` branch in our process.

PR #152, opened 2016-07-18
Db 784 remove old logic

jworcestBAH:
This PR includes the removal of old validation logic and some performance optimizations


dataactcore/migrations/versions/25b8f47e075c_add_length.py
```diff
@@ -0,0 +1,76 @@
...
+
+def downgrade_user_manager():
+    ### commands auto generated by Alembic - please adjust! ###
+    pass
+    ### end Alembic commands ###
+
+
+def upgrade_validation():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('file_columns', sa.Column('length', sa.Integer(), nullable=True))
```
nmonga91: Is the goal to handle `NULL` values from the database or do we want to default to `0` for the length column?

dataactvalidator/config/sqlrules/b9_award_financial.sql
```diff
@@ -7,7 +7,7 @@ SELECT af.row_number,
 	af.program_activity_code
 FROM award_financial as af
 WHERE af.submission_id = {0}
-	AND CAST(COALESCE(SUBSTR(af.beginning_period_of_availa,1,4),'0') AS integer) >= 2016
```
nmonga91: May have missed this during our many discussions last week, but why is the `SUBSTR` no longer necessary?

dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -126,31 +131,26 @@ def cleanRow(cls, row, fileType, validationInterface):
...
-            interfaces: InterfaceHolder object
 
         Returns:
             Padded value
         """
         # Check padded flag for this field and file
-        if value is not None and validationInterface.isPadded(field,fileType):
-            # If padded flag is true, get column length
-            padLength = validationInterface.getColumnLength(field, fileType)
+        if value is not None and field.padded_flag and field.length is not None:
```
nmonga91: This answers my previous question about handling nulls for the `length` field in the database. It looks like we're handling the `NULL` values instead of just handling a default value.

dataactcore/migrations/versions/25b8f47e075c_add_length.py
jworcestBAH: Many fields don't have a defined length, so we don't want to require this across the board.  If a length is provided, then we check against that length.

dataactvalidator/config/sqlrules/b9_award_financial.sql
jworcestBAH: The substr was preventing it from catching errors for 5 digit years, which broke our current unit tests.  The B9 changes are in this PR because the errors in the unit tests were still coming from the old implementation, which was removed in this PR, so I needed to get the B9/B10 query to actually start catching some errors.

dataactvalidator/filestreaming/fieldCleaner.py
jworcestBAH: Yes, I wanted to allow fields to not have a length defined, and leaving those as null seemed better than imposing an artificial default value like -1

nmonga91:
Looks good, merging!

PR #154, opened 2016-07-18
Fixed file order to match list in validation order as well as A to D …

jworcestBAH:
…order.  Making this order consistent will reduce confusion when looking at values in DB and simplify things if we merge job and validation DBs.


nmonga91:
Looks good, merging

PR #161, opened 2016-07-27
Db 803 submission status

jworcestBAH:
Submission status now distinguishes between submissions with errors and those without.

### DIFF
--- a/tests/fileTests.py
+++ b/tests/fileTests.py
@@ -48,6 +48,10 @@ class FileTests(BaseTestAPI):
         cls.setupFileData(cls.jobTracker, cls.errorDatabase,
             cls.test_metrics_submission_id)
 
+        cls.row_error_submission_id = cls.insertSubmission(
+            cls.jobTracker, cls.submission_user_id, cgac_code = "SYS", startDate = "10/2015", endDate = "06/2016", is_quarter = True)
+        cls.setupSubmissionWithError(cls.interfaces, cls.row_error_submission_id)
+
     def setUp(self):
         """Test set-up."""
         super(FileTests, self).setUp()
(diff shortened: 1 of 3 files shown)


tests/fileTests.py
```diff
@@ -284,6 +284,21 @@ def test_check_status(self):
...
+        self.assertEqual(response.status_code, 200, msg=str(response.json))
+        self.assertEqual(
+            response.headers.get("Content-Type"), "application/json")
+        json = response.json
+        errorReportSub = None
+        for submission in json["submissions"]:
+            if submission["submission_id"] == self.error_report_submission_id:
+                errorReportSub = submission
+        self.assertIsNotNone(errorReportSub)
+        self.assertEqual(errorReportSub["status"], "validation_successful")
```
bsweger: do we need to do any checking around the new `validation_errors` as well?

bsweger:
@jworcestBAH Done reviewing, and everything looks good (just one small unit test question).

Thinking ahead to the tech debt change I'm doing to capture total rows w/ errors in the job table...would it make sense as part of that for me to revisit jobTrackerInterface.py and use the new error count on the job table to set the new `validation_errors` and `validation_successful` statuses? That we we wouldn't have to pass around the errorDb interface. Though I suppose that wouldn't catch type errors.

One last thought--is there anything here that will need revisited once our warnings are in place? For example, if a job has warnings but no errors, do we need another status? If so, should we make a story for that so we don't forget?

tests/fileTests.py
jworcestBAH: Expanded the list_submissions test to check that

jworcestBAH:
The list submission setup function will need to specify severity when warnings are introduced.  From conversations so far, I don't think the frontend will display a separate submission status for warnings, but if this is desired we can add a call to the number of warnings function at that point.  
For switching to the error count on the job table, if that's not going to include all error types we should probably stick with this function so we don't get another scenario with the status saying no errors and a nonzero error count next to it.

bsweger:
@jworcestBAH thanks for expanded the test. looks good--merging!

PR #163, opened 2016-07-27
Python 3 doc update

bsweger:
Remove the vestiges of Python 2.7 from our documentation! Also bumped our status to beta in setup.py.

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -51,12 +51,15 @@ More complete install documentation is available on the PostgreSQL [wiki](https:
 
 ### Install Python and Create Virtual Environment
 
-The broker's backend components currently run on Python 2.7 (with the intention of migrating to Python 3x for the final release). These instructions will walk you through the process of installing Python and creating a Python-based virtual environment to house the DATA Act backend components. A virtual environment will isolate the broker software and its libraries from those running on your local system and prevent potential conflicts.
+The broker's backend components currently run on Python 3.x. These instructions will walk you through the process of installing Python and creating a Python-based virtual environment to house the DATA Act backend components. A virtual environment will isolate the broker software and its libraries from those running on your local system and prevent potential conflicts.
 
 If you already have a Python development environment on your machine and a preferred way of managing it, feel free to skip to the next section. We wrote the directions below for folks who don't have a Python environment up and running yet and need the quickest way to get started.
 
-1. Install Python 2.7 and pip (Python's package manager) using [these directions](https://docs.aws.amazon.com/cli/latest/userguide/installing.html#install-python "Install Python").
-2. Use pip to install virtualenv:
+1. Install Python 3.x:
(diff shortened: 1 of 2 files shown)


doc/CONTRIBUTING.md
```diff
@@ -72,14 +75,14 @@ If you already have a Python development environment on your machine and a prefe
...
 
 6. You should see some output that looks similar to the example below. Essentially, this command creates and activates a new virtualenv named `data-act` with its own set of Python libraries.  Anything you pip install from this point forward will be installed into the *data-act* environment rather than your machine's global Python environment. Your command line prompt indicates which (if any) virtualenv is active.
 
         rebeccasweger@GSA-xs-MacBook-Pro-4 ~
-        $ mkvirtualenv --python=/usr/local/bin/python2.7 data-act
-        Running virtualenv with interpreter /usr/local/bin/python2.7
-        New python executable in data-act-test/bin/python2.7
+        $ mkvirtualenv --python=/usr/local/bin/python3.4 data-act
+        Running virtualenv with interpreter /usr/local/bin/python3.4
+        New python executable in data-act-test/bin/python3.4
```
jworcestBAH: Everywhere above this we put 3.x, here we're doing specifically 3.4, do we want to either make this 3.x or specify that this will vary with their version?  Also, we may want to remove the above line: "rebeccasweger@GSA-xs-MacBook-Pro-4 ~", most users probably don't need that.

setup.py
```diff
@@ -34,14 +34,14 @@
...
+    author_email='databroker@fiscal.treasury.gov',
     license='CC0 1.0 Universal (CC0 1.0) Public Domain Dedication',
     classifiers=[
-        'Development Status :: 3 - Alpha',
+        'Development Status :: 4 - Beta',
         'Intended Audience :: Developers',
         'Topic :: Software Development :: Build Tools',
         'License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication',
-        'Programming Language :: Python :: 2.7'
+        'Programming Language :: Python :: 3.4'
```
jworcestBAH: I'm not really sure what this section is doing, but does this limit us to 3.4 in any way?  I run locally on 3.5, but I never actually run this setup.py file so I'm not sure if that would be a problem.

bsweger: @jworcestBAH it's metadata so shouldn't affect anything...will update

doc/CONTRIBUTING.md
bsweger: Good call on removing the extraneous terminal info. I intended the `mkvirtualenv` command to show a real implementation (there's a note above that the `--python` parameter should point to whatever Python executable you want to use).

I personally find it helpful to see how a command would work in practice but don't want to confuse anyone. What do you think?

jworcestBAH: That makes sense.  Given that users may have a tendency to just copy and paste, what do you think about a reminder either above or below this to substitute their own version in place of the 3.4?

bsweger: @jworcestBAH thanks--added a line above to clarify.

jworcestBAH:
Looks good, merging

PR #165, opened 2016-07-29
Db 712 where clause

bsweger:
This is the PR for [DB-848](https://federal-spending-transparency.atlassian.net/browse/DB-848), a long-standing tech debt task that became a blocker for the SF-133 validations.

* Remove the valid_record flag from the staging tables
* Instead, track total number of valid rows at the job level, and add that information to the job table
* Update unit tests to use the new job-level valid row count instead of the old valid_row flag in the staging tables


dataactcore/models/jobTrackerInterface.py
```diff
@@ -241,16 +240,20 @@ def getNumberOfRowsById(self,jobId):
         """ Get number of rows in file for job matching ID """
         return self.getJobById(jobId).number_of_rows
 
+    def getNumberOfValidRowsById(self, jobId):
+        """Get number of file's rows that passed validations."""
+        return self.getJobById(jobId).number_of_rows_valid
```
nmonga91: Since we're allowing NULL values in the DB for `number_of_rows_valid`, wouldn't we need to have that check here to convert NULL results to 0?

nmonga91: Also, would this have any impact on how `list_submissions` is calculating the error count for a submission?

jworcestBAH: It does not affect list_submissions, as that's going by total number of errors in error DB, not number of valid rows.

bsweger: Good question re: `NULL` handling. So, based on our Slack convo, `NULL` is expected in this field if a job is not yet complete or if it's not a file validation job. Thus, it does make sense for the calling function to make the 0 vs `NULL` interpretation.

As of right now, the only caller is the validation unit tests, which use `getNumberOfValidRowsById` to check the expected number of valid rows in the test files. If, for some reason, that call gets a `NULL` back, the test would fail (which is what we'd want).

bsweger: Love the question re: `list_submissions`--thanks for looking after the big picture.

The new valid row count on the job table is only used by the unit tests currently (though might be handy when we do the agency dashboard). Unlike error counts we store in the error db, which are by field/error type, this number just gives a high-level count of file rows that passed validations.

Does that answer your question?

nmonga91: Yes it does 👍

nmonga91:
Looks good, merging

PR #166, opened 2016-07-29
Db 448 warnings

jworcestBAH:
Splits errors into warnings and fatal errors.


dataactcore/models/errorInterface.py
```diff
@@ -82,6 +82,9 @@ def checkNumberOfErrorsByJobId(self, jobId):
         queryResult = self.session.query(ErrorMetadata).filter(ErrorMetadata.job_id == jobId).all()
         numErrors = 0
         for result in queryResult:
+            if result.severity_id != valDb.getRuleSeverityId(errorType):
+                # Don't count non-fatal
```
bsweger: Very minor point, but it looks like this function is set up to be more flexible than this comment implies (e..g, just finding non-fatal validation results).

dataactcore/models/validationInterface.py
```diff
@@ -48,10 +48,8 @@ def getFileTypeList(self):
         return [fileType.name for fileType in fileTypes]
 
     def getRuleSeverityId(self, name):
-        query = self.session.query(RuleSeverity).filter(RuleSeverity.name == name)
-        result = self.runUniqueQuery(query, "No rule severity found for specified name", "Multiple rule severities found for specified name")
-        return result.rule_severity_id
-
+        """ Return rule severity ID for this name """
+        return self.getNameFromDict(RuleSeverity, "SEVERITY_DICT", "rule_severity_id", name, "name")
```
bsweger: Thanks for making this change--good call.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -486,15 +503,20 @@ def runCrossValidation(self, jobId, interfaces):
             failures = Validator.crossValidateSql(comboRules.all(),submissionId)
             # get error file name
             reportFilename = self.getFileName(interfaces.errorDb.getCrossReportName(submissionId, row.first_file_name, row.second_file_name))
+            warningReportFilename = self.getFileName(interfaces.errorDb.getCrossWarningReportName(submissionId, row.first_file_name, row.second_file_name))
```
bsweger: I know you didn't change the actual `getFileName` code (beginning on line 157), but I noticed that we might want to make that function more robust by using the `os.path.join` function to ensure we build the file paths correctly across OSs, like we do in other places.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -371,10 +382,9 @@ def runValidation(self, jobId, interfaces):
                             continue
 
                     if not passedValidations:
-                        if failures:
+                        if self.writeErrors(failures, interfaces, jobId, shortColnames, writer, warningWriter, rowNumber):
                             rowErrorPresent = True
                             errorRows.append(rowNumber)
```
bsweger: How do we want to treat the warning/error count for the new `number_of_rows_valid` column in the job table? If I'm reading this line correctly, we're not going to strip a row of it's valid status if it only has warnings (because we're not doing `errorRows.append(rowNumber)` unless there's a fatal error.

However, we're treating everything that comes back in the sql error query as a reason to count a row as invalid, regardless of its fatal/warning status (see line 398).

In the longer term, it probably makes sense to add additional metadata to the job table if people request it, but in the short term, we should decide whether or not we want to consider warnings when marking a row as valid.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -486,15 +503,20 @@ def runCrossValidation(self, jobId, interfaces):
...
+            warningReportFilename = self.getFileName(interfaces.errorDb.getCrossWarningReportName(submissionId, row.first_file_name, row.second_file_name))
 
             # loop through failures to create the error report
-            with self.getWriter(regionName, bucketName, reportFilename,
-                                    self.crossFileReportHeaders) as writer:
+            with self.getWriter(regionName, bucketName, reportFilename, self.crossFileReportHeaders) as writer, \
+                 self.getWriter(regionName, bucketName, warningReportFilename, self.crossFileReportHeaders) as warningWriter:
                 for failure in failures:
-                    writer.write(failure[0:7])
+                    if failure[9] == interfaces.validationDb.getRuleSeverityId("fatal"):
```
bsweger: Been thinking about how we can reduce the number db calls. What do you think about doing a lookup the severity ids of "fatal" and "warning" set storing the as variables before the `for failure in failures:` loop starts? Or pulling back the values of the severity table as a dict? Either way, we'd then only have to do one set of database calls.

dataactvalidator/validation_handlers/validator.py
```diff
@@ -64,7 +64,7 @@ def validate(cls,record,csvSchema, fileType, interfaces):
         Returns:
         Tuple of three values:
         True if validation passed, False if failed
-        List of failed rules, each with field, description of failure, and value that failed
+        List of failed rules, each with field, description of failure, value that failed, rule label, and severity
```
bsweger: 👍 for consistently keeping these docstrings up to date

dataactbroker/README.md
```diff
@@ -545,15 +545,91 @@ Example output:
   "job_3012_error_url": "https...",
   "job_3006_error_url": "https...",
   "job_3010_error_url": "https...",
-  "job_3008_error_url": "https..."
+  "job_3008_error_url": "https...",
+  "cross_appropriations-program_activity": "https..."
```
bsweger: Thank you for keeping the API docs up to date!

It looks like the _Process Overview_ and _Available Validations_ sections of this doc could also use some updates. Some of those have nothing to do with this PR (e.g., _Available Validations_), but some of the material in _Process Overview_ might be clearer if we mention errors vs warning.

In particular, it might be good to mention which types of schema-based failures result in fatal errors, and which will generate warnings. 

Anyway, if this type of update is out of scope for the story, we can put in a story to address it. What do you think?

dataactbroker/handlers/fileHandler.py
```diff
@@ -290,7 +303,8 @@ def getStatus(self):
                     else:
                         jobInfo["duplicated_headers"] = []
                     jobInfo["error_type"] = self.interfaces.errorDb.getErrorType(jobId)
-                    jobInfo["error_data"] = self.interfaces.errorDb.getErrorMetricsByJobId(jobId,self.jobManager.getJobType(jobId)=='validation',self.interfaces)
+                    jobInfo["error_data"] = self.interfaces.errorDb.getErrorMetricsByJobId(jobId,self.jobManager.getJobType(jobId)=='validation',self.interfaces, severityId=self.interfaces.validationDb.getRuleSeverityId("fatal"))
```
bsweger: Since this is in a `for job in jobs` loop, might be another opportunity to save the severity ids to variables before the loop to reduce the number of db calls (and also the `getJobType(jobId)=="validation"`).

dataactbroker/routeUtils.py
```diff
@@ -28,6 +28,8 @@ def run_instance_function(accountManager, accountFunction, getSystemEmail = Fals
                 else:
                     # Currently no functions with user but not credentials flag
                     raise ValueError("Invalid combination of flags to run_instance_function")
+            elif(addTrueFlag):
```
bsweger: This is a part of the code base I admittedly don't have a terrific handle on. What does this `addTrueFlag` do?

bsweger:
@jworcestBAH done reviewing!

dataactcore/models/errorInterface.py
jworcestBAH: Yep, the comment was from the first version of the function before it expanded to handle warnings as well, rephrased to be more general.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Agreed, switched to os.path.join

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Good catch, my intent was that warnings would not make a row invalid, since they will still be able to submit those rows.  I've added a check in the runSqlValidations function to only include rows with a fatal error in the returned list.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: The getRuleSeverityId function goes through getNameFromDict, which actually caches the dict and only does a single DB call to populate it.

dataactbroker/README.md
jworcestBAH: I clarified the text in process overview and specified that the length rule is a warning in available validations.  The available validations section is actually obsolete now as it's referring to the rule types we have deleted.  I'm thinking we should delete this section, but we may want to preserve the into on type, required, and length checks in some form.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Modified this function to only get job type once per job, as above the severity ID dict is cached after the first call.

dataactbroker/routeUtils.py
jworcestBAH: When this is true it passes "True" into the function being called as the first argument.  This is used for the warning reports route to use the same function as the error reports route with an extra argument to get warnings.  If it would be clearer, we could replace this by having the route call a separate function in fileHandler that would call getErrorReportURLsForSubmission with the warning flag.

bsweger: I'm a bit torn on this. The `addTrueFlag` wasn't super clear to me, but I'm not in the part of the code base much. Can we page @nmonga91 for a consult?

nmonga91: So the options are either use the flag here or have an additional function that would also call the original error reports function? Am I understanding that correctly?

jworcestBAH: Simplified run_instance_function to just pass through an argument list, it's much cleaner now.

dataactbroker/routeUtils.py
```diff
@@ -1,37 +1,20 @@
...
-from dataactbroker.handlers.aws.session import LoginSession
 
 class RouteUtils:
     """ Holds utility functions for routes """
     SYSTEM_EMAIL = None
     CREATE_CREDENTIALS = None
 
     @staticmethod
-    def run_instance_function(accountManager, accountFunction, getSystemEmail = False, getSession = False, getUser = False, getCredentials = False, addTrueFlag = False):
+    def run_instance_function(accountManager, accountFunction, *functionArgs): #getSystemEmail = False, getSession = False, getUser = False, getCredentials = False, addTrueFlag = False):
```
bsweger: love this cleanup...do we need the old params for future reference, or can we remove those?

bsweger:
Looks good. One bit of commented code that we can clean up later, but wanted to get this into the code base.

PR #167, opened 2016-08-01
Db 797 single db

jworcestBAH:
Merges databases into a single DB


tests/baseTestAPI.py
```diff
@@ -184,10 +181,7 @@ def setUp(self):
     def tearDownClass(cls):
         """Tear down class-level resources."""
         cls.interfaces.close()
-        dropDatabase(cls.interfaces.userDb.dbName)
-        dropDatabase(cls.interfaces.jobDb.dbName)
-        dropDatabase(cls.interfaces.errorDb.dbName)
-        dropDatabase(cls.interfaces.validationDb.dbName)
+        #dropDatabase(cls.interfaces.jobDb.dbName)
```
bsweger: Is a debug leftover?

tests/baseTestValidator.py
```diff
@@ -79,9 +76,7 @@ def setUp(self):
     def tearDownClass(cls):
         """Tear down class-level resources."""
         cls.interfaces.close()
-        dropDatabase(cls.interfaces.jobDb.dbName)
-        dropDatabase(cls.interfaces.errorDb.dbName)
-        dropDatabase(cls.interfaces.validationDb.dbName)
+        #dropDatabase(cls.interfaces.jobDb.dbName)
```
bsweger: debug leftover?

dataactcore/migrations/env.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import with_statement
 from alembic import context
+# Load all DB tables into metadata object
+from dataactcore.models import domainModels
```
bsweger: do we need these imports? I had this file open in PyCharm, which says that domainModels, errorModels, jobModels, stagingModels, userModel, and validationModels are unused.

bsweger:
@jworcestBAH Done reviewing. 6,872 lines of code deleted--best PR ever!

Just a few comments re: the tearDown classes in the unit tests. 

Lastly, I was unable to run the initialization scripts when I put a custom db name into the config.yml `db_name` field.

tests/baseTestAPI.py
jworcestBAH: Uncommented the drop statement.  As all the interfaces point at the same DB now, the drop here can be applied to any one of them.

tests/baseTestValidator.py
jworcestBAH: Same as above.

dataactcore/migrations/env.py
jworcestBAH: Without importing these, the Base metadata object does not get the information about the tables and alembic tries to drop them all.  The point of these imports is to load all the tables into the common Base object they all import, which we then get from the next line "from dataactcore.models import baseModel".

jworcestBAH:
Still looking into the custom name issue, I had some difficulties with alembic on this.

jworcestBAH:
Fixed the issue with runMigration for different db names

bsweger:
This all looks good. Feels only right to note that the original design of the db interfaces and models made this merge much less of a big deal than I'd fear. Nice!

Merging!

PR #168, opened 2016-08-02
add data migration script

bsweger:
I know we discussed not using a script for the data migration, but I still felt more comfortable putting together something repeatable and programmatic. This uses `pg_dump` and `pg_restore` to grab data from our old databases and insert it into its new home.

The script is in python simply so that we can grab the database credentials from `DB_CONFIG`. Otherwise, it's just executing postgres shell commands. It should be safe to run the shell commands and pipe data from the dump to the restore, since we're controlling the input.

Lastly, the script is repeatable. If you run it multiple times, it will throw errors on primary key violations.

### DIFF
--- a/dataactcore/scripts/migrateDataBroker.py
+++ b/dataactcore/scripts/migrateDataBroker.py
@@ -1,0 +1,68 @@
+# migrate data using pg_dump and pg_restore
+# data copied from tables:
+# error_data:
+#   error_metadata
+#   file
+# job_tracker:
+#   job
+#   submission
+#   job_dependency
+# user_manager;
+#   users
+#   email_token
+# validator:
+#   appropriation
+#   award_financial
+#   award_financial_assistance
+#   object_class_program_activity
+
+# run on command line: python migrateDataBroker.py
+
+from dataactcore.config import CONFIG_DB
+import subprocess
+
+c = 'postgresql://{}:{}@{}/'.format(
+    CONFIG_DB['username'], CONFIG_DB['password'], CONFIG_DB['host'])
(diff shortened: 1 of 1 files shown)


dataactcore/scripts/migrateDataBroker.py
```diff
@@ -0,0 +1,68 @@
...
+
+# error_data
+db = 'error_data'
+source = '{}{}'.format(c, db)
+print('migrating {}'.format(db))
+cmd = 'pg_dump -d {} -t error_metadata -t file --data-only --format=c | ' \
+    'pg_restore -d {} --data-only'.format(source, target)
+p = subprocess.call(cmd, shell=True, stdout=subprocess.PIPE)
+print('return code = {}\n'.format(p))
+
```
jworcestBAH: We would also need the 'file' table here for check_status to continue to work normally.  This table holds information about whether a file completed validation or stopped with header errors, etc.

bsweger: @jworcestBAH yes, the `pg_dump` syntax is a little odd...if you want multiple tables, you have to do `-t` multiple times, so the file table is there after `error_metadata`

jworcestBAH:
Thanks for doing this, will definitely simplify the process.  Just one table that needs to be added, after that I'll merge and run this on dev.

dataactcore/scripts/migrateDataBroker.py
jworcestBAH: Ah didn't see it, thanks

jworcestBAH:
Looks good, merging

PR #169, opened 2016-08-02
escape db password

bsweger:
Escape the db password with quotes to handle special characters


jworcestBAH:
Okay, merging since this works on non-windows

PR #170, opened 2016-08-02
SQL Validations for Rules B19 and C19

mab6bf:


klrBAH:
[run tests]

bsweger:
@mab6bf Hi Mike...these queries look like the should work just fine for getting a general sense of how many duplicates there are. But if we want to actually report the line numbers for the offending rows, we might need to get fancy. I played around with the C19 query, and came up with this (haven't tested extensively):

```sql
SELECT 
	row_number,
	beginning_period_of_availa,
	ending_period_of_availabil,
	agency_identifier,
	allocation_transfer_agency,
	availability_type_code,
	main_account_code,
	sub_account_code,
	object_class,
	program_activity_code,
	parent_award_id,
	piid,
	uri,
	fain,
	transaction_obligated_amou 
FROM (
  SELECT 
  	af.row_number,
  	af.beginning_period_of_availa,
  	af.ending_period_of_availabil, 
  	af.agency_identifier,
  	af.allocation_transfer_agency,
  	af.availability_type_code,
  	af.main_account_code,
  	af.sub_account_code,
  	af.object_class,
  	af.program_activity_code,
  	af.parent_award_id,
  	af.piid,
  	af.uri,
  	af.fain,
  	af.transaction_obligated_amou,
        -- get row numbers by the grouping of fields that defines uniqueness
  	ROW_NUMBER() OVER(PARTITION BY af.beginning_period_of_availa, af.ending_period_of_availabil, af.agency_identifier, af.allocation_transfer_agency, af.availability_type_code, af.main_account_code, af.sub_account_code, af.object_class, af.program_activity_code, af.parent_award_id, af.piid, af.uri, af.fain, af.transaction_obligated_amou) AS row
  FROM award_financial as af
  WHERE af.submission_id = 401
) dups
WHERE
	dups.row > 1
```

This is our first foray into report dupes, so there's no set way to go. I do think it would be useful to identify for people the specific row numbers that are failing the check.

jworcestBAH:
Incidentally, the way our code is handling this right now treats each result as a separate error that goes into the reports and error metadata, so if I'm reading the above correctly a duplicate would give us two rows in the report for the two row numbers involved, which seems fine, and two occurrences of a B19 violation (which kind of seems like it should be one).  I'm also wondering if we would prefer for the error report to have one line that gives a list of rows under "Row number", that might make it easier for the user to see what the error is rather than reporting each duplicates on a different row of the error report.

bsweger:
@jworcestBAH Where we landed was: we'll treat the first occurrence (as determined by row_id) of a given field combination (e.g., tas/object class/program activity/reimbursable flag) as the "original" record, and it won't show up on the error report. Subsequent occurrences of that combination of fields will be reported as errors.

What do you think?

I suspect there are a number of ways to make the error reports more useful, but before changing our current process of 1 sql result = 1 error report row, I'd like to hear some feedback.

bsweger:
@jworcestBAH To be clear, you raise a great point. I'm just in favor of getting the actual validation rule done per the story and then tweaking the error reports based on agency feedback. The sooner we get the rule in, the sooner we can get the feedback.

jworcestBAH:
Treating each subsequent repetition as an error seems like a good approach.  So if there are 4 rows in a set of duplicates, we'll report it as 3 errors.  Each one in the error report will have it's own row number, will it also get the row number it's duplicating from this query?

bsweger:
@jworcestBAH just following up here for transparency/completeness...yes, if there are 4 duplicate rows, we'd expect to see 3 of them in the error report. The row number that it's duplicating is not available in the query as currently written.

I heard through the grapevine that you had a good idea about a future enhancement that could provide said row number, which would be very cool.

jworcestBAH:
Integration tests now skip B19/C19, merging.

PR #172, opened 2016-08-02
Finalize route now allows users in the same agency as the submission

jworcestBAH:
Also added a unit test for this scenario

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -200,9 +200,11 @@ class FileHandler:
             # Compare user ID with user who submitted job, if no match return 400
             job = self.jobManager.getJobById(jobId)
             submission = self.jobManager.getSubmissionForJob(job)
-            if(submission.user_id != LoginSession.getName(session)):
+            # Check that user's agency matches submission cgac_code or user id of submission
+            userId = LoginSession.getName(session)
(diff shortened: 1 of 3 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -200,9 +200,11 @@ def finalize(self):
             # Compare user ID with user who submitted job, if no match return 400
             job = self.jobManager.getJobById(jobId)
             submission = self.jobManager.getSubmissionForJob(job)
-            if(submission.user_id != LoginSession.getName(session)):
+            # Check that user's agency matches submission cgac_code or user id of submission
+            userId = LoginSession.getName(session)
+            if(submission.user_id != userId and submission.cgac_code != self.interfaces.userDb.getUserByUID(userId).cgac_code):
```
bsweger: Is this something that people in the `SYS` agency should be able to do? Or folks who are website admins? (honest question--I don't remember the original requirements around this).

bsweger:
@jworcestBAH done reviewing!

dataactbroker/handlers/fileHandler.py
jworcestBAH: Good point, I modified it to allow "SYS" users to finalize any job.  I believe we've been using website admin and agency admin for permissions on the route decorator, which for this route any user should be able to reach the route.

bsweger:
Looks good--merging!

PR #174, opened 2016-08-04
Db 800 d file endpoints

nmonga91:


dataactbroker/handlers/fileHandler.py
```diff
@@ -389,3 +411,144 @@ def getRss(self):
...
+        url = "" if d1_file.url is None else d1_file.url
+        error_message = "" if d1_file.error_message is None else d1_file.error_message
+
+        response = {"status": status, "url": url, "start": str(d1_file.start_date), "end": str(d1_file.end_date),
+                    "message": error_message}
+
+        return JsonResponse.create(StatusCode.OK, response)
+
+
+    def generateD2File(self):
```
jworcestBAH: There seems to be quite a bit of duplicate code between the two generate functions and the two check functions.  Could they be combined with an argument specifying D1 or D2?

nmonga91: I was thinking the exact same once we finished up. I think we can make that a subtask of the story to actually generate the files and combine it at that time. Since this is dummied up right now, it's not as big of a deal, but I totally agree with you. We can combine it and accept a `d_file_type` which will dictate which one to process.

nmonga91: All their supporting functions are generalized to accept which d file should be operated on, so the rest of the code already supports it.

nmonga91:
[run tests]

dataactcore/utils/jobQueue.py
```diff
@@ -1,36 +1,90 @@
...
-		queueUrl = ''.join([CONFIG_JOB_QUEUE['broker_scheme'], '://', CONFIG_JOB_QUEUE['username'], ':', CONFIG_JOB_QUEUE['password'], '@', CONFIG_JOB_QUEUE['url'], ':', str(CONFIG_JOB_QUEUE['port']), '//'])
+        # Create remote connection to the job queue
+        self.jobQueue = Celery('tasks', backend=backendUrl, broker=queueUrl)
 
-		# Create remote connection to the job queue
-		self.jobQueue = Celery('tasks', backend=backendUrl, broker=queueUrl)
+        @self.jobQueue.task(name='jobQueue.enqueue')
+        def enqueue(jobID):
+            # Don't need to worry about the response currently
+            url = ''.join([validatorUrl, '/validate/'])
```
jworcestBAH: Did we decide not to use validate_threaded at some point?

nmonga91: When we originally put this in there, we had just set it up to call /validate. I noticed validate_threaded but it wasn't ever used so I didn't bother messing with it. We could utilize it, maybe that would help with the validations being blocked

klrBAH:
[run tests]

dataactcore/utils/jobQueue.py
```diff
@@ -1,36 +1,90 @@
...
+                bucket = CONFIG_BROKER['aws_bucket']
+                region = CONFIG_BROKER['aws_region']
+
+                aws_file_name = "".join([str(user_id), "/", timestamped_name])
+
+                with open(timestamped_name, "wb") as file:
+                    # get request
+                    response = requests.get(file_url)
+                    # write to file
+                    file.write(response.content)
```
jworcestBAH: This does seem to require holding the entire file in memory.  That may be unavoidable here, but we should consider checking how much memory the validator instances have and confirming that the D1/D2 files will never be larger than that.

dataactcore/utils/jobQueue.py
```diff
@@ -1,36 +1,90 @@
...
+                with open(timestamped_name) as file:
+                    for line in reader(file):
+                            lines.append(line)
+
+                headers = lines[0]
+                with CsvS3Writer(region, bucket, aws_file_name, headers) as writer:
+                    for line in lines[1:]:
+                        writer.write(line)
+                        writer.finishBatch()
+
```
jworcestBAH: Why are we calling finishBatch after each line?  That kind of defeats the purpose of batching these lines...

dataactcore/utils/jobQueue.py
nmonga91: We'd need to confirm that the job manager instance can hold that since it's being downloaded locally on that instance and then uploaded to s3. We can have a tech debt task for it to clean up after itself and delete the local file once upload is done

klrBAH: FYI, prod instances are m3.xlarge: 15GB RAM, if that helps.

Katie Rose
UX/UI Developer
Strategic Innovations Group
Booz | Allen | Hamilton
301-825-7974 (o)
201-919-5120 (c)
rose_katherine@bah.com<mailto:rose_katherine@bah.com>

dataactcore/utils/jobQueue.py
nmonga91: Correct, indentation should be moved out of the for loop but still within the "with" block. Fixing it now.

dataactbroker/handlers/fileHandler.py
```diff
@@ -389,3 +411,144 @@ def getRss(self):
...
+
+        if status == "finished" and not d1_file.is_submitted and False: # TODO: Remove "and False" when submission of D1 is desired
+            job_id = self.jobManager.getJobBySubmissionFileTypeAndJobType(submission_id, "award_procurement", "file_upload").job_id
+            self.jobManager.updateDFileName(submission_id, d1_file.upload_file_name, "award")
+            result = self.finalize(jobId=job_id)
+            self.jobManager.markDFileAsSubmitted(d1_file.d_file_id)
+            if result.status_code != 200:
+                raise ResponseException(result.data)
+
+        url = "" if d1_file.url is None else d1_file.url
```
jworcestBAH: So it seems this is created as a signed URL when the file is generated, if this check happens later the url may have already expired.  Is the frontend actually trying to get the file from this?

dataactbroker/handlers/fileHandler.py
jworcestBAH: Okay let's include that as part of that story then.

dataactbroker/handlers/fileHandler.py
nmonga91: Yes it is. It stores the download link. I think the expiration was like 14k seconds, can't remember where I saw that though but yes, that should be handled somehow. If they come back a month later, they won't be able to download those files again with that link

dataactcore/utils/jobQueue.py
jworcestBAH: In general we should be using validate_threaded, validate is helpful for debugging, which is probably why the older code was using it and we forgot to switch it back.

nmonga91: Understood. I'll make a note of that to make a story for it.

klrBAH:
[run tests]

klrBAH:
[unit tests]

klrBAH:
[run tests]

klrBAH:
[run tests]

jworcestBAH:
Okay, merging now

PR #175, opened 2016-08-04
fix for out-of-sync valid_record column

bsweger:
The data migration script failed on sandbox because migration that runs against the old staging tables to drop `valid_record` was removed from the migration history and never runs.

Thus, the old tables have that column and the new ones do not, causing the migration script to error. Added a patch that just removes that column in the old db if it's there. 

Also added the `--single-transaction` flag to the`pg_restore` command to ensure that we don't end up with any half-migrated tables.

Again, this script will only work if the target tables are empty, so we'll have to coordinate with @klrBAH when in the deploy process it should run.


nmonga91:
Looks good, merging once jenkins is done

PR #178, opened 2016-08-04
remove old rule documentation: now using the copy in the schema repo

bsweger:
Remove the rule tracker file from the broker repo to prevent confusion. We agreed that we can use this file instead in the schema repo: https://github.com/fedspendingtransparency/schema/blob/master/broker/Validation_Rules_FOR%20BROKER.xlsx


jworcestBAH:
Looks fine, will merge as soon as Jenkins finishes

PR #181, opened 2016-08-08
Adding format check for static files path and requirement checks for …

nmonga91:
…associated keys

@bsweger


bsweger:
Looks good!

PR #182, opened 2016-08-09
Db 881 publish status

jworcestBAH:
Submission table now includes publish status, publishable flag, number of errors and warnings


dataactcore/models/jobModels.py
```diff
@@ -31,6 +39,11 @@ class Submission(Base):
     reporting_end_date = Column(Date)
     is_quarter_format = Column(Boolean, nullable = False, default = "False", server_default= "False")
     jobs = None
+    publishable = Column(Boolean, nullable = False, default = "False", server_default = "False")
+    publish_status_id = Column(Integer, ForeignKey("publish_status.publish_status_id", ondelete="CASCADE", name ="fk_publish_status_id"))
```
bsweger: Do we want a cascading delete FK here?

jworcestBAH: Good point, "SET NULL" may be a better option here, I'll change to that

dataactcore/models/baseInterface.py
```diff
@@ -17,7 +17,7 @@ class BaseInterface(object):
     Session = None
     engine = None
     session = None
-
+    interfaces = None
```
cmc333333: Minor, but it might be helpful to reference where this is being set (i.e. `InterfaceHolder`) in a comment to make the dependency a bit more explicit

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+    def getSubmissionById(self, submissionId):
+        """ Return submission object that matches ID"""
+        query = self.session.query(Submission).filter(Submission.submission_id == submissionId)
+        return self.runUniqueQuery(query, "No submission with that ID", "Multiple submissions with that ID")
+
+    def getSubmissionErrorInfo(self, submissionId):
+        """ Set number of errors and warnings for submission """
+        submission = self.getSubmissionById(submissionId)
+        submission.number_of_errors = self.interfaces.errorDb.sumNumberOfErrorsForJobList(self.getJobsBySubmission(submissionId), self.interfaces.validationDb)
+        submission.number_of_warnings = self.interfaces.errorDb.sumNumberOfErrorsForJobList(self.getJobsBySubmission(submissionId), self.interfaces.validationDb, errorType = "warning")
```
cmc333333: Curious: are the `number_of_errors` and `number_of_warnings` ever read? Perhaps they are dynamically queried in a reporting interface or similar? If not, it looks like these columns are set but never used.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+        submission.publishable = publishable
+        self.session.commit()
+
+    def setPublishStatus(self, statusName, submissionId = None, submission = None):
+        """ Set publish status to specified name"""
+        statusId = self.getPublishStatusId(statusName)
+        if submission is None:
+            if submissionId is None:
+                raise ValueError("Must call setPublishStatus with either submission or submission ID")
+            submission = self.getSubmissionById(submissionId)
```
cmc333333: Minor suggestion for aiding static analysis tools:
```python
def setPublishStatus(self, statusName, submissionOrId):
    if isinstance(submissionOrId, str):
        submission = self.getSubmissionById(submissionOrId)
    else:
        submission = submissionOrId
```
Same number of lines but errors can be found by linters rather than waiting for the runtime.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+        submission.publish_status_id = statusId
+        self.session.commit()
+
+    def updatePublishStatus(self, submissionId = None, submission = None):
+        """ If submission was already published, mark as updated.  Also set publishable back to false. """
+        if submission is None:
+            if submissionId is None:
+                raise ValueError("Must call setPublishStatus with either submission or submission ID")
```
cmc333333: Typo here -- you're inside the `updatePublishStatus` method now. It brings up a question, though: why is this copy-pasted? Should it be moved to a shared utility?

dataactvalidator/interfaces/interfaceHolder.py
```diff
@@ -8,10 +9,17 @@ class InterfaceHolder:
...
-        self.jobDb = ValidatorJobTrackerInterface()
-        self.errorDb = ValidatorErrorInterface()
-        self.stagingDb = ValidatorStagingInterface()
-        self.validationDb = ValidatorValidationInterface()
+        if BaseInterface.interfaces is None:
+            self.jobDb = ValidatorJobTrackerInterface()
+            self.errorDb = ValidatorErrorInterface()
+            self.stagingDb = ValidatorStagingInterface()
+            self.validationDb = ValidatorValidationInterface()
+            BaseInterface.interfaces = self
```
cmc333333: I'm confused. Both `dataactvalidator.interfaces.InterfaceHolder` and `dataactbroker.handlers.InterfaceHolder` are setting a global object, but they're setting it to different values. Is the idea that `dataactvalidator` can never be used with `dataactbroker`? Is there a way to implement this without requiring that implicit conflict?

tests/baseTestValidator.py
```diff
@@ -213,7 +215,7 @@ def uploadFile(cls, filename, user):
 
             if(cls.uploadFiles) :
                 # Use boto to put files on S3
-                s3conn = S3Connection()
+                #s3conn = S3Connection()
```
cmc333333: Good call -- should this line be removed entirely?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -522,6 +526,12 @@ def runCrossValidation(self, jobId, interfaces):
         errorDb.writeAllRowErrors(jobId)
         interfaces.jobDb.markJobStatus(jobId, "finished")
         CloudLogger.logError("VALIDATOR_INFO: ", "Completed runCrossValidation on submissionID: "+str(submissionId), "")
+        # Update error info for submission
+        interfaces.jobDb.getSubmissionErrorInfo(submissionId)
+        # TODO: Remove temporary step below
+        # Temporarily set publishable flag at end of cross file, remove this once users are able to mark their submissions
+        # as publishable
+        interfaces.jobDb.setPublishableFlag(submissionId, True)
```
bsweger: It's possible (probable, in fact) that I'm misunderstanding what `pulblishableFlag` is supposed to represent. Because of its name, I took it to mean that a submission is publishable if it passed all the validations. But if I read this correctly, we're setting publishable to True regardless of validation outcome.

If this is the intended behavior (i.e., the data store is also going to check the # of errors to determine the "good" submissions), does it make sense to use a different name? validation_complete?

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+        """ Set publish status to specified name"""
+        statusId = self.getPublishStatusId(statusName)
+        if submission is None:
+            if submissionId is None:
+                raise ValueError("Must call setPublishStatus with either submission or submission ID")
+            submission = self.getSubmissionById(submissionId)
+        submission.publish_status_id = statusId
+        self.session.commit()
+
+    def updatePublishStatus(self, submissionId = None, submission = None):
```
bsweger: Since it looks like there's only one place this function is called, does it make sense to simplify the signature and `if` statements below by only accepting `submission` as a parameter? Or is `submissionId` an option because the data store will be using this at some point?

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+        """ If submission was already published, mark as updated.  Also set publishable back to false. """
+        if submission is None:
+            if submissionId is None:
+                raise ValueError("Must call setPublishStatus with either submission or submission ID")
+            submission = self.getSubmissionById(submissionId)
+        publishedStatus = self.getPublishStatusId("published")
+        if submission.publish_status_id == publishedStatus:
+            # Submission already published, mark as updated
+            self.setPublishStatus("updated", submission = submission)
```
bsweger: Do we care if a submission in updated status is updated again?

dataactcore/models/jobTrackerInterface.py
bsweger: @cmc333333 apologies for not providing more context here...the background of this PR is that a downstream process is building an API that will grab validated data from the broker. I suspect that these fields are for use by that process, but James can say for sure.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
...
+            jobId: Job to set number for
+            numberOfErrors: Number to be set
+            errorType: Type of error to set, can be either 'fatal' or 'warning'
+
+        """
+        job = self.getJobById(jobId)
+        if errorType == "fatal":
+            job.number_of_errors = numberOfErrors
+        elif errorType == "warning":
+            job.number_of_warnings = numberOfErrors
```
bsweger: do we need a `session.commit()` here?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -413,6 +415,8 @@ def runValidation(self, jobId, interfaces):
             # Mark validation as finished in job tracker
             jobTracker.markJobStatus(jobId,"finished")
             errorInterface.writeAllRowErrors(jobId)
+            # Update error info for submission
+            jobTracker.getSubmissionErrorInfo(submissionId)
```
bsweger: If I'm interpreting this correctly, it looks like this code will update a submission's error/warning count after each step of the validation process occurs. That is cool...wouldn't have thought about it w/ this level of detail.

dataactcore/models/jobTrackerInterface.py
bsweger: Do we need a `session.commit()` here?

dataactcore/models/jobTrackerInterface.py
```diff
@@ -285,3 +285,63 @@ def getSubmissionStatus(self,submissionId,interfaces):
             else:
                 return "validation_successful"
         return "unknown"
+
+    def getSubmissionById(self, submissionId):
+        """ Return submission object that matches ID"""
+        query = self.session.query(Submission).filter(Submission.submission_id == submissionId)
+        return self.runUniqueQuery(query, "No submission with that ID", "Multiple submissions with that ID")
+
+    def getSubmissionErrorInfo(self, submissionId):
```
bsweger: a minor thing, but should this function be called `setSubmissionErrorInfo` since it's doing an update?

bsweger:
@jworcestBAH Done reviewing. Thanks for bearing with 2 reviewers as @cmc333333 learns the code base.

So good to start laying the groundwork for the datastore hand off! Two questions in addition to the inline comments:
* Should we have tests for the new # of error/warning fields, or do you see that as something that would get done once we start to use those numbers (or as part of our in-progress test revamp)?
* Would it be helpful to update the validator documentation with a high-level explanation about how a submission is marked as "ready" for publication, or do you think it makes sense to wait until we implement the final submission process for real?

I can see the answer to both of the above being "not right now," just want to make sure we're making an explicit decision to wait.

dataactcore/models/baseInterface.py
jworcestBAH: Good idea, I've added a comment explaining that

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Good catch, added a commit.  For number_of_errors and number_of_warnings, data store folks asked for these in the meeting we had last week, I believe they want to use number_of_warnings to see how "clean" the data is since a submission can be published with warnings.

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Changed to this form and combined the arguments

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Good call, moved the new code out to extractSubmission

dataactvalidator/interfaces/interfaceHolder.py
jworcestBAH: The validator and broker were initially much more separate than they are now.  While it's still the case that they won't both be running in the same process, I've been thinking for awhile that we should combine these two in some way, the only issue is that the individual interfaces are still different for validator and broker.  I'm thinking I could create a common base class in the core package that has most of the functionality, and rename the existing ones to ValidatorInterfaceHolder and BrokerInterfaceHolder to match our conventions for individual interfaces.  The subclasses would then only be responsible for passing the list of interface objects to be used up to the base class.  Thoughts?

tests/baseTestValidator.py
jworcestBAH: Whoops, forgot to do that, good catch.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: The purpose of this flag is to represent whether the data store should actually publish the submission.  It is supposed to be set when the user clicks publish on the frontend (which they can only do after validating without errors).  Setting it here is a temporary workaround to let the data store have submissions be marked as ready to publish before the functionality is added on the frontend.  I could add a check on number of errors here so that only submissions without errors are marked as publishable, since that's a better reflection of the final process.

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Following CM's suggestion above, I've combined these into one parameter that can take either a submission or submissionId

dataactcore/models/jobTrackerInterface.py
jworcestBAH: No, only if it has gone back to "published" status by being copied to the data store.  The purpose of this status is that the data store handles changed submissions differently than new submissions, so when the publishable flag is true they check this status to see whether the submission has been previously published.

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Yes, added this.

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Yes, given the possibility of updating a single file in a submission, this seemed the best way of keeping the number of errors and warnings accurate during updates.

dataactcore/models/jobTrackerInterface.py
jworcestBAH: I wasn't sure here, normally for a set function I expect to be passing in the value it's set to.  Here I'm getting the information from the error tables and then setting it on the submission table, so neither name really seems to fit.  Maybe populateSubmissionErrorInfo?

dataactvalidator/interfaces/interfaceHolder.py
cmc333333: I don't know the class diagram well enough to comment on the proposed structure, but I'd recommend something dependency-injection-y. Maybe have a global variable which gets set on validator (and broker) init and refer to that here (please excuse the terrible variable names):

```python
# dataactbroker/app.py
from dataactcore.models.baseInterface import BaseInterface
from dataactbroker.handlers import InterfaceHolder
BaseInterface.interface =  InterfaceHolder

# dataactvalidator/app.py
from dataactcore.models.baseInterface import BaseInterface
from dataactvalidator.interfaces import InterfaceHolder
BaseInterface.interface =  InterfaceHolder
```

My concern with the current design revolves around a constructor mutating global state. As it stands, we're creating an implicit coupling between the two InterfaceHolders such that the order of their creation affects how they'll function. We _know_ that we shouldn't create them at the same time, but it's not explicit. I'd argue that setting the values in the init scripts is a step in the right direction.

That said, I don't think it should be a blocker if we're okay with that tech debt.

dataactcore/models/jobTrackerInterface.py
bsweger: Aaaah, I see what you're saying re: `set`. Yeah, I like `populateSubmissionErrorInfo` as a way to signal that this is updating the db.

jworcestBAH:
Just added some checks in test_check_status for number_of_errors and number_of_warnings.  For the documentation of the publication process, I think that would be clearer once we add the publication route so we're not guessing about how that will work.  That also is still prone to change once Arman starts using it if there are problems.

dataactvalidator/interfaces/interfaceHolder.py
jworcestBAH: I'm not quite clear on how this approach would work, where does the InterfaceHolder get instantiated?  To provide some background, we have found in the past that any time a new thread starts it needs to instantiate a separate InterfaceHolder to avoid conflicts in the session and connection objects.  So InterfaceHolder has needed to be instantiated in other places, and we'd need a way of dealing with that.  We could manually set BaseInterface.interface each time we instantiate a new one, but that seems to have the same problem as doing it in the constructor.

To expand a little on the current architecture, dataactcore is the common package used by dataactbroker and dataactvalidator, which should never reference each other or be referenced from the core.  The broker app uses only the core and broker, and the validator app uses only the core and validator, so neither should be accessing both InterfaceHolders.  These originally lived in different repos, these days the separation is only enforced by usage, so I agree that finding a way to better integrate the two is a good idea, I'm just not sure how to apply this approach. We could go the other direction and completely separate them, with BaseInterface having both a brokerInterface and a validatorInterface property, with each app only referring to the corresponding property.  That would resolve the immediate concern of a conflict between the two being possible.

cmc333333: > any time a new thread starts it needs to instantiate a separate InterfaceHolder to avoid conflicts in the session and connection objects.

Gotcha. That's wading much deeper than I anticipated. At some point, we'll probably want to replace these with `flask.g`: http://flask.pocoo.org/docs/0.11/api/#application-globals, but that can can be kicked down the road a bit.

bsweger:
Thanks for the updates @jworcestBAH. I think we're good to go on this one. I very much like @cmc333333's suggestion for using Flask application globals, but for now I'd rather merge this as is, because we have some other db migrations piling up.

I added a [tech debt task](https://federal-spending-transparency.atlassian.net/browse/DB-935) to JIRA for the application globals (DB-935).

Good code reviewing, everyone!

bsweger:
(note: will merge once folks are done troubleshooting some things on dev)

PR #183, opened 2016-08-09
Reformatting dates in response for D1/D2 generation routes

nmonga91:


dataactcore/utils/jobQueue.py
```diff
@@ -34,7 +34,7 @@ def enqueue(jobID):
             response = requests.post(url, params)
             return response.json()
 
-        @self.jobQueue.task(name='jobQueue.generate_d1')
+        @self.jobQueue.task(name='jobQueue.generate_d_file')
```
bsweger: @nmonga91 I don't know enough about how the jobQueue works to understand if this change impacts validation functionality or if it's just a change to improve the the task name. Can you clarify? Thanks!

nmonga91: This is just the task name that's published when it runs. Below, it actually uses the correct task name to create the method reference. The method reference below is what's actually used but I wanted to update the task name to be consistent with the method name. Shouldn't have any affect otherwise

bsweger:
looks good--merging!

PR #184, opened 2016-08-09
Hotfix d file date format

nmonga91:


dataactcore/utils/jobQueue.py
```diff
@@ -34,7 +34,7 @@ def enqueue(jobID):
             response = requests.post(url, params)
             return response.json()
 
-        @self.jobQueue.task(name='jobQueue.generate_d1')
+        @self.jobQueue.task(name='jobQueue.generate_d_file')
```
jworcestBAH: Just to make sure I'm understanding this correctly, it looks like the function below takes which D file to generate as an argument?

nmonga91: correct, which is specified by `d_file_id` which is in the d_file_metadata db table

jworcestBAH:
Looks good, merging

PR #186, opened 2016-08-09
Resolve two small getting-started issues

cmc333333:
See commit comments for details


bsweger:
First DATA Act commit--welcome to the team!

Thanks for taking the time to improve the docs as you get set up...changes look good.

PR #189, opened 2016-08-10
Hotfix sys submission tagging

nmonga91:
"SYS" agency is now an option for submissions if the current, logged in user is part of the "SYS" agency

### DIFF
--- a/dataactbroker/handlers/domainHandler.py
+++ b/dataactbroker/handlers/domainHandler.py
@@ -23,7 +23,9 @@ class DomainHandler:
         self.userManager = interfaces.userDb
 
     def listAgencies(self):
-
+        """ Retrieves a list of all agency names and their cgac codes. If there is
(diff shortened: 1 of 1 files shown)


jworcestBAH:
I'm wondering if it would be better to do the append outside this function, just because this function looks like something that would potentially be reused if we want to list agencies for anything else.  If we don't do that, could we put a docstring on the listAgencies function warning about this exception so we don't accidentally misuse it?  Alternately, a checkForSys flag that defaults to false would perhaps make this clearer.

nmonga91:
So right now it's also being used for registration, but without an x-session-id the `isLogin` check returns false and skips it. I wouldn't mind moving it out but `getAllAgencies` should be the function being reused vs this one that's specifically for a route. I can also add the docstring

jworcestBAH:
Okay let's just add the docstring then, and have that mention that this adds SYS for those users

nmonga91:
@jworcestBAH docstring has been added

PR #191, opened 2016-08-10
Hotfix: Generating signed s3 urls on the fly and removing them from the database

nmonga91:


jworcestBAH:
Okay everything looks fine, only minor nitpick is that if possible I would prefer any name other than "public", given that the folder is not in fact public.

dataactbroker/handlers/fileHandler.py
```diff
@@ -472,9 +472,13 @@ def checkD1File(self, submission_id=None):
             if result.status_code != 200:
                 raise ResponseException(result.data)
 
-        url = "" if d1_file.url is None else d1_file.url
-        error_message = "" if d1_file.error_message is None else d1_file.error_message
+        user_id = LoginSession.getName(session)
+        slash_index = d1_file.upload_file_name.rfind("/")
+        timestamped_filename = d1_file.upload_file_name[slash_index + 1:]
```
cmc333333: Nothing worth holding this PR back, but this might be a little cleaner:

```python
timestamped_filename = d1_file.upload_file_name.split('/')[-1]
```

nmonga91: ah that is clever, did not think of that. Will make that change as part of another branch.

dataactbroker/handlers/fileHandler.py
```diff
@@ -472,9 +472,13 @@ def checkD1File(self, submission_id=None):
...
 
+        url = "" if status != "finished" else self.s3manager.getSignedUrl(path=str(user_id),
+                                                                          fileName=timestamped_filename, method="GET")
+        error_message = "" if d1_file.error_message is None else d1_file.error_message
```
cmc333333: Another small thing that's a bit easier to read:
```python
error_message = d1_file.error_message or ""
```

nmonga91: That has also been noted!

dataactbroker/handlers/fileHandler.py
```diff
@@ -472,9 +472,13 @@ def checkD1File(self, submission_id=None):
...
                 raise ResponseException(result.data)
 
-        url = "" if d1_file.url is None else d1_file.url
-        error_message = "" if d1_file.error_message is None else d1_file.error_message
+        user_id = LoginSession.getName(session)
+        slash_index = d1_file.upload_file_name.rfind("/")
+        timestamped_filename = d1_file.upload_file_name[slash_index + 1:]
 
+        url = "" if status != "finished" else self.s3manager.getSignedUrl(path=str(user_id),
+                                                                          fileName=timestamped_filename, method="GET")
```
cmc333333: Probably not important as this worked before, but in some of the calls to `getSignedUrl`, the first param has a trailing slash and in others it doesn't

nmonga91: For the path parameter, it works with and without so I think it's indifferent as to if it's provided or not. But we should be consistent either way.

nmonga91:
[run tests]

cmc333333:
:+1: dynamically generating the signed url should solve the expiry problem

PR #192, opened 2016-08-10
Reordered checks for token permissions

jworcestBAH:
This was hitting a bug for logged in users checking a token permission, as it tried to treat it like a user permission.


jworcestBAH:
[run tests]

nmonga91:
Looks good, merging

PR #195, opened 2016-08-10
File download comes from user's folder based on who generated the fil…

nmonga91:
…e for the submission


jworcestBAH:
Looks good, will merge as soon as Jenkins approves

PR #199, opened 2016-08-11
Db 862 sf133 schema

bsweger:
This is a small update to add `period` to the unique index on the `sf_133` table. PRing this first so @nmonga91 is able to bulk load the historical data to his local db and continue with DB-712 (GTAS validations).

The PR also includes some fixes to our migration chain.


dataactcore/migrations/versions/a0a4f1ef56ae_update_sf133_unique_index.py
```diff
@@ -0,0 +1,42 @@
...
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index('ix_sf_133_tas', table_name='sf_133')
+    op.create_index('ix_sf_133_tas', 'sf_133', ['tas', 'period', 'line'], unique=True)
```
nmonga91: This is creating the composite key that is composed of tas, period, AND line, correct?

bsweger: Yep...here's the actual DDL from my local db: 
```sql
CREATE UNIQUE INDEX "ix_sf_133_tas" ON "public"."sf_133" USING btree( "tas" ASC NULLS LAST, "period" ASC NULLS LAST, "line" ASC NULLS LAST );
```

nmonga91:
Looks good, merging!

PR #200, opened 2016-08-11
Hotfix: allowing other users to download regenerated D files and updating start/end dates

nmonga91:


bsweger:
Looks good--merging

PR #201, opened 2016-08-11
Hotfix: allowing other users to download regenerated D files and updating start/end dates

nmonga91:


bsweger:
@nmonga91 I did this merge locally and all looked good re: the alembic chain. Merging!

PR #202, opened 2016-08-11
change fiscal_year and period sf133 cols to int

bsweger:
Update the `period` and `fiscal_year` columns in `sf_133`:
* don't allow NULLs
* change datetype from text to integer

*Note:* you'll see some bespoke `op.execute` statements in the migration..that's to make the datatype change work if the table has existing data.


nmonga91:
Looks good, merging!

PR #203, opened 2016-08-11
Small tweaks to filesize tests

cmc333333:
The first changeset is the functional part; the second is just some cleanup


bsweger:
@cmc333333 Done reviewing. This looks great--thanks!

I ran the revised tests locally (don't think that Jenkins exercises the `local=True` path) and all is well.

PR #204, opened 2016-08-11
use pandas for domain values load and update SF-133 loader

bsweger:
Wanted to show the changes in progress to our domain load process. In particular, wanted to get @jworcestBAH's feedback on the work is divided between `loadFile.py` and `loaderUtils.py`.

The basic thought process was...for each set of domain values:
* delete existing database rows, if necessary (this varies depending on what's being loaded)
* read the .csv
* call the common data cleaning function
* do any post-processing specific that particular set of domain values (e.g., tas field in SF-133)
* call common db insert routine function


jworcestBAH:
The approach seems fine but I'm wondering if we could move the csv loading and deletion of previous records into LoaderUtils, those seems to be pretty similar for each loadFile function.

bsweger:
@jworcestBAH thanks for taking a look--I moved that function to `loaderUtils.py`. I also re-instated the `strip_commas` option (which I'd missed) and removed the `skip_duplicates` option. 

The logic for that latter decision is that how do de-dupe a file is specific to the file, and because sometimes we need to de-dupe on multiple columns, squeezing that into the current options structure became awkward.

bsweger:
This is still a WIP because there will be conflicts once the unittest branch is merged. In addition, I need to add a check to zero out specific SF 133 line numbers if necessary (still have to look at the data to see if we need to do that).

That said, what's here so far is ready for review. There's substantial changes to how we're loading domain values, so if folks can review this now, reviewing the remaining work should be a lot faster.

dataactvalidator/filestreaming/csvS3Reader.py
```diff
@@ -24,18 +18,23 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
         s3Bucket = s3connection.lookup(bucket)
         self.s3File = s3Bucket.lookup(filename)
         self.isLocal = False
-        if(self.s3File == None):
-            raise ValueError("".join(["Filename provided not found on S3: ",str(filename)]))
+        if not self.s3File:
```
cmc333333: Very minor, but recommended practice for testing against `None` (as opposed to `False`, the empty string, etc) is:
```python
if self.s3File is None:
```

dataactvalidator/filestreaming/csvS3Reader.py
```diff
@@ -24,18 +18,23 @@ def openFile(self,region,bucket,filename,csvSchema,bucketName,errorFilename):
...
         s3Bucket = s3connection.lookup(bucket)
+        if not s3Bucket:
+            raise ValueError("Bucket {} not found in region {}".format(
+                bucket, region))
         self.s3File = s3Bucket.get_key(filename)
-        if(self.s3File == None):
-            raise ValueError("".join(["Filename provided not found on S3: ",str(filename)]))
+        if not self.s3File:
+            raise ValueError("Filename {} not found in bucket {}".format(
+                filename, bucket))
```
cmc333333: This function is almost identical to `openFile`; would it make sense for one to call the other or to pull out the shared code into a `initializeS3File` method?

dataactvalidator/filestreaming/csvS3Reader.py
```diff
@@ -53,9 +52,10 @@ def _getNextPacket(self):
         Gets the next packet from the file returns true if successful
         """
         offsetCheck = self.packetCounter *  CsvAbstractReader.BUFFER_SIZE 
-        header ={'Range' : "".join(['bytes=',str(offsetCheck),'-',str(offsetCheck +CsvAbstractReader.BUFFER_SIZE - 1)])}
+        header ={'Range': "".join(
+            ['bytes=', str(offsetCheck), '-', str(offsetCheck +CsvAbstractReader.BUFFER_SIZE - 1)])}
```
cmc333333: Super minor, but this is another candidate for `str.format()`

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    model = CGAC
+
+    # for CGAC, delete and replace values
+    interface.session.query(model).delete()
+    interface.session.commit()
+
+    # read CGAC values from csv
+    data = pd.read_csv(filename, dtype=str)
+    # toss out rows with missing CGAC codes
+    data = data[data['CGAC'].notnull()]
```
cmc333333: Slick!

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
+                data[col] = data[col].str.replace(",", "")
 
+        # add created_at and updated_at columns
+        data = data.assign(
+            created_at=cls.currentTimeFunction, updated_at=cls.currentTimeFunction)
 
+        return data
+
+    @classmethod
+    def insertDataframe(cls, df, table, engine):
```
cmc333333: Hmmm. I'm torn here. I see benefit in allowing `pandas` to do its thing wrt SQL generation, but I'm concerned about adding a non-ORM route to manipulating the data.

Is there a place to put this that'd fit into the ORM a bit better? Maybe within the ORM models or on the relevant "interface" object? This might also allow you to add a validation that the right columns are present.

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    interface = ValidatorValidationInterface()
+    model = ObjectClass
+
+    # for object class, delete and replace values
+    interface.session.query(model).delete()
+    interface.session.commit()
+
+    data = pd.read_csv(filename, dtype=str)
+    # toss out blank rows
+    data.dropna(inplace=True)
```
cmc333333: It might be nice to document why we're using `dropna` here, `notnull` elsewhere, and neither in other places.

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    # insert to db
+    LoaderUtils.insertDataframe(data, model.__table__.name, interface.engine)
+
+
+def loadSF133(filename, fiscal_year, fiscal_period, force_load=False):
+    interface = ValidatorValidationInterface()
+    model = SF133
+
+    existing_records = interface.session.query(model).filter(
+        and_(model.fiscal_year == fiscal_year, model.period == fiscal_period))
```
cmc333333: Curious: do you need the `and_` here? It looks like it's implicit: http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.filter

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    if force_load:
+        # force a reload of this period's current data
+        print('Force SF 133 load: deleting existing records for {} {}'.format(
```
cmc333333: In general, `logging` is better than `print`s; it allows more configuration, storage/redirection, etc. Can we start that trend here? Something like:

```python
import logging
# ...
# ...
logger.info('Force SF 133 load: deleting existing records for %s %s',
            fiscal_year, fiscal_period)
```

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+         "agency_identifier": {"pad_to_length": 3},
+         "main_account_code": {"pad_to_length": 4},
+         "sub_account_code": {"pad_to_length": 3},
+         "amount": {"strip_commas": True}}
+    )
+    # todo: find out how to handle dup rows (e.g., same tas/period/line number)
+    # line numbers 2002 and 2012 are the only duped line numbers,
+    # and they are not used by the validation rules, so for now
+    # just remove them before loading our SF-133 table
+    dupe_line_numbers = ['2002', '2102']
```
cmc333333: Thanks for the comments! I don't understand them yet, but they're much better than having magic numbers

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    data = data[~data.line.isin(dupe_line_numbers)]
+
+    # add concatenated TAS field for internal use (i.e., joining to staging tables)
+    data['tas'] = data.apply(lambda row: formatInternalTas(row), axis=1)
```
cmc333333: Also slick!

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    # insert to db
+    LoaderUtils.insertDataframe(data, model.__table__.name, interface.engine)
+
+    # todo: insert 0 line numbers if necessary for validation rules
+
+
+def formatInternalTas(row):
+    """Concatenate TAS components into a single field for internal use."""
+    # This formatting should match formatting in dataactcore.models.stagingModels concatTas
+    tas = '{}{}{}{}{}{}{}'.format(
```
cmc333333: This is a case where the `''.join` approach might actually be easier to read

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+    print("Loading program activity")
+    if localProgramActivity is not None:
+        loadProgramActivity(localProgramActivity)
+    else:
+        loadProgramActivity(os.path.join(basePath, "program_activity.csv"))
 
     # SF 133 is kept on S3, so need to download that
     reader = CsvS3Reader()
 
     if localSFPath is not None:
```
cmc333333: There's a lot of new logic between these two branches of the `if` -- would it make sense to chunk them out into separate methods/functions?

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+        sf133Files = glob.glob(os.path.join(localSFPath, 'sf_133*.csv'))
+        for sf133 in sf133Files:
+            file = os.path.basename(sf133).replace('.csv', '')
+            fileParts = file.split('_')
+            if len(fileParts) < 4:
+                print('{}Skipping SF 133 file with invalid name: {}'.format(
+                    os.linesep, sf133))
+                continue
+            year = file.split('_')[-2]
+            period = file.split('_')[-1]
```
cmc333333: This text munging doesn't feel good. Would a regex be easier to understand?

```python
SF_RE = re.compile(r'sf_133_(?P<year>\d{4})_(?P<period>\d+)\.csv')
# ... 
match = SF_RE.match(sf133)
if not match:
    # log error
    continue
loadSF133(sf133, match.group('year'), match.group('period'))
```

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+                    CONFIG_BROKER['sf_133_folder'])[-1].replace('.csv', '')
+                fileParts = file.split('_')
+                if len(fileParts) < 4:
+                    print('{}Skipping SF 133 file with invalid name: {}'.format(
+                        os.linesep, sf133))
+                    continue
+                year = file.split('_')[-2]
+                period = file.split('_')[-1]
+                print('{}Starting {}...'.format(os.linesep,sf133.name))
+                loadSF133(sf133, year, period)
```
cmc333333: This is _very_ similar to the `localSFPath is not None` path above. Can they be combined?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
+    def cleanData(cls, data, model, fieldMap, fieldOptions):
+        """ Cleans up a dataframe that contains domain values
 
         Args:
-            filename: CSV to load
-            model: ORM object for table to be loaded
-            interface: interface to DB table is in
-            fieldMap: dict that maps columns of the csv to attributes of the ORM object
+            data = dataframe of domain values
+            fieldMap: dict that maps columns of the dataframe csv to our db columns
```
cmc333333: I recognize that the `fieldMap` wasn't introduced in this PR, but it feels hairy. Is there a way to define those values statically instead of at method invocation time?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
             fieldOptions: dict with keys of attribute names, value contains a dict with options for that attribute.
-                Current options are "pad_to_length" which if present will pad the field with leading zeros up to
-                specified length, and "skip_duplicate" which ignores subsequent lines that repeat values.
+                Current options are:
+                 "pad_to_length" which if present will pad the field with leading zeros up to
+                specified length
+                "skip_duplicate" which ignores subsequent lines that repeat values
+                "strip_commas" which removes commas
```
cmc333333: Again, I know these options weren't introduced here, but this sounds like a great use case for a `namedtuple` rather than a dict

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
-                    try:
-                        interface.session.merge(record)
-                        interface.session.commit()
+        # toss out blank rows
+        data.dropna(inplace=True, how='all')
+
+        # Fix up cells that have spaces instead of being empty.
+        # Set the truly empty cells to None so they get inserted to db as NULL
+        # TODO: very ugly function below...is there a better way?
+        data.applymap(lambda x: str(x).strip() if len(str(x).strip()) else None)
```
cmc333333: Maybe two steps?
```python
data.applymap(lambda x: str(x).strip())
data.applymap(lambda x: x or None)
```
Pandas might have some helpful function, too?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -1,10 +1,13 @@
...
-from dataactvalidator.validation_handlers.validator import Validator
+from datetime import datetime
 
 class LoaderUtils:
 
+    # define some data-munging functions that can be applied to
+    # pandas dataframes as necessary
+    padFunction = lambda field, padTo: str(field).strip().zfill(padTo)
+    currentTimeFunction = lambda x: datetime.utcnow()
+    cleanColNamesFunction = lambda field: str(field).lower().strip().replace(" ","_").replace(",","_")
```
cmc333333: Any reason these are static variables as opposed to straight functions? They don't seem too tightly coupled with `LoaderUtils` and a `def` is generally easier to read than a `lambda`

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
+        # Set the truly empty cells to None so they get inserted to db as NULL
+        # TODO: very ugly function below...is there a better way?
+        data.applymap(lambda x: str(x).strip() if len(str(x).strip()) else None)
+
+        # clean the dataframe column names
+        data.rename(columns=lambda x: cls.cleanColNamesFunction(x), inplace=True)
+        # make sure all values in fieldMap parameter are in the dataframe/csv file
+        for field in fieldMap:
+            if field not in list(data.columns):
+                raise ValueError("{} is required for loading table{}".format(field, model))
```
cmc333333: Noting that this is a prime example for why the `fieldMap` approach is problematic. Rather than relying of methods/functions with named arguments to verify the parameters at call-time (and potentially before, using static analysis), we're throwing an exception half way through this function.

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
-                        print("".join(["Warning: Skipping this row: ",str(row)]))
-                        print("".join(["Due to error: ",str(e)]))
-                        interface.session.rollback()
-                        continue
-            interface.session.commit()
+        # apply column options as specified in fieldOptions param
+        for col, options in fieldOptions.items():
+            if "pad_to_length" in options:
+                # pad to specified length
+                data['{}'.format(col)] = data['{}'.format(col)].apply(
```
cmc333333: Can these indexes be shorted to `data[col]` or `data[str(col)]`?

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
+                    lambda x: cls.padFunction(x, options['pad_to_length']))
```
cmc333333: NBD, but if we're passing around lots of functions and lambdas, it might make sense to curry the parameters. For example:
```python
def mk_pad_fn(length):
    return lambda field: str(field).strip().zfill(length)
# ...
data[...] = data[...].apply(mk_pad_fn(options['pad_to_length']))
```

This should also possible with `functools.partial`, though it feels less readable:
```python
data[...] = data[...].apply(partial(
    cls.padFunction, padTo=options['pad_to_length']))
```

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -25,85 +28,62 @@ def compareRecords (recordA,recordB, fields) :
...
+            if 'strip_commas' in options and options['strip_commas']:
```
cmc333333: Little thing, but this is slightly more Pythonic:
```python
if options.get('strip_commas'):
```

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -1,53 +1,216 @@
...
+        row['ending_period_of_availabil'] if row['ending_period_of_availabil'] else '0000',
+        ' ' if pd.isnull(row['availability_type_code']) else row['availability_type_code'],
+        row['main_account_code'] if row['main_account_code'] else '0000',
+        row['sub_account_code'] if row['sub_account_code'] else '000')
+    return tas
 
 
 def loadDomainValues(basePath, localSFPath = None, localProgramActivity = None):
-    """ Load all domain value files, localSFPath is used to point to a SF-133 file, if not provided it will be downloaded from S3  """
+    """Load all domain value files, localSFPath is used to point to a SF-133 file, if not provided it will be downloaded from S3."""
```
cmc333333: `localSFPath` no longer points to a file, but a directory. We should definitely update this docstring, but may also want to modify the variable name.

cmc333333:
Added lots of little comments, but I think this is largely cleaner than what existed before. Unfortunately, tests are currently failing.

cmc333333:
[run-tests]

dataactvalidator/filestreaming/loadFile.py
bsweger: Good point. This `dropna` (oof, which should actually be using the `how=all` parameter) is intended to handle the case where an incoming .csv has a bunch of extraneous blank rows at the end. That's universal enough that I moved (the corrected version of) this line to the `cleanData` function in `loaderUtils.py`.

dataactvalidator/filestreaming/loadFile.py
bsweger: As it turns out, I added this to handle a missing CGAC code value in our .csv. That missing value should actually be `000`, so ended up removing this line.

dataactvalidator/filestreaming/loadFile.py
bsweger: You're right--thanks..will take that out.

dataactvalidator/filestreaming/loadFile.py
bsweger: Yes, great idea. There are several functions in loadFile.py that print, and I got the logging working by setting up the logger at the beginning of each.

What's the recommended approach to avoid this repetitive code? Put all this stuff in a class? Decorator? Global variable of some kind?

cmc333333: The strategy I learned under @vrajmohan and @jmcarp's excellent tutelage is to create a module-level `logger` and configure everything at the app's entry points. It'd look something like this:

```python
# my/module/here.py
import logging

logger = logging.getLogger(__name_)

def some_func():
    logger.info("Some Func called")    # or debug, warning, etc.

# run.py
import logging

if __name__ == '__main_':
    logging.basicConfig()    # Can be configured with threshold level here
    runProgram()
```

A better long-term strategy would be to include config information in the project's settings so it can be configured without modifying code -- Django does this, but Flask's intentionally minimal.

bsweger: If @vrajmohan and @jmcarp recommend this approach, then it shall be so. Thanks!

dataactvalidator/filestreaming/loaderUtils.py
bsweger: So, namedtuples seem pretty cool. What I'm not sure of is how, in this scenario, they would help us catch this error at call time.

The thing I don't see yet is how switching the `cleanData` params from dictionaries to named tuples will prevent having to do this check, which is making sure that the passed items are supplied as headers in the incoming .csv (which need to be cleaned up a bit before the check).

Can you elaborate?

dataactvalidator/filestreaming/loaderUtils.py
bsweger: Yes, yes they can.

dataactvalidator/filestreaming/loadFile.py
bsweger: I reworded the comment slightly to make it clearer that the line numbers in question are not surrogate keys or line numbers in a file...they represent specific designations on a budget report that really shouldn't be duplicated (however, we do not validate against those lines, hence the skipping until we dig a little deeper).

dataactvalidator/filestreaming/loaderUtils.py
bsweger: Thanks for taking a look at this. The applymap is pretty expensive to do on these large dataframes, so I'm hesitant to do it twice. I've been through the Pandas stuff and couldn't find anything magic that will handle the case where a cell has spaces. I did play with setting whitespace cells to NaN when the file was read, but that cascading into some other issues.

In conclusion--it would be good to make this process less fragile but am hesitant to pull this thread.

dataactvalidator/filestreaming/loaderUtils.py
bsweger: I'm torn too. We had discussed the ramifications of bypassing the ORM shortly before joined us, and agreed that we could do so as part of the domain values load.

That said, this did bite me a few times during development and testing, so I'm amenable to re-thinking this. However, I'm thinking that the benefits of getting this in sooner so we can test the pieces of the load and making sure the deploy works with all this new data outweigh the costs of refactoring at this point.

What do you think?

dataactvalidator/filestreaming/loaderUtils.py
cmc333333: Ah, I misread this as validating `fieldMap` rather than validating `data.columns`. Sorry for the confusion!

dataactvalidator/filestreaming/loaderUtils.py
cmc333333: Sounds good. To be super clear, these comments are meant to be suggestions only. I think the only thing that's blocking this PR is tests at this point.

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -130,30 +138,55 @@ def loadSF133(filename, fiscal_year, fiscal_period, force_load=False):
...
     # add concatenated TAS field for internal use (i.e., joining to staging tables)
     data['tas'] = data.apply(lambda row: formatInternalTas(row), axis=1)
+
+    # zero out line numbers not supplied in the file
+    pivot_idx = ['created_at', 'updated_at', 'agency_identifier', 'allocation_transfer_agency',
+                 'availability_type_code', 'beginning_period_of_availa', 'ending_period_of_availabil',
+                 'main_account_code', 'sub_account_code', 'tas', 'fiscal_year', 'period']
+    data.amount = data.amount.astype(float)  # this line triggers the settingwithcopy warning
+    data = pd.pivot_table(data, values='amount', index=pivot_idx, columns=['line'], fill_value=0).reset_index()
+    data = pd.melt(data, id_vars=pivot_idx, value_name='amount')
```
cmc333333: So many new terms to learn, like `melt`!

dataactvalidator/filestreaming/loaderUtils.py
```diff
@@ -65,7 +73,7 @@ def cleanData(cls, data, model, fieldMap, fieldOptions):
             if "pad_to_length" in options:
                 # pad to specified length
                 data['{}'.format(col)] = data['{}'.format(col)].apply(
-                    lambda x: cls.padFunction(x, options['pad_to_length']))
+                    cls.padFunction, args=(options['pad_to_length'],))
```
cmc333333: This threw me for a loop; I expected `args` to be the _first_ arguments to the function rather than the last. Good to know!

dataactvalidator/filestreaming/loadFile.py
```diff
@@ -199,9 +201,37 @@ def formatInternalTas(row):
...
+        if CONFIG_BROKER["use_aws"]:
+            # get list of SF 133 files in the config bucket on S3
+            s3connection = boto.s3.connect_to_region(CONFIG_BROKER['aws_region'])
+            s3bucket = s3connection.lookup(CONFIG_BROKER['aws_bucket'])
+            # get bucketlistresultset with all sf_133 files
+            sf133_files = s3bucket.list(
+                prefix='{}/sf_133'.format(CONFIG_BROKER['sf_133_folder']))
+            sf133_list = [SF133File(sf133, os.path.basename(sf133.name)) for sf133 in sf133_files]
+
+    return sf133_list or []
```
cmc333333: There's a branch where the `sf133_list` variable isn't defined, but it may not be sensible:
`localSF133Dir is None` and `CONFIG_BROKER["use_aws"] == False`

cmc333333: Also, very nice refactor!

bsweger:
@cmc333333 Okay, I made a lot of your suggested changes. Could keep tweaking, but I think the time has come for another look for things that need to get fixed before merge.

This is still WIP status while I check in w/ Katie for a heads-up on the one-time load of some historical data.

dataactvalidator/filestreaming/loadFile.py
bsweger: Thanks--will fix the branch thing. At some point, I would like to know more about what we should expect to happen for various permutations of the `use_aws` param and the local option, but will pull that thread another day.

cmc333333:
I really like the dataframe approach; I can't validate that business logic is identical but that's what tests are for :)

PR #206, opened 2016-08-11
Add some sql-rule validations + skeleton for more

cmc333333:
Rough cut; we'll probably want to need to add more setup information, but this
should get us started. See test_a16_appropriations for example usage.

conftest.py adds a fixture related to database setup (monkeying what is in the
baseTestValidator.py). We'll probably also want to use something like
factory-boy to make generating example models easier.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -1,0 +1,27 @@
+from random import randint
+
+import pytest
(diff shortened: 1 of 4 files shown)


nmonga91:
Looks good, merging!

PR #207, opened 2016-08-12
DB-570 (A30, A32, A33, B20) and DB-859 (A18 and A19)

bsweger:
Getting this out as PR to start the conversation while @mab6bf is OOO.


dataactvalidator/config/sqlrules/a18_cross_file.sql
```diff
@@ -1,10 +1,26 @@
...
-ON approp.tas = op.tas AND approp.submission_id = op.submission_id
+	approp.row_number,
+	approp.allocation_transfer_agency,
+	approp.agency_identifier,
+	approp.beginning_period_of_availa,
+	approp.ending_period_of_availabil,
+	approp.availability_type_code,
+	approp.main_account_code,
+	approp.sub_account_code,
+	approp.gross_outlay_amount_by_tas_cpe,
```
cmc333333: Why does this need to add a bunch of new fields in the `select` and `group by`? Generally, we shouldn't need to `group by` fields that aren't part of the `having` (and potentially `where`) clause. Is `approp.row_number` not unique within the table?

dataactvalidator/config/sqlrules/a19_cross_file.sql
```diff
@@ -1,11 +1,26 @@
...
-ON approp.tas = op.tas AND approp.submission_id = op.submission_id
+	approp.row_number,
+   approp.allocation_transfer_agency,
+	approp.agency_identifier,
+	approp.beginning_period_of_availa,
+	approp.ending_period_of_availabil,
+	approp.availability_type_code,
+	approp.main_account_code,
+	approp.sub_account_code,
+	approp.obligations_incurred_total_cpe,
```
cmc333333: Same note as above. Also, obviously nbd, but spacing on `allocation_transfer_agency` is off.

dataactvalidator/config/sqlrules/a18_cross_file.sql
```diff
@@ -1,10 +1,26 @@
 SELECT
-    approp.row_number,
-    approp.tas,
-    approp.gross_outlay_amount_by_tas_cpe,
-    SUM(op.gross_outlay_amount_by_pro_cpe) as gross_outlay_amount_by_pro_cpe_sum
-FROM appropriation approp JOIN object_class_program_activity op
-ON approp.tas = op.tas AND approp.submission_id = op.submission_id
+	approp.row_number,
```
bsweger: To answer @cmc333333's question about the addition of all the news in the `SELECT` and `GROUP BY` clauses....

This is essentially a bug fix. The new columns added to these clauses are the 7 fields that make up the TAS, the key that joins financial data files (e.g. appropriations, object_class_program_activity, and award_financial) together. 

For internal use, we created a column in each submission table (`tas`) that represents the combined 7 fields, We can use that column to make joins between the tables easier, but we can't use it for external display, which is why this query needed an update.

dataactvalidator/config/sqlrules/a19_cross_file.sql
```diff
@@ -1,11 +1,26 @@
...
+	SUM(op.obligations_incurred_by_pr_cpe) as obligations_incurred_by_pr_cpe_sum
```
cmc333333: Probably not an issue, but wanted to note it: the rules file says "negative sum"

dataactvalidator/config/sqlrules/a18_cross_file.sql
bsweger: The answer to the `select` and `group by` part of your question: https://github.com/fedspendingtransparency/data-act-broker-backend/pull/207#discussion_r74628071

The `group by` fields need to match the fields in the `select`. 

To your last question, `approp.row_number` is a bit misleading--it refers to the row number on the .csv submitted by the user, so it is unique within a submission but not within the entire table.

Good questions, all!

dataactvalidator/config/sqlrules/a19_cross_file.sql
bsweger: No, that's a good catch. That was a late-breaking change to this rule, and we updated the acceptance criteria to make the change as part of this story.

dataactvalidator/config/sqlrules/a30_appropriations.sql
```diff
@@ -0,0 +1,16 @@
...
+	approp.ending_period_of_availabil,
+	approp.availability_type_code,
+	approp.main_account_code,
+	approp.sub_account_code
+FROM appropriation AS approp
+WHERE approp.submission_id = {}
+	AND NOT EXISTS (
+		SELECT 1
+		FROM object_class_program_activity AS op
+		WHERE approp.tas IS NOT DISTINCT FROM op.tas
```
cmc333333: Small detail, but do we really want to allow null matches here? Put differently, if file A and B both have a null `tas`, _should_ this rule say they're ok?

If not, we can simplify to `approp.tas = os.tas`.

dataactvalidator/config/sqlrules/a32_appropriations.sql
```diff
@@ -0,0 +1,18 @@
...
+FROM appropriation AS approp
+WHERE approp.submission_id = {0}
+	AND EXISTS (
+		SELECT a.row_number,
+			a.tas
+		FROM appropriation AS a
+		WHERE a.row_number <> approp.row_number
+			AND a.tas IS NOT DISTINCT FROM approp.tas
+			AND a.submission_id = {0}
+	);
```
cmc333333: Note that this'll display _two_ errors when there's a duplicate. If that's not okay (maybe we only want to display the second error or something), we can swap out 
`a.row_number <> approp.row_number` for `a.row_number > approp.row_number` or similar.

Super minor nit, but `a` may not be a good identifier here. I've seen a lot of "self" comparison functions use `other` or `right`

dataactvalidator/config/sqlrules/a33_appropriations.sql
```diff
@@ -0,0 +1,21 @@
...
+		ON sf.period = sub.reporting_fiscal_period
+			AND sf.fiscal_year = sub.reporting_fiscal_year
+WHERE sub.submission_id = {0}
+	AND NOT EXISTS (
+		SELECT 1
+		FROM appropriation AS approp
+			JOIN submission AS sub
+				ON approp.submission_id = sub.submission_id
+		WHERE sf.tas IS NOT DISTINCT FROM approp.tas
+			AND sub.submission_id = {0}
```
cmc333333: Minor, but could this query be shorted to:
```sql
AND NOT EXISTS (
    SELECT 1
    FROM appropriation AS approp
    WHERE sf.tas IS NOT DISTINCT FROM approp.tas
    AND approp.submission_id = {0}   --- or just sub.submission_id
)
```
I'm not sure we gain anything by joining the submission table since we already know it exists?

dataactvalidator/config/sqlrules/b20_award_financial.sql
```diff
@@ -0,0 +1,20 @@
...
+	af.object_class
+FROM award_financial AS af
+WHERE af.submission_id = {}
+	AND NOT EXISTS (
+		SELECT 1
+		FROM object_class_program_activity AS op
+		WHERE af.tas IS DISTINCT FROM op.tas
+			AND af.program_activity_code IS DISTINCT FROM op.program_activity_code
+			AND af.object_class IS DISTINCT FROM op.object_class
+			AND af.submission_id = op.submission_id
```
cmc333333: Should this be:
```sql
WHERE af.submission_id = op.submission_id AND (
    af.tas IS DISTINCT FROM op.tas
    OR af.program_activity_code IS DISTINCT FROM op.program_activity_code
    OR af.object_class IS DISTINCT FROM op.object_class
)
```
? Otherwise _all three_ fields have to be different to trigger an error, right?

dataactvalidator/config/sqlrules/a30_appropriations.sql
bsweger: @cmc333333 Welcome to our vortex o' `NULL`s. You are now an official member of the team.

Lemme get your thoughts on this:
1. `tas` is required in both tables. If either or both has a blank `tas`, that will be reported by the required field checks that happen earlier in the validation process
2. we have a requirement not to report a pile of downstream errors that result from missing data. 
3. so I think Mike's idea here is to not produce an error record when either or both `tas`'s are `NULL` because we're already reporting that as missing data

bsweger: Does the query in the `NOT EXISTS` clause also need to filter on `submission_id`?

dataactvalidator/config/sqlrules/a32_appropriations.sql
bsweger: We had a lot of back and forth about the best way to surface errors for duplicate rows, since we don't have a way to know which row is the "right" one. I'm fine with reporting all rows involved, as long as we use that approach consistently for all rules related to dupes.

Agreed with @cmc333333 that a different alias for readability would be nifty.

dataactvalidator/config/sqlrules/a18_cross_file.sql
bsweger: Per @cmc333333's observation on A19, we also need to incorporate a change to A18 to handle some new math in the updated rule. I've Kristina for a clarification on this to make sure I'm grokking the rule text correctly.

bsweger:
As an additional check, I ran these SQL statements against some real data:

- A18...see Slack...had some additional questions about the new math for this query
- A19 - submission 416 on prod (with the [negative sum logic](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/207/commits/db4631b0464337b3a2925e2b0359861d60915ab2) added): got back two error rows that check out (they're off by a few cents, I suspect we'll get some feedback about this!) 👍 
- A30 - submission 416 on prod - Reported 3 A records with a TAS that is not B. Confirmed that those three TAS's are, in fact, missing. 👍 
- A32_appropriations: (tested this against prod submisison 245). it's one of our test files, so all TASs were duplicates, and the query returned all 10 rows. 👍 
- A33: can't run until schema changes in DB-712 are merged to dev
- B20: can't get this one to work (as written or with [CM's proposed change](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/207#discussion_r74633062)). Was testing using submission 242 on dev--set an objectclass in File C to 'zzz', and that didn't turn up in the query

dataactvalidator/config/sqlrules/a33_appropriations.sql
mab6bf: Yep, I think that should work

dataactvalidator/config/sqlrules/b20_award_financial.sql
mab6bf: Good catch

dataactvalidator/config/sqlrules/a18_cross_file.sql
mab6bf: Confirmed with Kristine this morning that the original implementation of A18 should stand, instead of the updated one with additional math that we received recently. The text will be updated in the next RSS, so we will be leaving A18 as-is for now.

dataactvalidator/config/sqlrules/a18_cross_file.sql
```diff
@@ -1,10 +1,26 @@
...
+GROUP BY approp.row_number,
+	approp.allocation_transfer_agency,
+	approp.agency_identifier,
+	approp.beginning_period_of_availa,
+	approp.ending_period_of_availabil,
+	approp.availability_type_code,
+	approp.main_account_code,
+	approp.sub_account_code,
+	approp.gross_outlay_amount_by_tas_cpe
 HAVING approp.gross_outlay_amount_by_tas_cpe <> SUM(op.gross_outlay_amount_by_pro_cpe)
```
cmc333333: Curious -- should this be 
```sql
 HAVING approp.gross_outlay_amount_by_tas_cpe <> gross_outlay_amount_by_pro_cpe_sum
```
They're equivalent, but one might describe the relationships that are being compared better.

dataactvalidator/config/sqlrules/a30_appropriations.sql
jworcestBAH: Yes, it definitely should include submission ID, we don't want to have matches to past submissions satisfy this rule.

bsweger: No. It does not.

bsweger: @jworcestBAH I hit submit w/o seeing your note, so wanted to clarify that my `No. It does not.` is not responding to you. In addition, it might also be a wrong comment, so there's that too.

jworcestBAH:
Tests passing with these rules being skipped in integration test

PR #209, opened 2016-08-15
Db 827 separation of integration and unit tests

nmonga91:


tests/integration/baseTestValidator.py
```diff
@@ -214,8 +214,7 @@ def uploadFile(cls, filename, user):
         bucketName = CONFIG_BROKER['aws_bucket']
         regionName = CONFIG_BROKER['aws_region']
 
-        path = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
-        fullPath = path + "/" + filename
+        fullPath = os.path.join(CONFIG_BROKER['path'], "tests", "integration", "data", filename)
```
cmc333333: Wow, yes this is much much cleaner.

tests/integration/fileTypeTests.py
```diff
@@ -101,11 +101,14 @@ def load_definitions(interfaces, force_tas_load):
         SchemaLoader.loadAllFromPath(join(CONFIG_BROKER["path"],"dataactvalidator","config"))
         SQLLoader.loadSql("sqlRules.csv")
         # Load domain values tables
-        loadDomainValues(join(CONFIG_BROKER["path"],"dataactvalidator","config"),join(CONFIG_BROKER["path"],"tests","sf_133.csv"),join(CONFIG_BROKER["path"],"tests","program_activity.csv"))
+        loadDomainValues(
+            join(CONFIG_BROKER["path"], "dataactvalidator", "config"),
+            join(CONFIG_BROKER["path"], os.path.join("tests", "integration", "data"), "sf_133.csv"),
+            join(CONFIG_BROKER["path"], os.path.join("tests", "integration", "data"), "program_activity.csv"))
```
cmc333333: So you need the separate `os.path.join`s here? I'd think the initial `join` would do the trick.

tests/unit/dataactcore/test_status_codes.py
```diff
@@ -0,0 +1,24 @@
...
+        assert StatusCode.CLIENT_ERROR == 400
+
+    def test_status_login_required(self):
+        assert StatusCode.LOGIN_REQUIRED == 401
+
+    def test_status_internal_error(self):
+        assert StatusCode.INTERNAL_ERROR == 500
+
+    def test_intentional_failure(self):
+        assert StatusCode.OK == 201
```
cmc333333: NBD, but we don't really need a class here; we certainly don't need the setup/teardown methods.

nmonga91: This is now resolved

tests/integration/fileTypeTests.py
nmonga91: I didn't change this portion, Becky had done this but it looks like first join handles the basePath parameter in loadDomainValues while the others lead to a specific test file that's in a completely different directory

klrBAH:
[run tests]

klrBAH:
[run tests]

cmc333333:
Excited to see it!

PR #211, opened 2016-08-15
Db 712 sf 133

nmonga91:


dataactcore/models/jobModels.py
```diff
@@ -4,6 +4,17 @@
 from sqlalchemy.orm import relationship
 from dataactcore.models.baseModel import Base
 
+
+def generateFiscalYear(context):
+    """ Generate fiscal year based on the date provided """
+    reporting_end_date = context.current_parameters['reporting_end_date']
+    return reporting_end_date.year
```
jworcestBAH: So this isn't quite enough, if the month is Oct-Dec, you need to add one to the year to bring it into the correct fiscal year.

dataactvalidator/config/sqlrules/a10_appropriations.sql
```diff
@@ -0,0 +1,13 @@
...
+    approp.borrowing_authority_amount_cpe,
+    SUM(sf.amount) as total_amount
+FROM appropriation as approp
+    INNER JOIN sf_133 as sf ON approp.tas = sf.tas
+    INNER JOIN submission as sub ON approp.submission_id = sub.submission_id AND
+        sf.period = sub.reporting_fiscal_period AND
+        sf.fiscal_year = sub.reporting_fiscal_year
+WHERE approp.submission_id = {} AND
+    sf.line in (1340, 1440)
+GROUP BY approp.row_number, approp.borrowing_authority_amount_cpe
```
jworcestBAH: I don't understand the purpose of grouping by the borrowing authority here, grouping just by row number should be sufficient shouldn't it?

nmonga91: when using GROUP BY, the column that need to be grouped on are all columns NOT in the aggregate function. So in this case, we're only summing on `sf.amount` so all other columns must be included in the GROUP BY statement

dataactcore/models/jobModels.py
nmonga91: This is updated and will be pushed once testing issues have been resolved.

dataactvalidator/config/sqlrules/b12_object_class_program_activity_1.sql
```diff
@@ -2,21 +2,21 @@ SELECT
...
+    OR COALESCE(op.ussgl498100_upward_adjustm_cpe, 0) <> 0
+    OR COALESCE(op.ussgl480200_undelivered_or_cpe, 0) <> 0
+    OR COALESCE(op.ussgl480200_undelivered_or_fyb, 0) <> 0
+    OR COALESCE(op.ussgl483200_undelivered_or_cpe, 0) <> 0
+    OR COALESCE(op.ussgl488200_upward_adjustm_cpe, 0) <> 0
+    OR COALESCE(op.ussgl490200_delivered_orde_cpe, 0) <> 0
+    OR COALESCE(op.ussgl490800_authority_outl_fyb, 0) <> 0
+    OR COALESCE(op.ussgl490800_authority_outl_cpe, 0) <> 0
+    OR COALESCE(op.ussgl498200_upward_adjustm_cpe, 0) <> 0)
+    AND COALESCE(op.by_direct_reimbursable_fun, '') = ''
```
jworcestBAH: So this rule is also supposed to check that the object class is three digits.  If object class is four digits, this flag is not required.

tests/unit/dataactvalidator/utils.py
```diff
@@ -12,15 +12,21 @@ def fetch_sql_tpl(filename):
...
 
 
-def models_are_valid(rule_file, staging_db, *models):
+def insert_submission(db, submission):
+    db.session.add(submission)
+    db.session.commit()
+    return submission.submission_id
+
+
+def run_sql_rule(rule_file, staging_db, submission_id, *models):
```
cmc333333: Separating out submission creation from `run_sql_rule` makes this a bit less self-contained. What do you think of this signature instead?

```python
def run_sql_rule(rule_file, staging_db, submission=None, models=None):
    if submission is None:
        submission = SubmissionCreationHere()
    if models is None:
        models = []
    submision_id = insert_submission(staging_db, submission)
    ...
```

tests/unit/dataactvalidator/test_b15_object_class_program_activity.py
```diff
@@ -0,0 +1,51 @@
...
+    assert run_sql_rule(_FILE, database.stagingDb, submission_id, sf, op) == 0
+
+
+def test_failure(database):
+    tas = "".join([_TAS, "_failure"])
+
+    submission_id = insert_submission(database.jobDb, Submission(user_id=1, reporting_fiscal_period=1,
+                                                                 reporting_fiscal_year=2016))
+
+    sf = SF133(line=2104, tas=tas, period=1, fiscal_year=2016, amount=1, agency_identifier="sys",
```
cmc333333: It looks like the key difference between these tests is the `amount`. Is there a way to write the tests so that's more apparent? Maybe something like:

```python
def test_amount_[something-descriptive](database):
    # all of test_success
    sf.period = 1
    assert run_sql_rule(...) == 1
```

tests/unit/dataactvalidator/utils.py
nmonga91: Agreed, I will put that in

dataactcore/migrations/versions/449ab366f333_adding_fiscal_year_and_period_to_submission.py
```diff
@@ -0,0 +1,42 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('submission', sa.Column('reporting_fiscal_period', sa.Integer(), server_default='0', nullable=False))
```
bsweger: Is `nullable=False` to handle existing data? Or is it a timing thing re: where the reporting period info gets added to a submission record?

nmonga91: I made it nullable because the reporting start/end dates are not nullable now, so it made sense to force these to be populated as well since it happens automatically. And we wouldn't want these cleared out on an update by accident either.

dataactcore/models/jobModels.py
```diff
@@ -35,8 +49,10 @@ class Submission(Base):
     datetime_utc = Column(DateTime)
     user_id = Column(Integer, nullable=False) # This refers to the users table in the User DB
     cgac_code = Column(Text)
-    reporting_start_date = Column(Date)
-    reporting_end_date = Column(Date)
+    reporting_start_date = Column(Date, nullable=False)
+    reporting_end_date = Column(Date, nullable=False)
+    reporting_fiscal_year = Column(Integer, nullable=False, default=generateFiscalYear, server_default='0')
```
bsweger: Admittedly I'm not fully conversant in sqlalchemy `default` vs `server_default` but is there a reason both wouldn't be set to your `generateFiscalYear` function?

dataactcore/models/jobModels.py
```diff
@@ -35,8 +49,10 @@ class Submission(Base):
...
+    reporting_fiscal_period = Column(Integer, nullable=False, default=generateFiscalPeriod, server_default='0')
```
bsweger: same question as above

dataactcore/models/jobModels.py
nmonga91: server_default can't take a function, it only takes a string value

dataactcore/migrations/versions/449ab366f333_adding_fiscal_year_and_period_to_submission.py
nmonga91: We could allow nulls, but since it's expected that reporting end date will always be there, didn't make much sense. Allowing nulls would simplify things and remove that server_default. Would we need to update historical data for this since it's just used in the validations?

dataactvalidator/config/sqlrules/a10_appropriations.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    approp.row_number,
+    approp.borrowing_authority_amount_cpe,
+    SUM(sf.amount) as total_amount
```
bsweger: @nmonga91 I don't think this was widely communicated, but we have an informal naming convention going for aggregate column aliases: `[something description]_[aggregate_being_applied].

So in this case, `SUM(sf.amount)` might be aliased as `sf_133_sum` or even `sf_133_1340_1440_sum`, though I see how that could get cumbersome in rules that involve a lot of SF 133 lines.

dataactvalidator/config/sqlrules/a11_appropriations.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    approp.row_number,
+    approp.spending_authority_from_of_cpe,
+    SUM(sf.amount) as total_amount
```
bsweger: Same comment as above re: aliases for aggregate columns. I think doing a global change of `total_amount` to `sf_133_sum` would go a long way towards consistency with other rules w/o being a heavy lift.

dataactvalidator/config/sqlrules/a12_appropriations.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    approp.row_number,
+    approp.adjustments_to_unobligated_cpe,
+    SUM(sf.amount) as total_amount
```
bsweger: same comment as above re: alias

dataactvalidator/config/sqlrules/a14_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    approp.row_number,
+    approp.gross_outlay_amount_by_tas_cpe,
+    sf.amount
```
bsweger: need column alias here (see above musings about that)

dataactvalidator/config/sqlrules/a15_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    approp.row_number,
+    approp.unobligated_balance_cpe,
+    sf.amount
```
bsweger: need column alias here (see above musings about that)

dataactvalidator/config/sqlrules/a22_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    approp.row_number,
+    approp.obligations_incurred_total_cpe,
+    sf.amount
```
bsweger: need column alias here (see above musings about that)

dataactvalidator/config/sqlrules/a23_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    approp.row_number,
+    approp.status_of_budgetary_resour_cpe,
+    sf.amount
```
bsweger: need column alias here (see above musings about that)

dataactvalidator/config/sqlrules/a8_appropriations.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    approp.row_number,
+    approp.budget_authority_appropria_cpe,
+    SUM(sf.amount) as total_amount
```
bsweger: see above comments re: column alias

dataactvalidator/config/sqlrules/b12_object_class_program_activity_1.sql
bsweger: No doubt you've thought of this already, @nmonga91 but would also add that we might also want to add the objectclass to the `SELECT` statement so it shows up on the error report.

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -19,7 +28,10 @@ A20,Must be valid Agency ID in the Common Government-wide Accounting Classificat
...
 A21,Indicator must be X or blank,Indicator must be X or blank,n,program_activity,fatal,,a21_object_class_program_activity
 A21,Indicator must be X or blank,Indicator must be X or blank,n,award_financial,fatal,,a21_award_financial
+A22,ObligationsIncurredTotalByTAS_CPE= CPE value for GTAS SF 133 line #2190,Failed rule: ObligationsIncurredTotalByTAS_CPE= CPE value for GTAS SF 133 line #2190,n,appropriations,fatal,,a22_appropriations
+A23,StatusOfBudgetaryResourcesTotal_CPE= CPE value for GTAS SF 133 line #2500,Failed rule: StatusOfBudgetaryResourcesTotal_CPE= CPE value for GTAS SF 133 line #2500,n,appropriations,fatal,,a23_appropriations
 A24,StatusOfBudgetaryResourcesTotal_CPE = BudgetAuthorityAvailableAmountTotal_CPE,StatusOfBudgetaryResourcesTotal_CPE = BudgetAuthorityAvailableAmountTotal_CPE,n,appropriations,fatal,,a24_appropriations
+A29,DeobligationsRecoveriesRefundsByTAS_CPE = CPE aggregate value for GTAS SF 133 line 1021+1033,Failed rule: DeobligationsRecoveriesRefundsByTAS_CPE = CPE aggregate value for GTAS SF 133 line 1021+1033,n,appropriations,fatal,,a29_appropriations
```
bsweger: I think A29 is a warning, no?

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -41,6 +53,8 @@ B12,"Reimbursable flag indicator is required when reporting obligation or outlay
 B12,"Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".","Reimbursable flag indicator is required when reporting obligation or outlay USSGL account balances (excluding downward adjustments SGL accounts). Valid values are ""R"" for ""ReimbursableFundingSource"" and ""D"" for ""DirectFundingSource"".",n,program_activity,fatal,,b12_object_class_program_activity_1
```
bsweger: On the latest version of validation_rules_for_broker, it looks like the status of B12 changed from fatal to warning

dataactcore/migrations/versions/449ab366f333_adding_fiscal_year_and_period_to_submission.py
bsweger: @nmonga91 Sorry--my brain was backwards when asking that question. Yes, completely agree that we should not allow nulls on these fields. Do we need to worry about existing submissions and making sure we backfill those with the correct fiscal period/year? What would happen if someone clicks on an old submission after we deploy this?

dataactcore/models/jobModels.py
bsweger: Thanks for the clarification! Then my question is...why do we need the `server_default` here? We would never want these fields to be set to 0, so having both options in the model was confusing when I read this.

We so something similar when loading the `tas` field on the staging tables, and there is no `server_default` on those models. If there's a good reason to have it, I'll make a note to go change that so we're consistent.

bsweger:
@nmonga91 Thanks for this--will be great to get these SF-133 rules in. I left some inline comments in the SQL and noted a few places in `sqlRules.csv` where the `severity_name` doesn't match the latest rules.

dataactcore/models/jobModels.py
nmonga91: the only reason it had 0 was because we are adding a non-nullable column which for legacy data would have nulls, so it sets the default to 0 to handle those potential errors. This there for a one time solution because the `reporting_end_date` that it needs should never be null, that's why I used server_default vs following what the `concatTas` does, but I can see why that route would be helpful but again, it'll only be need for this initial release of this column

dataactcore/migrations/versions/449ab366f333_adding_fiscal_year_and_period_to_submission.py
nmonga91: That field is only used for the validations currently so nothing should break if it's not backfilled. It just won't use them in validations because the year and period don't match.

dataactvalidator/config/sqlrules/b12_object_class_program_activity_1.sql
nmonga91: I've added that in :) it'll be going in with the next push

dataactvalidator/config/sqlrules/a10_appropriations.sql
nmonga91: This will be resolved in the next push

nmonga91:
@bsweger I've addressed almost all feedback. I will just need to update sqlRules.csv according to the latest in validation_rules_for_broker.csv. Thanks for the feedback!

tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,21 +1,24 @@
...
+    pass
+    #assert not run_sql_rule(_FILE, database.stagingDb, award)
 
 
 def test_nonnull_authority(database):
     award = Appropriation(job_id=1, row_number=1, is_first_quarter=True,
                           budget_authority_unobligat_fyb=5)
-    assert models_are_valid(_FILE, database.stagingDb, award)
+    pass
+    #assert run_sql_rule(_FILE, database.stagingDb, award)
```
cmc333333: Should these be `assert run_sql_rule(_FILE, database.stagingDb, objects=[award]) == 1` and similar?

nmonga91: @cmc333333 yes you're correct but they were failing the tests so I just forced them to pass since they weren't part of this story

bsweger:
@nmonga91 built and tested this on the sandbox, so we are ready to merge!

bsweger:
[run-tests]

PR #212, opened 2016-08-16
Db 860 unicode fix

jworcestBAH:
We now explicitly use utf-8 through the validation read and write and the job manager's copying of the file.  The problematic D2 file now completes validation (with errors on missing values).


dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -207,7 +207,6 @@ def _getLine(self):
             self.packetCounter +=1
 
             #Get the current lines
-            packet = packet.decode("utf-8") if not self.isLocal else packet
             currentBytes = self.unprocessed + packet
```
cmc333333: Sanity check: `currentBytes` is supposed to hold a unicode string, not a byte sequence, correct?

dataactcore/utils/jobQueue.py
```diff
@@ -51,16 +52,17 @@ def generate_d_file(api_url, file_name, user_id, d_file_id, interface_holder, ti
...
                 aws_file_name = "".join([str(user_id), "/", timestamped_name])
 
-                with open(timestamped_name, "wb") as file:
+                with open(timestamped_name, "w") as file:
                     # get request
                     response = requests.get(file_url)
                     # write to file
-                    file.write(response.content)
+                    response.encoding = "utf-8"
+                    file.write(str(response.text))
```
cmc333333: NBD, but `response.text` should be a `str` already, hey?

dataactvalidator/filestreaming/csvAbstractReader.py
jworcestBAH: Yes that variable is rather poorly named, it is intended to hold a string, unprocessed is the part of the last packet after the final line break, which is also a string.

dataactcore/utils/jobQueue.py
jworcestBAH: Yes, that's fair, this was originally response.content, which was not a string

cmc333333:
Looks good -- both `CsvS3Reader` and `CsvLocalReader` are reading and writing unicode content.

PR #213, opened 2016-08-16
Initial check-in of DB-570 without unit tests

cmc333333:
This is #207, but without the conflicts from new unittests. We can merge this independently.


cmc333333:
There's an implicit dependency on #211 already; we may just want to fix #207

cmc333333:
We're merging #211, so let's just fix #207 rather than splitting this out.

PR #214, opened 2016-08-17
temporarily bypass the SF-133 load

bsweger:
Since we're temporarily bulk loading the historical SF-133 data, we'll also need to temporarily bypass the current SF-133 loader, which will delete everything in that table whenever we deploy.

The bulk load is a temporary measure to ensure that we have a set of FY 2016 SF-133 data for use by the GTAS validations. The [completed DB-862 story](https://federal-spending-transparency.atlassian.net/browse/DB-862) will restore the SF-133 load.

### DIFF
--- a/dataactvalidator/filestreaming/loadFile.py
+++ b/dataactvalidator/filestreaming/loadFile.py
@@ -28,20 +28,25 @@ def loadDomainValues(basePath, localSFPath = None, localProgramActivity = None):
     print("Loading object class")
     loadObjectClass(os.path.join(basePath,"object_class.csv"))
 
-    # SF 133 is kept on S3, so need to download that
-    reader = CsvS3Reader()
+    # Since we're temporarily bulk loading the historical SF-133 data,
+    # we'll also need to temporarily bypass the current SF-133 loader,
+    # which will delete everything in that table whenever we deploy.
 
-    if localSFPath is not None:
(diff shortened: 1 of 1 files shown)


dataactvalidator/filestreaming/loadFile.py
```diff
@@ -28,20 +28,25 @@ def loadDomainValues(basePath, localSFPath = None, localProgramActivity = None):
...
     loadObjectClass(os.path.join(basePath,"object_class.csv"))
 
-    # SF 133 is kept on S3, so need to download that
-    reader = CsvS3Reader()
+    # Since we're temporarily bulk loading the historical SF-133 data,
+    # we'll also need to temporarily bypass the current SF-133 loader,
+    # which will delete everything in that table whenever we deploy.
+
+    # # SF 133 is kept on S3, so need to download that
+    # reader = CsvS3Reader()
```
nmonga91: I know this is commented and not used right now, but the reader should be initialized in the `else` because at that point it is confirmed that it's not coming from a local path but instead from S3

nmonga91:
Looks good, merging

dataactvalidator/filestreaming/loadFile.py
bsweger: Good call @nmonga91--will move that when we re-instate the SF-133 load.

PR #216, opened 2016-08-19
Db 712 update integration tests

jworcestBAH:
Limiting mixed tests to subset of rules, adding new rules will only require updating valid sample files.


tests/integration/fileTypeTests.py
```diff
@@ -29,57 +30,25 @@ def setUpClass(cls):
...
-            "crossApprop": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossApprop, 1],
-            "crossPgmAct": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossPgmAct, 2],
-            "crossAwardFin": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossAwardFin, 3],
-            "crossAward": [statusReady, jobTypeCsv, str(submissionIDs[11]), s3FileNameCrossAward, 4],
-            "crossFile": [statusReady, jobTypeValidation, str(submissionIDs[11]), None, None],
-            "appropValidShortcols": [statusReady, jobTypeCsv, str(submissionIDs[12]), s3FileNameAppropValidShortcols, 1],
-            "programMixedShortcols": [statusReady, jobTypeCsv, str(submissionIDs[13]), s3FileNameProgramMixedShortcols, 2],
-            "awardFinMixedShortcols": [statusReady, jobTypeCsv, str(submissionIDs[14]), s3FileNameAwardFinMixedShortcols, 3],
-            "awardValidShortcols": [statusReady, jobTypeCsv, str(submissionIDs[15]), s3FileNameAwardValidShortcols, 4]
+            "awardValid": [statusReady, jobTypeCsv, str(submissionIDs[8]), s3FileNameAwardValid, 4]
```
bsweger: I know this is legacy code, but as long we're cleaning up our tests, it would be great to replace the file type "magic numbers" in the `jobInfoList` dict with a lookup from the `file_type` table.

tests/integration/fileTypeTests.py
```diff
@@ -97,19 +66,18 @@ def setUpClass(cls):
...
-        # rules
-        to_delete = interfaces.validationDb.session.query(RuleSql).filter(
-            RuleSql.rule_label.in_(('A30', 'A32', 'A33', 'B20', 'B19', 'C19')))
-        for rule in to_delete:
-            interfaces.validationDb.session.delete(rule)
-        interfaces.validationDb.session.commit()
+        if ruleList is not None:
+            # If rule list provided, drop all other rules
+            to_delete = interfaces.validationDb.session.query(RuleSql).filter(not_(
+                RuleSql.rule_label.in_(ruleList)))
```
bsweger: Nice--thanks for putting the list somewhere more universal! I like the way this is implemented as strictly a test thing w/o adding another param to the application's sqlloader.

tests/integration/fileTypeTests.py
```diff
@@ -29,57 +30,25 @@ def setUpClass(cls):
...
 
         # Create submissions and get IDs back
         submissionIDs = {}
-        for i in range(0, 16):
-            if i == 7:
-               # Award financial mixed will be second quarter
-                submissionIDs[i] = cls.insertSubmission(cls.jobTracker, user, datetime(2015,3,15))
-            else:
-                submissionIDs[i] = cls.insertSubmission(cls.jobTracker, user)
+        for i in range(0, 9):
```
bsweger: Do we need 9 submissions?

bsweger:
@jworcestBAH done reviewing--just a few questions above. I'm willing to merge w/o fixing the magic file_type numbers, since they are legacy and we still have some test cleanup to do.

Do you imagine that if we ever backfill our unit tests with things other than validation rules, that checks like whitespace, etc. would be moved from these tests to those?

tests/integration/fileTypeTests.py
jworcestBAH: So the reason the submission numbers are done this way is we don't want to have conflicting submission IDs between this and mixedFileTests.  They're in separate DBs, but the reports get written to the same place, so those would be overwritten if they used the same submission ID.  I could rearrange the numbers so that this is 1-4 and mixed is 5-15.

tests/integration/fileTypeTests.py
jworcestBAH: Fixed

jworcestBAH:
If we had smaller unit tests for things like reading whitespace and special characters from csv, yes.  Definitely the bad job id, job id for wrong type, etc. could be converted to smaller unit tests.

bsweger:
Thanks for the magic number update and the submission id clarification. Looks good--merging!

bsweger:
Note: still working through a bug in our local broker installs, so we will forgo that check until it's fixed. Still think we're good to merge, since the integration tests should ensure the test files will pass.

PR #217, opened 2016-08-19
Updates to submissions do not recreate D files

jworcestBAH:
Fixes the bug of hanging after updates on cross-file

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -164,18 +164,20 @@ class FileHandler:
                 raise ResponseException("Must include at least one file for an existing submission",
                                         StatusCode.CLIENT_ERROR)
 
-            for extFileType in FileHandler.EXTERNAL_FILE_TYPES:
-                if extFileType == "award":
-                    filename = CONFIG_BROKER["d2_file_name"]
-                elif extFileType == "award_procurement":
-                    filename = CONFIG_BROKER["d1_file_name"]
+            if not existingSubmission:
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -164,18 +164,20 @@ def submit(self,name,CreateCredentials):
...
-                    filename = CONFIG_BROKER["d1_file_name"]
-
-                if(not self.isLocal):
-                    uploadName = str(name) + "/" + s3UrlHandler.getTimestampedFilename(filename)
-                else:
-                    uploadName = filename
-                responseDict[extFileType + "_key"] = uploadName
-                fileNameMap.append((extFileType, uploadName, filename))
+            if not existingSubmission:
+                # Existing submissions do not recreate D jobs
```
bsweger: Thanks for the comment!

bsweger:
@jworcestBAH Done reviewing. Looks good--just have one question about existing submissions. If a user needs to modify a submission and has to regenerate the D files as a result, is that something we currently handle, or is that future story? If the former, does this new logic accommodate that scnenario?

jworcestBAH:
Regeneration of D files works by calling the generate_d2_file and check_d2_file routes, that process is unaffected by this change.  By not resetting the statuses here, the user still has the option of regenerating the D2 file on the cross-file page, but we do not require the d2 file to be regenerated whenever a submission is changed (which the frontend does not support anyway).

bsweger:
Thanks for the context! Looks good..merging!

PR #218, opened 2016-08-22
Moved completion on File table to be marked at same time as completio…

jworcestBAH:
…n on Job table

### DIFF
--- a/dataactvalidator/validation_handlers/validationManager.py
+++ b/dataactvalidator/validation_handlers/validationManager.py
@@ -117,7 +117,6 @@ class ValidationManager:
                 raise ResponseException("Bad job type for validator",
                                         StatusCode.INTERNAL_ERROR)
             self.runValidation(jobId, interfaces)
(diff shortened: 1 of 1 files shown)


bsweger:
@jworcestBAH This looks good--will merge.

It does raise the question, though--now that we've consolidated databases, does it make sense to consider keeping the info about file/job status in a single place so we don't have to worry about things getting out of sync?

jworcestBAH:
I've wondered about that as well, currently the File table holds a list of headers missing or duplicated as well as the row_errors_present flag that don't necessarily make sense for upload jobs, file generation jobs, or the cross-file job.  The status list for this table also distinguishes various types of file-level validation errors that are not included in the JobStatus table.  So it may make sense to merge these two, but it will take some work to preserve the validation-specific information in a way that makes sense.

bsweger:
BTW, was reminded by our error status issue yesterday to log this as a post-db cleanup task: https://federal-spending-transparency.atlassian.net/browse/DB-1036

PR #219, opened 2016-08-22
First attempt at loading FSRS data

cmc333333:
Looks like a lot of code but most of the lines are lists of field names. That said, it wouldn't be too hard to split this PR up based on the commit history if that'd be easier to review.

This adds models associated with the FSRS web service for agency award/subaward data. As we'll be pulling in this data from them on a routine basis, the models mirror their structure almost verbatim. We can trim or transform some of these fields if we find them not useful in the future.

To accomplish said export and aid related unittests, this also adds a fork of the `suds` library and `factory_boy` to our list of requirements. The former allows for relatively painless communications with a SOAP service while the latter generates fake data for tests relatively easily.

Note that we can move code around if there are better places for all of this logic to live.


requirements.txt
```diff
@@ -19,4 +19,5 @@ celery==3.1.23
 SQLAlchemy-Utils==0.32.1
 xmlrunner==1.7.7
 pytest==2.9.2
-pytest-cov==2.3.1
\ No newline at end of file
+pytest-cov==2.3.1
+suds-jurko==0.6
```
bsweger: 😂

tests/unit/dataactvalidator/test_fsrs.py
```diff
@@ -0,0 +1,56 @@
...
+
+def newClient_call_args(monkeypatch, **config):
+    """Set up and create a newClient request with the provided config. Return
+    the arguments which were sent to the suds client"""
+    mock_client = Mock()
+    monkeypatch.setattr(fsrs, 'Client', mock_client)
+    config = {'fsrs_service': config}
+    monkeypatch.setattr(fsrs, 'CONFIG_BROKER', config)
+    fsrs.newClient()
+    return mock_client.call_args[1]
```
bsweger: Can you talk a bit about what `[1]` is doing here? I looked up the Mock `call_args` function in the docs and it says that it takes a tuple or None. Is this some more pytest magic at work?

cmc333333: Sure thing -- I'll update it to something like the following if it's cleared?

```python
args, kwargs = mock_client.call_args 
```

We only care about the `kwargs`, but I haven't found a good way to retrieve that by name from a `unittest.mock.call` other than decomposing the tuple.

dataactvalidator/fsrs.py
```diff
@@ -0,0 +1,33 @@
...
+
+    # The WSDL is missing an import; it's so common that suds has a work around
+    parsedWsdl = urlparse(wsdlUrl)
+    importFix = doctor.Import('http://schemas.xmlsoap.org/soap/encoding/')
+    importFix.filter.add(   # Main namespace is the wsdl domain
+        '{}://{}/'.format(parsedWsdl.scheme, parsedWsdl.netloc))
+
+    options['doctor'] = doctor.ImportDoctor(importFix)
+
+    if serviceConfig.get('username') and serviceConfig.get('password'):
```
bsweger: What will happen downstream if the config file is missing this username/password? Will we need to check for this in `dataactcore/config.py` and provide some kind of warning?

tests/unit/dataactvalidator/test_fsrs.py
bsweger: Aaaah--yes, I see now. I was so intent on reading the letter of the docs that I completely missed what you were doing, so maybe your proposal would be clearer. Thanks!

dataactvalidator/fsrs.py
cmc333333: I tried to keep the connection-dependent code relatively tight, but the import script will error out with a connection failure if not configured correctly. I expected that the majority of developers wouldn't be running those import scripts and hence wouldn't need the connection information; if that's a safe assumption, maybe the warning should be in the import script rather than `config.py`?

requirements.txt
```diff
@@ -20,4 +20,6 @@ SQLAlchemy-Utils==0.32.1
 xmlrunner==1.7.7
 pytest==2.9.2
 pytest-cov==2.3.1
-suds-jurko==0.6
+# suds is outdated but more compatible with FSRS than other libs. Pin a
+# version here
+-e git+https://github.com/andersinno/suds-jurko.git@44e27560b328ec6482459ea2d75516849dc209ff#egg=suds
```
bsweger: Thanks for the explanation in the commit message. This change would've definitely been a head-scratcher otherwise.

dataactvalidator/scripts/loadFSRS.py
```diff
@@ -0,0 +1,32 @@
...
+from dataactcore.models.baseInterface import databaseSession
+from dataactcore.models.fsrs import FSRSAward
+from dataactvalidator.fsrs import retrieveAwardsBatch
+
+def loadFSRS(minId):
+    """Fetch and store a batch of Award/Sub-Award records, replacing previous
+    entries."""
+    awards = list(retrieveAwardsBatch(minId))
+    ids = [a.id for a in awards]
+    with databaseSession() as sess:
```
bsweger: Just to make sure I understand this correctly...

The `databaseSession()` you added to `baseInterface.py` provides a way to do a few things in a discrete database session using `WITH...AS`, and this is an alternative to how we've been doing this to date (i.e., creating a new whateverInterface())?

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
```diff
@@ -0,0 +1,73 @@
...
+
+import pytest
+
+from dataactcore.models.fsrs import FSRSAward, FSRSSubaward
+from dataactvalidator.scripts import loadFSRS
+from tests.unit.dataactcore.factories import (
+    FSRSAwardFactory, FSRSSubawardFactory)
+
+
+@pytest.yield_fixture()
```
bsweger: Been reading up on pytest fixtures in the docs, and it looks like the `yield_fixture` decorator is deprecated (_i.e._, you don't need it to use `yield` for teardown).

To be clear, I'm reading these docs not to check up on you, but to understand what is happening.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
```diff
@@ -0,0 +1,73 @@
...
+def no_award_db(database):
```
bsweger: At the risk of outing myself for missing something obvious, I can't wrap my head around what `database` is and how it's working. I get that this stands in for a db in the unit tests, but was unable to find anything helpful in the pytest fixtures docs. Do you know of a good resource for grokking this?

dataactvalidator/scripts/loadFSRS.py
cmc333333: Ah yes, thanks for pointing this out. I found the database session code is rather complex, with several "interfaces", shared static fields, and connection closing on object deletion. Rather than than trying to figure that out, I went with a context manager for simplicity. I can swap it out for the existing approach -- I just didn't (still don't, entirely) know what that approach was.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
cmc333333: Oh neat, I'll swap those out.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
cmc333333: Sorry @bsweger! The database `fixture` is defined in `tests/unit/conftest.py` as a quick way to set up a database for the unit tests. As noted above, I'm not comfortable with the existing database connection approach, so I've tried to abstract it all away into some sort of "connection". Unfortunately, the `database` fixture is yielding an "InterfaceHolder" while `databaseSession` is yielding a Sqlalchemy `Session`; I'd like to be consistent, but I'm also not sure which approach is a better match for our work. What do you think?

dataactvalidator/scripts/loadFSRS.py
bsweger: I really like the simplicity of what you're doing here, and I vote for keeping it this way. Paging @jworcestBAH who can better speak to the original intent of the interfaces and make sure we're not missing anything here.

bsweger: Eh, gonna go ahead and page @nmonga91 too so he can see the context manager at work and provide his .02.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
bsweger: No need to apologize! I was confused because I'm still learning how pytest passes around fixtures, and wasn't sure where `database` was coming from--appreciate the clarification on that.

IMO, the simplest possible approach is always the best match for our work. Our InterfaceHolder was created back when we had 5 databases and 3 code repos for the broker's backend. I'm very amenable to revisiting how best to interact with the database now that we've streamlined a bit.

For now, I'm good just knowing what this `database` is and where it came from 😄

bsweger: That said, I think it's important to use a real database when unit testing our validation rules. Outside of validation rules, though, do you think we should be mocking a database for the unit tests?

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
```diff
@@ -0,0 +1,73 @@
...
+    assert fetch_duns(2) == 'To Be Replaced'
+    assert no_award_db.query(FSRSSubaward).count() == 5
+
+    award3 = FSRSAwardFactory(id=2, duns='Replacing')
+    monkeypatch.setattr(loadFSRS, 'retrieveAwardsBatch',
+                        Mock(return_value=[award3]))
+    loadFSRS.loadFSRS(0)
+    assert fetch_duns(1) == 'Not Altered'
+    assert fetch_duns(2) == 'Replacing'
+    assert no_award_db.query(FSRSSubaward).count() == 4
```
bsweger: If I read this correctly, the subaward count drops from 5 to 4 because the subaward for award 2 was the victim of a cascading FK delete when award 2 was updated.  Which makes me wonder--is it even legal to have an award without the subaward component?

In the overall context of federal awards, it definitely is. This question is related to a larger question about the way the FSRS models are set up, which I'll ask on the models commit, where it will make more sense.

dataactcore/models/fsrs.py
```diff
@@ -0,0 +1,81 @@
...
+    contracting_office_id = Column(String)
+    contracting_office_name = Column(String)
+    treasury_symbol = Column(String)
+    dollar_obligated = Column(String)
+    date_signed = Column(Date)
+    transaction_type = Column(String)
+    program_title = Column(String)
+
+
+class FSRSSubaward(Base, CommonAttributes):
```
bsweger: Can you say more about the what the `FSRSAward` and `FSRSSubaward` models represent? By definition, everything in FSRS is related to a subaward, yes? That said, I have not closely read the FSRS technical docs.

I do see separate FSRS docs for financial assistance awards and contract awards--is that what these models represent?

dataactvalidator/fsrs.py
bsweger: Yeah, you're right. We're getting down into functionality that isn't relevant to local developers. As long as we can be nice and provide a warning somewhere, I am good.

bsweger:
@cmc333333 done reviewing! Lots to think about here, and some great tools introduced to the code base. Thank you!

The question of how best to work with our databases is one that I think we should think through as a team. Thanks for provide a concrete example to look at--I really like the simplicity.  I'm good to keep this as is and merge once we work through the comments and then follow-up by getting a db convo on our calendars.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
cmc333333: > do you think we should be mocking a database for the unit tests?

That's a good question; it's very tempting to use the existing test database since we'll use it in the validation tests, but that doesn't mean it's a good idea. For example, if I want to run just the tests around these models, I need to wait an extra few seconds for the database to be set up and tore down. Eh, I don't think the benefits outweigh the costs here -- I'm okay sticking with postgres unless someone objects.

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
cmc333333: You are reading that correctly (though I should comment it better). The WSDL (i.e. XML schema) allows for awards with no subawards, but I don't have a sense if that's ever used. I'm content to modify the test to be more realistic.

dataactcore/models/baseInterface.py
```diff
@@ -157,3 +159,10 @@ def getNameFromDict(self, model, dictName, fieldName, fieldValue, idField):
                 return key
         # If not found, raise an exception
         raise ValueError("Value: " + str(fieldValue) + " not found in dict: " + str(dict))
+
+
+@contextmanager
+def databaseSession():
+    interface = BaseInterface()
+    yield interface.session
+    interface.close()
```
jworcestBAH: We would need to redesign the interfaces to accommodate this usage.  By closing the connection after each use, any existing interfaces will hit errors because they share a connection.

dataactvalidator/scripts/loadFSRS.py
jworcestBAH: There is also a getSession() function in BaseInterface that returns the session object.  This seems equivalent, except for the caution above about closing the shared connection after each use.  There is also a small possibility that doing things this way will create a new engine, connection, and session for each usage, I'm not sure that we've really tested the constructor for this case.  If you intended this to create a separate engine, connection, and session, that would require some refactoring of the static variables involved.

dataactcore/models/fsrs.py
cmc333333: Yessum. I'm not super familiar with the data, but my understanding so far is that FSRS data connects a "prime" awardee with subawards and subcontracts. The XML schema around this indicates that there's a 1-to-many relationship between awards ("FFATAExport") and subawards/subcontracts ("SubcontractorExport"), though it does allow for zero subawards/subcontracts. More mechanistically, the XML result is a list of FFATAExports, each of which may contain an array of SubcontractorExports. Aside from the class names, the models here map almost verbatim (they flatten out nested addresses and top-paid data) to the XML schema, which is probably more conservative than we need. I agree that we should think about how to filter and transform this data, but I'd prefer to defer that until knowing more about the problem space.

Should I try to summarize that in the docstrings?

dataactcore/models/baseInterface.py
cmc333333: Gotcha, so we'd see something like:
```python
interface = BaseInterface()
interface.connection    # not null
with databaseSession() as sess:
    doSomeThingWith(sess)
interface.connection    # now null
```

What do you think about skipping that shared connection altogether? Say we separate out the engine/connection/session creation into a separate function:
```python
class BaseInterface:
    def __init__(self):
        self.engine, self.connection, self.session = newDatabaseContext()
        ...
```
Then we can choose not to use those static variables:
```python
def databaseSession():
    engine, connection, session = newDatabaseContext()
    yield session
    engine.dispose()
    connection.close()
    session.close()
```
This'd recreate and close the database connection with each `databaseSession` call, but we'd explicitly know when the connections were valid (i.e. through the end of the context). This might be easier to reason about than the static vars + garbage collection approach.

jworcestBAH: Very nice, I like this approach.  For the call from the constructor, just make sure it happens only if self.session is None, otherwise we'll have a repeat of last week's issue with connections staying open.

dataactcore/models/fsrs.py
bsweger: I think the disconnect for me is the term _subawards_. There are two types of awards:

1. financial assistance awards (grants, loans, insurance, etc.)
2. contracts

Each type of award can have 0 or more subawards, and those subawards will have different data depending on the type of prime award they are connected to. It sounds like FSRS is using the term _subaward_ to refer to sub financial assistance awards. What do you think about being more explicit/accurate than that in our broker models and using the more nuanced term _subassistance_ or something like that?

_Subaward_ is confusing because they're all subawards, if that makes sense.

Lemme know if this is turning into a convo best had synchronously!

tests/unit/dataactvalidator/test_scripts_loadFSRS.py
bsweger: I think my disconnect is still around award data coming from FSRS and the award data we already have in the broker, which is sourced from more primary sources like the Federal Procurement Data System (FPDS) for contracts and the Award Submission Portal (ASP) for financial assistance awards.

Can we check in synchronously or via Slack to sort this?

cmc333333:
Very late but @bsweger this is ready for review whenever. I basically duplicated all of the models and named them `Procurement/Subcontract` and `Grant/Subgrant` to distinguish. As we discussed before, this is just trying to synchronize with the FSRS models, not map into our needs just yet.

dataactvalidator/scripts/loadFSRS.py
```diff
@@ -28,6 +29,12 @@ def maxCurrentId():
 
 
 if __name__ == '__main__':
-    num_awards, num_subawards = loadFSRS(maxCurrentId() + 1)
-    print("Inserted/Updated {} awards, {} subawards".format(
-        num_awards, num_subawards))
+    if not configValid():
+        print("No config key for broker/fsrs_service/wsdl")
```
bsweger: Thanks for adding a check here! Is this a place where we'd want to use your suggested logging method?

That shouldn't keep us from proceeding--more just wondering if the general pratice is to do logging wherever you're displaying this type of info or if it's more nuanced than that.

dataactcore/models/baseInterface.py
```diff
@@ -161,8 +144,32 @@ def getNameFromDict(self, model, dictName, fieldName, fieldValue, idField):
         raise ValueError("Value: " + str(fieldValue) + " not found in dict: " + str(dict))
 
 
+def dbConnection():
+    """Use the config to set up a database engine and connection"""
+    if not CONFIG_DB:
+        raise ValueError("Database configuration is not defined")
+
+    dbName = CONFIG_DB['db_name']
```
bsweger: Will this throw a key error if `db_name` isn't in the `CONFIG_DB` dict?

dataactcore/models/baseInterface.py
```diff
@@ -161,8 +144,32 @@ def getNameFromDict(self, model, dictName, fieldName, fieldValue, idField):
...
 @contextmanager
 def databaseSession():
-    interface = BaseInterface()
-    yield interface.session
-    interface.close()
+    engine, connection = dbConnection()
+    sessionMaker = scoped_session(sessionmaker(
+        bind=engine, autoflush=True))
+    session = sessionMaker()
+    yield session
```
bsweger: Checking myself on how this works...so anything after `yield session` automatically gets done after the code in the invoking (for lack of a better word) `with` gets executed?

dataactvalidator/scripts/loadFSRS.py
cmc333333: I agree -- I ended up adding a logger in a later revision.

dataactcore/models/baseInterface.py
cmc333333: It will indeed, matching the current behavior. Do you think we should raise a `ValueError` instead?

dataactcore/models/fsrs.py
```diff
@@ -65,15 +65,22 @@ class GrantAttributes(FSRSAttributes):
...
-class PrimeAwardAttributes:
+class _PrimeAwardAttributes:
     """Attributes shared by FSRSProcurements and FSRSGrants"""
     internal_id = Column(String)
     date_submitted = Column(DateTime)
     report_period_mon = Column(String)
     report_period_year = Column(String)
 
+    @classmethod
+    def nextId(cls, sess):
```
bsweger: Good idea!

dataactcore/models/baseInterface.py
cmc333333: You got it. You can think of everything within the `with` expression as being executed in place of `yield session`. It's very much like a generator in the sense that the function is "paused".

dataactcore/models/baseInterface.py
bsweger: Ah, I see, this handles the case where `dbName` is in the config file but doesn't have a value. I'm with you now..yeah, we can leave this, though I'm generally in favor of explicitly checking for things that could go awry in a user-configured file.

If I was doing this again,does it make sense to do a `.get` so people get a

dataactcore/models/baseInterface.py
bsweger: This is so cool. Thank you for simplifying the db connection!

dataactvalidator/fsrs.py
```diff
@@ -7,19 +7,34 @@
 from suds.xsd import doctor
 
 from dataactcore.config import CONFIG_BROKER
-from dataactcore.models.fsrs import FSRSAward, FSRSSubaward
+from dataactcore.models.fsrs import (
```
bsweger: Did an earlier commit have some warning messages if there was a problem with the FSRS service? Did those move somewhere else?

dataactvalidator/scripts/loadFSRS.py
```diff
@@ -1,40 +1,24 @@
...
-        try:
-            num_awards, num_subawards = loadFSRS(maxCurrentId() + 1)
-            print("Inserted/Updated {} awards, {} subawards".format(
-                num_awards, num_subawards))
-        except TransportError as exc:
-            print("Error communicating with FSRS: {}".format(exc))
+    logging.basicConfig(level=logging.INFO)
+    with databaseSession() as sess:
+        if not configValid():
+            logger.error("No config for broker/fsrs/[service]/wsdl")
```
bsweger: Disregard previous question about the error message!

bsweger:
This looks great @cmc333333 -- thank your accepting the mission of subwards and for making some improvements to the overall code base along the way. Merging!

PR #220, opened 2016-08-23
Unit tests for B19 and C19; updated a typo in sqlRules

mab6bf:


tests/unit/dataactvalidator/test_b19_object_class_program_activity.py
```diff
@@ -0,0 +1,98 @@
...
+                                     allocation_transfer_agency='1', availability_type_code='1',
+                                     main_account_code='1', sub_account_code='1', object_class='1',
+                                     program_activity_code='2', by_direct_reimbursable_fun='r')
+
+    op11 = ObjectClassProgramActivity(job_id=1, row_number=1, beginning_period_of_availa='1',
+                                     ending_period_of_availabil='1', agency_identifier='1',
+                                     allocation_transfer_agency='1', availability_type_code='1',
+                                     main_account_code='1', sub_account_code='1', object_class='1',
+                                     program_activity_code='1', by_direct_reimbursable_fun='d')
+
```
jworcestBAH: Since most of these fields are optional, it would probably be good to have some test cases where some of the values are missing.

tests/unit/dataactvalidator/test_c19_award_financial.py
```diff
@@ -0,0 +1,105 @@
...
+    af14 = AwardFinancial(job_id=1, row_number=1, beginning_period_of_availa='1', ending_period_of_availabil='1',
+                         agency_identifier='1', allocation_transfer_agency='1', availability_type_code='1',
+                         main_account_code='1', sub_account_code='1', object_class='1', program_activity_code='1',
+                         parent_award_id='1', piid='1', uri='1', fain='2', transaction_obligated_amou='1')
+
+    af15 = AwardFinancial(job_id=1, row_number=1, beginning_period_of_availa='1', ending_period_of_availabil='1',
+                          agency_identifier='1', allocation_transfer_agency='1', availability_type_code='1',
+                          main_account_code='1', sub_account_code='1', object_class='1', program_activity_code='1',
+                          parent_award_id='1', piid='1', uri='1', fain='1', transaction_obligated_amou='2')
+
```
jworcestBAH: Similar to above, in this case the typical usage is that only one of (piid, fain, uri) will be populated, so it would be good to have a test that covered that case.

tests/unit/dataactvalidator/test_b19_object_class_program_activity.py
```diff
@@ -0,0 +1,107 @@
...
+                                     allocation_transfer_agency='1', availability_type_code='1',
+                                     main_account_code='1', sub_account_code='1', object_class='1',
+                                     program_activity_code='1', by_direct_reimbursable_fun='r')
+
+    op2 = ObjectClassProgramActivity(job_id=1, row_number=1, beginning_period_of_availa='1',
+                                     ending_period_of_availabil='1', agency_identifier='1',
+                                     allocation_transfer_agency='1', availability_type_code='1',
+                                     main_account_code='1', sub_account_code='1', object_class='1',
+                                     program_activity_code='1', by_direct_reimbursable_fun='r')
+
```
jworcestBAH: To make sure the query is functioning as intended, we should have this check leave out one field, e.g. by setting allocation_transfer_agency = None for both.

jworcestBAH:
Since C19 no longer exists, we should drop the test for it, should we also drop the rule in this PR or do that in a separate one?

jworcestBAH:
After looking at the rule definition again, it uses the tas field, which should handle the case I was concerned about with missing values in the "successful" case, so I'm merging now.

PR #222, opened 2016-08-25
Db 843 file f routes

jworcestBAH:
Created consolidated routes and functions for files D1, D2, E, and F.  This PR does not remove the existing routes, so it does not require coordination with frontend deploy.  A later PR will remove the D-specific functionality.


.gitignore
```diff
@@ -38,3 +38,4 @@ shared-local-instance.db
 dataactcore/scripts/dropMany.py
 runDynamo.py
 dataactvalidator/config/sf_133.csv
+runTests.py
```
cmc333333: We don't want to check in changes to runTests.py?

dataactbroker/handlers/fileHandler.py
```diff
@@ -163,14 +164,17 @@ def submit(self,name,CreateCredentials):
...
+                # Don't add external files to existing submission
                 for extFileType in FileHandler.EXTERNAL_FILE_TYPES:
                     if extFileType == "award":
                         filename = CONFIG_BROKER["d2_file_name"]
                     elif extFileType == "award_procurement":
                         filename = CONFIG_BROKER["d1_file_name"]
+                    elif extFileType == "awardee_attributes":
+                        filename = CONFIG_BROKER["awardee_file_name"]
+                    elif extFileType == "sub_award":
+                        filename = CONFIG_BROKER["sub_award_file_name"]
```
cmc333333: Not a big deal, but now that there are four branches, would it make sense to move this into a mapping? Something like:
```python
FILE_TYPE_TO_NAME = {
    "award": CONFIG_BROKER["d2_file_name"],
    "award_procurement": CONFIG_BROKER["d1_file_name"],
    ...
}
...
filename = FILE_TYPE_TO_NAME[extFileType]
```

It'd also be a good idea to standardize these names (if possible) so that we could use, e.g.
```python
filename = CONFIG_BROKER[extFileType + "_file_name"]
```

dataactbroker/handlers/fileHandler.py
```diff
@@ -297,8 +301,16 @@ def getStatus(self):
             for jobId in jobs:
                 jobInfo = {}
                 jobType = self.jobManager.getJobType(jobId)
+
                 if jobType != "csv_record_validation" and jobType != "validation":
                     continue
+
+                # TODO Skip D1 file until validation is added, remove these lines once D1 validation is added
```
cmc333333: Is there a way to make this more prominent? Maybe make a separate issue if one doesn't already exist? I don't think anyone's going to see this comment if they aren't specifically looking for it.

dataactbroker/handlers/fileHandler.py
```diff
@@ -406,19 +418,22 @@ def uploadFile(self):
...
-            return JsonResponse.create(exc, exc.status)
+            return JsonResponse.error(exc, exc.status)
 
         submission_id = requestDict.getValue("submission_id")
         start_date = requestDict.getValue("start")
         end_date = requestDict.getValue("end")
-
+        job = jobDb.getJobBySubmissionFileTypeAndJobType(submission_id, self.EXTERNAL_FILE_TYPE_MAP["D1"], "file_upload")
+        job.start_date = datetime.strptime(start_date,"%m/%d/%Y").date()
+        job.end_date = datetime.strptime(end_date,"%m/%d/%Y").date()
```
cmc333333: Do we need any validation here? If the data comes from the user, do we risk and explosion?

cmc333333: Unrelated, but it looks like this job isn't used until lines 448. Would it make sense to move this creation logic closer to where the job's used?

.gitignore
jworcestBAH: That's a different runTests, I was trying to have a script that would send the pytest shell command to run both the integration and unit tests, and I assumed we didn't want that in the repo.  I'll rename this to something less confusing.

dataactbroker/handlers/fileHandler.py
jworcestBAH: I will add my notes on the D1 changes that are needed to DB-925, I have a list of the places that got these TODO notes.

dataactbroker/handlers/fileHandler.py
jworcestBAH: The frontend has users pick from a calendar that sends the dates in that format, so there is no free entry.  If a bad format is sent, this will raise an exception and return a 500.  We could catch that and return a 400 with a more generic message if that would be preferable.

jworcestBAH: These lines are storing the data in the jobs table to be returned in the later call to /check_generation_status/.  448 stores other info in the jobs table for the same purpose.  My thinking in putting these lines here was to be closer to where start_date and end_date are created so it's clearer what is being stored.  That said, I can see how it could be clearer to have all the job-related lines in one place, so I'll move these down to 448.

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
 
         return JsonResponse.create(StatusCode.OK, response)
 
+    def loadGenerateRequest(self):
```
cmc333333: Not a big deal, but I found this method name confusing. Would something like `getSubmissionAndFileType` be too on the nose?

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
...
+                                    StatusCode.CLIENT_ERROR)
+            return JsonResponse.error(exc, exc.status)
+
+        submission_id = requestDict.getValue("submission_id")
+        file_type = requestDict.getValue("file_type")
+        return submission_id, file_type
+
+    def generateFile(self):
+        """ Start a file generation job for the specified file type """
+        submission_id, file_type = self.loadGenerateRequest()
```
cmc333333: Will this explode if the request is not formatted? `loadGenerateRequest` will return a single element, not a tuple.

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +497,134 @@ def checkD1File(self, submission_id=None):
...
+                        result = "failed"
+                        uploadJob.error_message = "Validation completed but row-level errors were found"
+                    else:
+                        result = "finished"
+        self.interfaces.jobDb.session.commit()
+        return result
+
 
     def generateD2File(self):
         """ Initiates the generation of D2 """
```
jworcestBAH: Incidentally, there is a good bit of repeated code here from the D1 function.  The next PR will be consolidating these along with E and F to reduce that replication.  The reason that's being done as a separate PR is that it will require more coordination with the frontend to ensure the routes being removed are no longer used.

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
...
+        """ Start a file generation job for the specified file type """
+        submission_id, file_type = self.loadGenerateRequest()
+        job = self.interfaces.jobDb.getJobBySubmissionFileTypeAndJobType(submission_id, self.EXTERNAL_FILE_TYPE_MAP[file_type], "file_upload")
+        # Check prerequisites on upload job
+        if not self.interfaces.jobDb.runChecks(job.job_id):
+            exc = ResponseException("Must wait for completion of prerequisite validation job", StatusCode.CLIENT_ERROR)
+            return JsonResponse.error(exc, exc.status)
+        if file_type == "D1":
+            # TODO could this function return S3 file location
+            result = self.generateD1File()
```
cmc333333: Newbie question, but it looks like all of the communications between these functions is implicit via the shared `self.request` object. Given that we just derived the `job` and `submission_id`, though, would it make sense to pass them along? That'd make writing unit tests around `generateD1File` significantly simpler, too.

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
...
+            exc = ResponseException("File type must be either D1, D2, E or F", StatusCode.CLIENT_ERROR)
+            return JsonResponse.error(exc, exc.status)
+
+        # TODO mark S3 file location in DB
+        # Mark file generation upload as finished
+        self.interfaces.jobDb.markJobStatus(job.job_id,"finished")
+        # Return same response as check generation route
+        return self.checkGeneration()
+
+    def checkGeneration(self):
```
cmc333333: Similar here: why not pass in the `submission_id` and `file_type` rather than re-deriving it?

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
...
+        responseDict["status"] = self.mapGenerateStatus(uploadJob, validationJob)
+        responseDict["file_type"] = file_type
+        responseDict["message"] = uploadJob.error_message or ""
+        if uploadJob.filename is None:
+            responseDict["url"] = ""
+        else:
+            if CONFIG_BROKER["use_aws"]:
+                responseDict["url"] = s3UrlHandler().getSignedUrl("d-files",uploadJob.filename, bucketRoute=None, method="GET")
+            else:
+                responseDict["url"] = uploadJob.filename
```
cmc333333: No biggie, but these nested ifs can be collapsed to be less pointy:
```python
if uploadJob.filename is None:
    ...
elif CONFIG_BROKER["use_aws"]:
   ...
else:
   ...
```

dataactbroker/handlers/fileHandler.py
jworcestBAH: Good catch, changed loadGenerateRequest to raise an exception in that case.  Exceptions get translated to the appropriate response object in dataactbroker.routeUtils

dataactbroker/handlers/fileHandler.py
```diff
@@ -479,9 +498,134 @@ def checkD1File(self, submission_id=None):
...
+                # If validation job exists, check that status as well
+                if validationStatus in ["waiting", "ready", "running"]:
+                    result = "waiting"
+                elif validationStatus in ["invalid", "failed"]:
+                    if validationJob.error_message is None:
+                        if uploadJob.error_message is None:
+                            if validationStatus == "invalid":
+                                uploadJob.error_message = "Generated file had file-level errors"
+                            else:
+                                uploadJob.error_message = "Validation job had an internal error"
```
cmc333333: The logic here is nested _very_ deeply (we're at 6 levels of `if`). Is there a way to simplify the logic so that it's easier to follow? Right now I need to effectively execute the function in my head to figure out which path is taken.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Only thing there is that it looks a lot like our naming of interface functions to pull from DB.  I was hoping to make this clear that it's pulling from the http request, would something like getRequestParams make more sense, maybe getRequestParamsForGenerate since this only applies to the file generation routes?

dataactbroker/handlers/fileHandler.py
cmc333333: Trusting JavaScript for input validation is asking for trouble; is it safe to say the rest of the application does this? We can make a separate issue if so.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Right, this is in an awkward halfway state where generateD1File and generateD2File can be called directly from the old routes.  The next PR will be removing those routes and consolidating the generateD1File and generateD2File functions into a more generic function that handles E and F as well.  The consolidated function will certainly take submission ID and file type as arguments, probably the job as well depending on what it returns.

dataactbroker/handlers/fileHandler.py
jworcestBAH: I could make those optional arguments here, but as this function is the first called from the /check_generate_status/ route it needs to be able to pull the values from the request if they're not provided.

dataactcore/aws/s3UrlHandler.py
```diff
@@ -43,6 +43,7 @@ def _signUrl(self,path,fileName,bucketRoute,method="PUT") :
         """
         if(s3UrlHandler.ENABLE_S3) :
             s3connection = boto.s3.connect_to_region(s3UrlHandler.REGION)
+            print(str(s3connection))
```
cmc333333: Looks like a stray debug line

dataactbroker/handlers/fileHandler.py
jworcestBAH: Added that option.

dataactbroker/handlers/jobHandler.py
```diff
@@ -259,6 +259,8 @@ def addUploadJobs(self,filenames,submissionId,existingSubmission):
         jobsRequired = []
         # Dictionary of upload ids by filename to return to client
         uploadDict = {}
+        d1ValId = None
+        cValId = None
```
cmc333333: If I understand the logic correctly, these vars won't be set unless the `filenames` are in a specific order. Is there a way to be more explicit? Maybe we pull the file C and D1 files out of that list and run them first?

dataactbroker/handlers/fileHandler.py
jworcestBAH: Thanks, fixed

dataactcore/models/jobTrackerInterface.py
```diff
@@ -229,6 +230,62 @@ def checkJobDependencies(self,jobId):
...
+        Args:
+        jobId -- ID of job to be checked
+        Returns:
+        True if all passed, otherwise False or exception
+        """
+        if(self.checkJobReady(jobId) and self.checkPrerequisites(jobId)):
+            # All passed
+            return True
+        else:
+            return False
```
cmc333333: Might be useful to simplify to:
```python
return self.checkJobReady(jobId) and self.checkPrerequisites(jobId)
```

dataactcore/aws/s3UrlHandler.py
jworcestBAH: Yep, removed

dataactcore/models/jobTrackerInterface.py
```diff
@@ -229,6 +230,62 @@ def checkJobDependencies(self,jobId):
...
+        result = self.checkJobUnique(query)
+        # Found a unique job
+        if(result.job_status_id != self.getJobStatusId("ready")):
+            # Job is not ready
+            # Job manager is not yet implemented, so for now doesn't have to be ready
+            return True
+            # TODO when job manager exists, change to exception below
+            #exc = ResponseException("Job is not ready",StatusCode.CLIENT_ERROR,None,ValidationError.jobError)
+            #exc.status = 400
+            #raise exc
```
cmc333333: Not a big deal, but rather than commenting all of this code out, it might be cleaner to just exit early, e.g.
```python
def checkJobReady(self, jobId):
    # Comment explaining why we're exiting early
    return True
    query = ...
```
I think the benefit there is that readers don't need to look through branches to figure out that the function is only stubbed.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -229,6 +230,62 @@ def checkJobDependencies(self,jobId):
...
+
+        return True
+
+    def checkPrerequisites(self, jobId):
+        """ Checks that specified job has no unsatisfied prerequisites
+        Args:
+        jobId -- job_id of job to be run
+
+        Returns:
+        True if prerequisites are satisfied, False otherwise
```
cmc333333: Note that this doesn't actually return False if the prereqs aren't satisfied; it raises an exception

dataactcore/models/jobTrackerInterface.py
jworcestBAH: So this is much older code that is just being moved into the core in this PR.  Actually the job manager is now implemented, but it has been useful in the past to be able to rerun validations on a finished job, so I think I will remove this check entirely.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -229,6 +230,62 @@ def checkJobDependencies(self,jobId):
...
+        """
+        # Get list of prerequisites
+        queryResult = self.session.query(JobDependency).filter(JobDependency.job_id == jobId).all()
+        for prereq in queryResult:
+            query = self.session.query(Job).filter(Job.job_id == prereq.prerequisite_id)
```
cmc333333: It looks like our ORM models aren't using Sqlalchemy's [relationships](http://docs.sqlalchemy.org/en/latest/orm/tutorial.html#building-a-relationship), which would make this a bit simpler. Rather than crafting our own queries, relationships push that logic into the ORM.

Not worth holding anything back, but worth considering.

cmc333333: Somewhat related, would it be worthwhile to just move all of this Python logic into the query? I don't know the Sqlalchemy functions well enough, but using a made-up syntax:
```python
query = self.session.query(JobDependency).join(Job, on=JobDependency.job_id == Job.job_id).filter(
    Job.status_id=="finished")
if query.exists():
    raise Exception
```

tests/integration/fileTests.py
```diff
@@ -433,6 +438,61 @@ def test_metrics(self):
...
+        self.assertEqual(json["status"], "failed")
+        self.assertEqual(json["file_type"], "F")
+        self.assertEqual(json["url"],"")
+        self.assertEqual(json["message"],"File was invalid")
+
+        # E file generation should error because D1 job did not validate
+        postJson = {"submission_id": self.generation_submission_id, "file_type": "E", "start":"01/02/2016", "end":"02/03/2016"}
+        response = self.app.post_json("/v1/generate_file/", postJson, headers={"x-session-id":self.session_id}, expect_errors=True)
+
+        print("Response: " + str(response.json))
```
cmc333333: Looks like a stray debug line

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Fixed the docstring

dataactbroker/fileRoutes.py
```diff
@@ -102,3 +102,17 @@ def get_protected_files():
...
+
+    @app.route("/v1/generate_file/", methods=["POST"])
+    @permissions_check
+    def generate_file():
+        """ Generate file from external API """
+        fileManager = FileHandler(request, isLocal=IS_LOCAL, serverPath=SERVER_PATH)
+        return RouteUtils.run_instance_function(fileManager, fileManager.generateFile)
+
+    @app.route("/v1/check_generation_status/", methods=["POST"])
+    @permissions_check
```
cmc333333: Maybe out of scope, but I would assume that a "check" function would respond to a `GET` rather than a `POST`. I'd expect something like:

http://example.com/v1/generation_status/D1?session=mytoken

dataactbroker/handlers/fileHandler.py
cmc333333: Both options sound better to me

cmc333333:
Sorry for the delay; done reviewing now. I added lots of comments, but they're all take-em-or-leave-em. The one thing we do need to get sorted out before merging are the test failures

dataactbroker/handlers/fileHandler.py
jworcestBAH: Changed to use the second approach

dataactbroker/handlers/fileHandler.py
jworcestBAH: Changed to return a 400 and raise a ResponseException rather than a 500.

jworcestBAH:
The test failures are from Jenkins having an older version of the config file, will update that after I get through the other changes.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Okay, changed

dataactbroker/handlers/fileHandler.py
jworcestBAH: Simplified this section using a couple dictionary lookups, should be clearer now.

dataactcore/models/jobTrackerInterface.py
jworcestBAH: This got simplified down to a single function with no ready check

dataactcore/models/jobTrackerInterface.py
jworcestBAH: For relationships, we are using those in some places, but they are only populated if you explicitly do a joinedload or a subqueryload, so I have mostly been avoiding relying on them as they will be empty when passing around an instance created without those.  That's not the case here, so we could use this here, I'll take a look after I figure out the C and D1 issue above.

tests/integration/fileTests.py
jworcestBAH: Removed

dataactbroker/fileRoutes.py
jworcestBAH: These routes were defined in https://gist.github.com/kevinli-work/59f7310afc789fdeeb7b0993772df03d, so I followed what was in that.  This route does take a couple parameters, which is why it's a post route.  We could make this a GET route and have the parameters be part of the URL, but that isn't really consistent with how we've done similar routes like /v1/check_status/.  If we want to change this pattern, it should probably be its own story to go through and identify the routes that should be done this way, with hours included for the necessary frontend changes to use the altered routes.

dataactbroker/handlers/jobHandler.py
jworcestBAH: Refactored this to run C and D1 before the rest

dataactcore/models/jobTrackerInterface.py
jworcestBAH: Refactored to use relationships

jworcestBAH:
Finished with my changes, let me know if there's anything I missed.

dataactbroker/handlers/fileHandler.py
```diff
@@ -576,47 +578,42 @@ def mapGenerateStatus(self, uploadJob, validationJob = None):
...
-                elif validationStatus == "finished":
-                    if errorsPresent:
-                        result = "failed"
-                        uploadJob.error_message = "Validation completed but row-level errors were found"
-                    else:
-                        result = "finished"
+                uploadJob.error_message = "Validation job had an internal error"
+
+        elif uploadJob.error_message is None:
+            uploadJob.error_message = validationJob.error_message
```
cmc333333: Very nice refactor! The branching is easier to follow now.

cmc333333:
Lots of back and forth here; thanks for tolerating my digressions! Better thanks for writing tests :)

PR #223, opened 2016-08-25
Introduction of Factory Boy

cmc333333:
Factories should make generating nonsense data a little bit easier, which means our tests can focus on only the essential information. This includes factories for SF133 and Appropriations and uses them in one of the unittests as a demonstration.


bsweger:
This looks good @cmc333333 (have already looked at much of this code in the FSRS PR)-- thanks for introducing this.

PR #225, opened 2016-08-25
re-instate sf133 loader

bsweger:
Re-instate the current SF 133 loader so we can load test data.


jworcestBAH:
Looks good, merging

PR #227, opened 2016-08-29
Db 843 drop d routes

jworcestBAH:
This PR removes the old routes for D1/D2 and consolidates external file generation functionality.

Merging should wait until frontend no longer uses the old routes, and possibly should wait until the PR Nipun will be submitting with changes to file generation

- [x] Backend code review is complete
- [x] Frontend confirms old D1/D2 routes are no longer used
- [x] Nipun confirms this PR does not need to wait on DB-897


dataactbroker/handlers/fileHandler.py
```diff
@@ -244,6 +244,44 @@ def finalize(self, jobId=None):
...
+        except ResponseException as exc:
+            if isinstance(exc.wrappedException, NoResultFound):
+                # Submission does not exist, change to 400 in this case since route call specified a bad ID
+                exc.status = StatusCode.CLIENT_ERROR
+                message = "Submission does not exist"
+                error_occurred = True
+            else:
+                raise exc
+        try:
+            self.checkSubmissionPermission(submission)
```
bsweger: It seems like we have to do this check a lot throughout the file route code. Does it make sense to refactor to use some kind of decorator instead?

dataactbroker/handlers/fileHandler.py
```diff
@@ -411,48 +449,62 @@ def uploadFile(self):
             # Unexpected exception, this is a 500 server error
             return JsonResponse.error(e,StatusCode.INTERNAL_ERROR)
 
-    def generateD1File(self):
-        """ Initiates the generation of D1 """
-        jobDb = self.interfaces.jobDb
-        requestDict = RequestDictionary(self.request)
+    def startGenerationJob(self, submission_id, file_type):
+        """ Initiates the generation of D1
```
bsweger: Thanks for updating the docstring--does this part specific to the D1 file need to be changed too?

dataactbroker/handlers/fileHandler.py
```diff
@@ -411,48 +449,62 @@ def uploadFile(self):
...
 
-        if not (StringCleaner.isNumeric(submission_id) and StringCleaner.isDate(start_date) and StringCleaner.isDate(end_date)):
-            exc = ResponseException("submission id, start, and/or end cannot be parsed into their appropriate types", StatusCode.CLIENT_ERROR)
-            return JsonResponse.error(exc, exc.status)
+        """
+        jobDb = self.interfaces.jobDb
+        file_type_name = self.EXTERNAL_FILE_TYPE_MAP[file_type]
+
+        if file_type in ["D1", "D2"]:
+            # Populate start and end dates
```
bsweger: Would it be useful to expand this comment to indicate whether these incoming start/end dates represent calendar year or fiscal year dates?

dataactbroker/handlers/fileHandler.py
```diff
@@ -411,48 +449,62 @@ def uploadFile(self):
...
-        end_date = requestDict.getValue("end")
+        Returns:
+            Tuple of boolean indicating successful start, and error response if False
 
-        if not (StringCleaner.isNumeric(submission_id) and StringCleaner.isDate(start_date) and StringCleaner.isDate(end_date)):
-            exc = ResponseException("submission id, start, and/or end cannot be parsed into their appropriate types", StatusCode.CLIENT_ERROR)
-            return JsonResponse.error(exc, exc.status)
+        """
+        jobDb = self.interfaces.jobDb
+        file_type_name = self.EXTERNAL_FILE_TYPE_MAP[file_type]
```
bsweger: I realize EXTERNAL_FILE_TYPE_MAP wasn't introduced in this PR, but for consistency does it make sense to load that dictionary from the file_type table (which would perhaps need an additional user-submitted vs generated flag)?

dataactbroker/handlers/fileHandler.py
```diff
@@ -411,48 +449,62 @@ def uploadFile(self):
...
-        d_file_id = self.jobManager.createDFileMeta(submission_id, start_date, end_date, "d1", CONFIG_BROKER["award_procurement_file_name"], upload_file_name)
-        job = jobDb.getJobBySubmissionFileTypeAndJobType(submission_id, self.EXTERNAL_FILE_TYPE_MAP["D1"], "file_upload")
-        try:
-            job.start_date = datetime.strptime(start_date,"%m/%d/%Y").date()
-            job.end_date = datetime.strptime(end_date,"%m/%d/%Y").date()
-        except ValueError as e:
-            # Date was not in expected format
-            raise ResponseException(str(e),StatusCode.CLIENT_ERROR,ValueError)
+
+        job = jobDb.getJobBySubmissionFileTypeAndJobType(submission_id, self.EXTERNAL_FILE_TYPE_MAP[file_type_name], "file_upload")
```
bsweger: Related question about the dictionaries at the top of `fileHandler.py`...should the mapping of the file abbreviations (e.g., D2), be stored as part of the file_type table?

dataactbroker/handlers/fileHandler.py
```diff
@@ -507,6 +552,9 @@ def checkGeneration(self, submission_id = None, file_type = None):
         """
         if submission_id is None or file_type is None:
             submission_id, file_type = self.getRequestParamsForGenerate()
+        # Check permission to submission
+        self.checkSubmissionById(submission_id, file_type)
```
bsweger: Is the logic in this `checkSubmissionById` function a good contender to user for decorating some of these functions? Or is there another way we can elevate permissions checking so that it's done before we get here?

I actually don't know that right answer..pinging @cmc333333 and @nmonga91 to see if they have .02.

dataactcore/config_example.yml
```diff
@@ -38,17 +38,23 @@ broker:
     # Uses "{}" because it requires string formatting to insert query parameters
     # If the URL is provided, the file name provided will be used when uploading to S3
     d1_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
+    d1_file_name: d1_data.csv
+    award_procurement_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
```
bsweger: Do we want to put the real endpoint in our sample config? I haven't used it so am unsure if there's anything preventing other people from hitting it and possibly endangering the health of the service.

dataactcore/config_example.yml
```diff
@@ -38,17 +38,23 @@ broker:
...
     award_procurement_file_name: d1_data.csv
 
     # File D2 API
     # If the URL is provided, the file name provided will be used when uploading to S3
     d2_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
+    d2_file_name: d2_data.csv
+    award_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
```
bsweger: Same question as above re: the endpoint.

dataactbroker/README.md
```diff
@@ -773,6 +773,153 @@ Example output if there are no files available:
...
+# File Generation Routes
+
+## Generate Files
+**Route:** `/generate_file`
+
+**Method:** `POST`
+
+This route sends a request to the backend to utilize the relevant external APIs and generate the relevant file for the metadata that is submitted.
+
+**Deprecation Notice:** This route replaces `/generate_d1_file` and `/generate_d2_file`.
```
bsweger: Thanks for being explicit about what's deprecated. This raises a good question that I don't think we need to answer right now, but I'll create a post-launch story for it...how will we handle API versioning?

dataactbroker/README.md
```diff
@@ -773,6 +773,153 @@ Example output if there are no files available:
...
+
+### Body Description
+
+* `submission_id` - **required** - an integer representing the ID of the current submission
+* `file_type` - **required** - a string indicating the file type to generate. Allowable values are:
+	* `D1` - generate a D1 file
+	* `D2` - generate a D2 file
+	* `E` - generate a E file
+	* `F` - generate a F file
+* `start` - **required for D1/D2 only** - the start date of the requested period, in `MM/DD/YYYY` string format
```
bsweger: It might be important to be explicit here about what we mean by "period." A calendar date range? A fiscal period?

dataactbroker/README.md
```diff
@@ -773,6 +773,153 @@ Example output if there are no files available:
...
+	"status": "failed",
+	"file_type": "D1",
+	"url": "",
+    "start": "01/01/2016",
+    "end": "03/31/2016",
+	"message": "The server could not reach the Federal Procurement Data System. Try again later."
+}
+```
+
+*State:* No file generation request has been made for this submission ID before
```
bsweger: I'm not sure what _No file generation request has been made for this submission ID before_ means. Does it mean that someone tried to invoke a File D status check before they invoked a request to create it?

bsweger:
@jworcestBAH Done reviewing. Thanks for this--it's great to have this consolidated approach to the routes. Also, thank you, as always, for keeping the docs updated.

The only things I'd flag as really needing changed before we merge are the README/comment suggestions around date/period clarification. The fiscal year/quarter/calendar year confusion on this project is real, so being explicit where we can would be a service to our future colleagues (and selves).

The other suggestions are just that. It would be nice to consider some of them as tech debt tasks as we can, but I don't think they would prevent us from moving forward.

dataactbroker/handlers/fileHandler.py
nmonga91: I think it is a good contender. We would want to check permissions before invoking any code for the generation to avoid conflicts.

cmc333333: One cent per thought:
1. Decorators are great for signifying something is special about a specific function. I think that if what we're flagging is "this method needs valid permissions", a decorator is good. If instead we're saying "fetch me the permissions", a normal function might be a better bet
2. I'd generally recommend separating permissions checking from business logic. If we performed permissions checks in one place, it's easier to test and maintain. We could then proceed in these functions with that "permissions are good" precondition.

dataactbroker/handlers/fileHandler.py
jworcestBAH: It does need access to the submission from the request.  In theory the decorator could pull that out of the request as well, but it may mean pulling the submission twice for each call.  The submit files route only does the check if it's an existing submission, but that could be an option passed to the decorator.  I'll do some testing around this later, I'm not quite sure yet how to set this up.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Good catch, fixed

dataactbroker/handlers/fileHandler.py
jworcestBAH: Expanded the comment to specify calendar year along with formatting

dataactbroker/handlers/fileHandler.py
jworcestBAH: This dict is actually mapping file letters (D1, D2, E, F) to the names.  Currently the letters aren't defined anywhere in the database, but the frontend is calling these routes with just the letter, so we needed this mapping to be available.

dataactbroker/README.md
jworcestBAH: This documentation was taken from Kevin's definition of the routes.  That said, it looks like this left out the /v1/ preceding these routes, so I've added that to the doc

dataactbroker/README.md
jworcestBAH: Changed "period" to "date range", let me know if there's another term that would be clearer

dataactbroker/README.md
jworcestBAH: Right, this is the case where /check_generations_status/ is called before /generate_file/

dataactbroker/handlers/fileHandler.py
bsweger: Sorry--yes, it was a more broad question about storing these important attributes in the table vs. a dict in the code.

dataactbroker/README.md
bsweger: Cool, thanks, I didn't even notice you'd left out the`V1`. It was smart for you to set up the versioning on the API. I created a post-launch story about setting up a process to notify any API users when we upgrade from V1 or make other changes that could impact people.

dataactbroker/README.md
bsweger: Thanks for clarifying.

dataactbroker/handlers/fileHandler.py
jworcestBAH: This map is now built from the file_type table, which now includes letter_name

dataactbroker/handlers/fileHandler.py
jworcestBAH: File type map is now loaded from file_type table

dataactbroker/handlers/fileHandler.py
jworcestBAH: I'm thinking this could be a separate story to make submission level permissions look more like our route-level permissions which do use a decorator.  I'm not going to be able to implement this today as part of this PR, but I'll create a story for it.

bsweger:
@jworcestBAH Thanks for making those changes and for updating the file_type table. It sounds like we need a larger convo (i.e., not a blocker to this PR) on permission checking within the app.

Confirmed that config sample can contain the real D1/D2 endpoints.

dataactbroker/handlers/fileHandler.py
jworcestBAH: This is now in JIRA as https://federal-spending-transparency.atlassian.net/browse/DB-1037

jworcestBAH:
Incorporated changes from D1/D2 API PR

nmonga91:
[run tests]

nmonga91:
Looks good, merging

PR #228, opened 2016-08-29
Start of file F generation

cmc333333:
Configures a mapping between FSRS field names and file F field names. This isn't complete (we'll add more mappings), but is ready for review. The next step will be to convert these to save the generated rows into a CSV.


cmc333333:
[run tests]

klrBAH:
[run tests]

bsweger:
@cmc333333 This looks good. I didn't carefully check each and every field mapping, knowing that this is a first pass. I think this is good to merge, but let me know if you wanted more detailed eyes on those mappings.

cmc333333:
Thanks @bsweger! I think we'll need a more thorough check at some stage, but it doesn't need to be now. As long as you can understand what the mapping say (i.e. the `None` vs `str` vs `tuple`) stuff, I think we're okay.

bsweger:
Yep, I got the structure you set up (thanks for the comments). I think we're good to go--merging!

PR #230, opened 2016-08-30
Mock out sesEmail.send

cmc333333:
This prevents actual emails from being sent during the integration tests


jworcestBAH:
Looks good, merging

PR #231, opened 2016-08-30
DB-897: Generating D files both locally and to S3

nmonga91:


dataactbroker/handlers/fileHandler.py
```diff
@@ -562,7 +567,8 @@ def checkGeneration(self, submission_id = None, file_type = None):
         if uploadJob.filename is None:
             responseDict["url"] = ""
         elif CONFIG_BROKER["use_aws"]:
-            responseDict["url"] = s3UrlHandler().getSignedUrl("d-files",uploadJob.filename, bucketRoute=None, method="GET")
+            path, file_name = uploadJob.filename.split("/")
```
cmc333333: Would it make sense to future proof this a bit by allowing there to be multiple slashes in the filename?
```python
path, file_name = uploadJob.filename.rsplit("/", 1)
```

dataactbroker/handlers/fileHandler.py
```diff
@@ -644,7 +650,12 @@ def generateD2File(self):
         # Generate and upload d2 file to S3
         user_id = LoginSession.getName(session)
         timestamped_name = s3UrlHandler.getTimestampedFilename(CONFIG_BROKER["award_file_name"])
-        upload_file_name = "".join([str(user_id), "/", timestamped_name])
+
+        if self.isLocal:
+            upload_file_name = "".join([CONFIG_BROKER['broker_files'], timestamped_name])
```
cmc333333: Should this incorporate the `user_id` at all? Ideally we'd have a very similar set up between local and AWS.

dataactbroker/handlers/fileHandler.py
nmonga91: I didn't think we'd need it since the download is only coming from the user's directory on the specified S3 bucket. If a user ever gets nested directories, I would agree, but we should check if there will ever be such a scenario.

dataactbroker/handlers/fileHandler.py
nmonga91: Generally local files are stored there (based on how submit files works locally) so that is why this follows the same model. But that can be altered.

dataactcore/config.py
```diff
@@ -77,6 +81,8 @@
     # just use the directory
     broker_files = os.path.split(broker_files)[0]
 normpath(broker_files)
+if broker_files[-1] != os.path.sep:
+    broker_files = "".join([broker_files, os.path.sep])
```
cmc333333: Will this config be used at all if we're using aws? If so, the `os.path.sep` won't match what AWS is expecting (which is always a forward slash).

Minor details:
* `str.endswith()` is a little nicer than `[-1]`. It handles empty strings and the hypothetical case where `os.path.sep` is not a single character
* the `"".join(` idiom is a little strange for string concatenation. Particularly if you're just joining two strings, a plus is more common:
```python
broker_files = broker_files + os.path.sep
# or
broker_files += os.path.sep
```

dataactcore/utils/jobQueue.py
```diff
@@ -36,52 +38,79 @@ def enqueue(jobID):
...
-                job_manager.setDFileStatus(d_file_id, "failed")
+                self.update_d_file_status(job_manager, d_file_id, "failed")
                 raise e
 
         self.enqueue = enqueue
         self.generate_d_file = generate_d_file
 
+    def get_xml_response_content(api_url):
+        """ Retrieve XML Response from the provided API url """
+        return str(requests.get(api_url, verify=False).content)
```
cmc333333: `requests` has a `text` attribute that'll do the string encoding for you:
```python
return requests.get(api_url, verify=False).text
```

nmonga91: Good call, i'll add that in.

dataactcore/config.py
nmonga91: Broker files is used when `use_aws` is set to false, that is why `os.path.sep` is needed. and I agree with the simpler concatenation. Will make that change.

dataactcore/utils/jobQueue.py
```diff
@@ -36,52 +38,83 @@ def enqueue(jobID):
...
+            # get request
+            response = requests.get(file_url)
+            # write to file
+            response.encoding = "utf-8"
+            file.write(response.text)
+
+    @staticmethod
+    def update_d_file_status(job_manager, d_file_id, status):
+        """ Update the D file status to the one specified via the Job Manager """
+        job_manager.setDFileStatus(d_file_id, status)
```
cmc333333: I think this was added to make tests a bit easier, but I wonder if we can avoid it. I don't think we're gaining anything by adding the layer of indirection.

dataactcore/utils/jobQueue.py
```diff
@@ -36,52 +38,83 @@ def enqueue(jobID):
...
+
+    @staticmethod
+    def download_file(local_file_path, file_url):
+        """ Download a file locally from the specified URL """
+        with open(local_file_path, "w") as file:
+            # get request
+            response = requests.get(file_url)
+            # write to file
+            response.encoding = "utf-8"
+            file.write(response.text)
```
cmc333333: Nice and clean function

dataactcore/utils/jobQueue.py
nmonga91: I think we can avoid this if we make `job_manager` a static variable and mock that. That's something I can play around with. What do you think?

tests/unit/dataactcore/test_job_queue.py
```diff
@@ -0,0 +1,46 @@
...
+from dataactcore.config import CONFIG_BROKER
+from unittest.mock import Mock
+from pytest import raises
+from dataactcore.utils.responseException import ResponseException
+import os.path
+
+
+def test_generate_d_file_success(monkeypatch):
+    """ Test successful generation of D1 and D2 files """
+    local_file_name = "12345_test_file.csv"
```
cmc333333: The test here isn't cleaning up after itself, which means that we could delete lines 15 through 16 and have "passing" tests on subsequent runs. Some options:

* the  `[tmpdir](http://doc.pytest.org/en/latest/tmpdir.html)` fixture was made for this kind of work, but it'd be a bit tricky to integrate with our config system. Effectively, we'd want the tests to modify the `CONFIG_BROKER` at set up and reset it at tear down. This'd be super useful in the future, but might not be worth figuring out right now
* using a random file name helps us a little. The file's not deleted at the end of the test, but at least we'd generally catch errors

tests/unit/dataactcore/test_job_queue.py
```diff
@@ -0,0 +1,46 @@
...
+    result_xml = "<results>test_file.csv</results>"
+
+    jq = JobQueue()
+    monkeypatch.setattr(jq, 'update_d_file_status', Mock())
+    monkeypatch.setattr(jq, 'download_file', Mock())
+    monkeypatch.setattr(jq, 'get_xml_response_content', Mock(return_value=result_xml))
+
+    jq.generate_d_file(Mock(), 1, 1, Mock(), local_file_name, True)
+
+    assert os.path.isfile(file_path)
```
cmc333333: Minor, but one testing style that I think is particularly good:
1. assert stable state
2. run function
3. assert changed state

Here that'd look like:
```python
assert not os.path.isfile(file_path)
jq.generate_d_file(Mock(), 1, 1, Mock(), local_file_name, True)
assert os.path.isfile(file_path)
```

This not only shows the reader what's changing, but also avoids false negatives

tests/unit/dataactcore/test_job_queue.py
```diff
@@ -0,0 +1,46 @@
...
+        file.write("test")
+
+    result_xml = ""
+
+    jq = JobQueue()
+    monkeypatch.setattr(jq, 'update_d_file_status', Mock())
+    monkeypatch.setattr(jq, 'get_xml_response_content', Mock(return_value=result_xml))
+
+    with raises(ResponseException) as excinfo:
+        jq.generate_d_file(Mock(), 1, 1, Mock(), local_file_name, True)
```
cmc333333: Nice. Does it make sense to also verify this exception is raised when the file simply doesn't exist?

dataactcore/utils/jobQueue.py
cmc333333: I'd recommend staying away from static variables if you can.

Since we're mocking out `interface_holder` already (from which `job_manager` is derived), are we done?

cmc333333:
Looks good @nmonga91 -- I added some comments on bits and pieces, but this is headed in a great direction

tests/unit/dataactcore/test_job_queue.py
nmonga91: There's nothing that checks if the file actually exists locally until you try to download via the URL provided in the response so I would leave it out for now.

cmc333333:
@nmonga91 I fixed a little bug in the unit tests. I think this all looks good, so if that last changeset makes sense to you, let's merge

cmc333333:
@nmonga91 pointed out the file name issue, but otherwise was cool with the changes, so I'm merging.

PR #233, opened 2016-08-31
Mark files E and F as finished while skipping generation

jworcestBAH:
This will mark these files as finished so the frontend doesn't get stuck waiting for them

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -509,8 +509,7 @@ class FileHandler:
             jq.generate_d_file.delay(get_url, user_id, job.job_id, InterfaceHolder, timestamped_name, self.isLocal)
         else:
             # TODO add generate calls for E and F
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -509,6 +509,8 @@ def startGenerationJob(self, submission_id, file_type):
             jq.generate_d_file.delay(get_url, user_id, job.job_id, InterfaceHolder, timestamped_name, self.isLocal)
         else:
             # TODO add generate calls for E and F
+            job.job_status_id = jobDb.getJobStatusId("finished")
```
nmonga91: instead of doing this directly in fileHandler, shouldn't this be calling a function in jobHandler, something similar to `setDFileStatus`?

jworcestBAH: Right, changed to use markJobStatus

nmonga91:
Looks good, merging

PR #234, opened 2016-08-31
Only creating file generation error message if status is failed

jworcestBAH:
Fixed a bug with error messages being created on jobs with no errors


nmonga91:
Looks good, merging

PR #236, opened 2016-08-31
Correcting zip_4 field name and marking business fields optional

jworcestBAH:
These changes are needed to avoid errors in reading the D2 files


nmonga91:
Looks good, will merge once Jenkins is done

jworcestBAH:
Ah right sample files need to match the field name change

bsweger:
@jworcestBAH should we update the *Raw.csv version of these files too? I tripped myself up once because I reloaded from those, and they didn't match their counterparts.

PR #238, opened 2016-09-01
Add file F asynchronous task

cmc333333:
This also refactors the jobQueue quite a bit to make it easier to test. It's probably easiest to step through these changesets one by one rather than trying to look at the whole diff at once.


dataactcore/utils/jobQueue.py
```diff
@@ -1,120 +1,158 @@
...
-from csv import reader
 from dataactcore.utils.jsonResponse import JsonResponse
+from dataactcore.utils import fileF
 from dataactcore.utils.responseException import ResponseException
 from dataactcore.utils.statusCode import StatusCode
 
+
+def enqueue(jobID):
+    """POST a job to the validator"""
+    validatorUrl = '{validator_host}:{validator_port}'.format(
```
nmonga91: Great way to implement string interpolation!

cmc333333: Thank you!

dataactcore/utils/jobQueue.py
```diff
@@ -1,120 +1,158 @@
...
+        csv_writer = CsvS3Writer(region, bucket, file_name, header)
+
+    with csv_writer as writer:
+        for line in body:
+            writer.write(line)
+        writer.finishBatch()
+    return file_name
+
+
+@contextmanager
```
nmonga91: What does this decorator do? i'm not familiar with it

tests/unit/dataactcore/test_job_queue.py
```diff
@@ -1,55 +1,76 @@
...
 
-    with raises(ResponseException):
-        jq.generate_d_file(Mock(), 1, 1, Mock(), local_file_name, True)
+    assert os.path.isfile(file_path)
+
+    with open(file_path) as f:
+        return [row for row in csv.reader(f)]
+
+
+def test_generate_f_file(monkeypatch, mock_broker_config_paths):
```
nmonga91: Is this supposed to cover both success and failure?

dataactcore/utils/jobQueue.py
cmc333333: The `contextmanager` decorator is a shorthand for creating "contexts" (`with` statements). To create something like
```python
with mycontext() as ctx:
    do_something_with(ctx)
```
we can either create an object with `__enter__` and `__exit__` magic methods or we can use the `contextmanager` decorator. These two examples are more or less equivalent:

```python
class mycontext:
    def __enter__(self):
        some_setup_here()
        return "context"

    def __exit__(self):
        some_teardown_here()


@contextmanager
def mycontext():
    some_setup_here()
    yield "context"
    some_teardown_here()
```
As you can see, the later's much easier to read.

tests/unit/dataactcore/test_job_queue.py
cmc333333: There's not really a "fail" state, here -- we'd just not get any results. This tests focuses on how the CSVs are generated rather than the underlying queries (which are tested elsewhere)

nmonga91: gotcha, then this looks good to me

dataactcore/utils/jobQueue.py
nmonga91: Thank you for clarifying!

nmonga91:
[run tests]

nmonga91:
Looks good, merging!

cmc333333:
I'm going to close this PR in favor of one built off #248; trying to separate out the file_d stuff would be too difficult in a merge

PR #239, opened 2016-09-01
Added File D1 model and schema

jworcestBAH:
This adds the staging table for D1 so the C-D1 validations can be started.  Also includes the field definition files for populating the file_columns table


dataactvalidator/filestreaming/schemaLoader.py
```diff
@@ -10,7 +10,7 @@ class SchemaLoader(object):
     This class will load a schema file and writes the set of validation rules to validation database.
 
     """
-    fieldFiles = {"appropriations":"appropFields.csv","award":"awardFields.csv","award_financial":"awardFinancialFields.csv","program_activity":"programActivityFields.csv"}
+    fieldFiles = {"appropriations":"appropFields.csv","award":"awardFields.csv","award_financial":"awardFinancialFields.csv","program_activity":"programActivityFields.csv","award_procurement":"awardProcurementFields.csv"}
```
bsweger: Not sure how much work this would be, but do you think it's worth renaming our current `award` type file to something like `award_assistance` for consistency/clarity?

bsweger:
@jworcestBAH this looks good..note that I didn't want to hold you up, so I didn't go through and match the model/csv fields against the IDD, assuming that anything wrong there would shake out as you proceed with the story.

I left one comment about our current award file that we might want to consider as we wrap up the D file work this sprint. But that thought is beyond the scope of this PR.

PR #243, opened 2016-09-02
Hotfix interface closing dev

jworcestBAH:
Interface holder now calls close method of BaseInterface rather than duplicating the functionality.  This also resolves an issue where a new interface holder would attempt to reuse closed sessions.

### DIFF
--- a/dataactbroker/handlers/interfaceHolder.py
+++ b/dataactbroker/handlers/interfaceHolder.py
@@ -25,35 +25,5 @@ class InterfaceHolder:
 
     def close(self):
         """ Close all open connections """
-        InterfaceHolder.closeOne(self.jobDb)
-        InterfaceHolder.closeOne(self.errorDb)
-        InterfaceHolder.closeOne(self.userDb)
-        InterfaceHolder.closeOne(self.validationDb)
-
-    @staticmethod
-    def closeOne(interface):
-        """ Close all aspects of one interface """
-        if(interface == None):
-            # No need to close a nonexistent connection
-            return
-        try:
-            if(interface.session == None):
-                # If session is None, skip closing
-                return
-        except AttributeError as e:
-            # If interface has no session, skip closing
-            return
-
-        # Try to close the session and connection, on error try a rollback
-        try:
-            interface.session.close()
-        except:
-            try:
-                interface.session.rollback()
-                interface.session.close()
-            except Exception as e:
-                exc_type, exc_obj, exc_tb = sys.exc_info()
-                trace = traceback.extract_tb(exc_tb, 10)
-                CloudLogger.logError('Broker DB Interface Error: ', e, trace)
-                del exc_tb
-                raise
+        if self.jobDb is not None:
+            self.jobDb.close()
--- a/dataactbroker/permissions.py
+++ b/dataactbroker/permissions.py
@@ -39,7 +39,7 @@ def permissions_check(f=None,permissionList=[]):
                                 break
 
                     finally:
-                        InterfaceHolder.closeOne(userDb)
+                        userDb.close()
                     if(validUser) :
                         return f(*args, **kwargs)
                     errorMessage  = "Wrong User Type"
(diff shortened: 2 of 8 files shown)


dataactbroker/handlers/interfaceHolder.py
```diff
@@ -25,35 +25,5 @@ def __init__(self):
...
-                interface.session.rollback()
-                interface.session.close()
-            except Exception as e:
-                exc_type, exc_obj, exc_tb = sys.exc_info()
-                trace = traceback.extract_tb(exc_tb, 10)
-                CloudLogger.logError('Broker DB Interface Error: ', e, trace)
-                del exc_tb
-                raise
+        if self.jobDb is not None:
+            self.jobDb.close()
```
bsweger: I love the simplification we're going for here. What I'm unsure of is why jobDB is the remaining interface that needs to be explicitly closed here. Is this somehow not covered by the cleanup code in the validation and broker `app.py`?

dataactbroker/permissions.py
```diff
@@ -39,7 +39,7 @@ def decorated_function(*args, **kwargs):
                                 break
 
                     finally:
-                        InterfaceHolder.closeOne(userDb)
+                        userDb.close()
```
bsweger: Does this mean we're destroying the userdb engine every time this decorator is used? Could that be a contributing factor to the multitude of db connections that @kevinli-work and @klrBAH were discussing yesterday?

I get that this PR doesn't change the behavior and is just executing it a different way. Could well be something for me to look at when I'm doing the db story, but am curious if you know anything about the original intent.

dataactvalidator/interfaces/interfaceHolder.py
```diff
@@ -23,31 +23,5 @@ def __init__(self):
...
-            interface.session.close()
-        except:
-            try:
-                interface.session.rollback()
-                interface.session.close()
-            except:
-                pass
\ No newline at end of file
+        if self.jobDb is not None:
+            self.jobDb.close()
```
bsweger: Same question as above, for the broker interfaceHolder...this isn't covered by the cleanup code in app.py?

bsweger:
@jworcestBAH done reviewing. thanks for this--there's a few questions above, since I want to understand what's happening. Happy to discuss synchronously if that works better.

dataactbroker/handlers/interfaceHolder.py
jworcestBAH: So what's happening here is that when InterfaceHolder is closed, all the interfaces share the connection, so only one needs to be closed to close the connection and set BaseInterface.interfaces back to None.  It doesn't matter which is closed, so I chose jobDb, any of them would work.

dataactbroker/permissions.py
jworcestBAH: Yes, this decorator is creating and closing a connection.  It happens before the route gets started and calls RouteUtils.run_instance_function, which creates the connection that will be used during this route call.

dataactbroker/handlers/interfaceHolder.py
jworcestBAH: I will change the cleanup code in app.py to call this, the idea is to have that this repeated in different places but to instead always redirect it to the BaseInterface.close() method

jworcestBAH: Actually, I could change BaseInterface.close() to be a static method, that would make this cleaner

dataactbroker/permissions.py
bsweger: Okay--thanks for the explanation. I'll look at this more when digging more deeply int o the db code.

dataactbroker/handlers/interfaceHolder.py
bsweger: I think part of what's tripping me up here is that I can see the teardown happening at the end of the route, but can't see a corresponding setup happening at the beginning of the route. Where are the BaseInterface interfaces being set?

jworcestBAH:
This does not seem to be working on sandbox, so we may need a different approach

jworcestBAH:
These PRs are pre-empted by the 1097 PR, so I'm closing them

PR #245, opened 2016-09-06
Refactor database connection manager

cmc333333:
One step towards centralizing database connection management. This uses `flask.g` where available to resolve some threading issues while also pulling database connection info into a separate class. Ideally, we'll remove all of the database information from `BaseInterface` and instead pass it along explicitly or grab it from the thread-safe context, `flask.g`.

### DIFF
--- a/dataactbroker/scripts/initialize.py
+++ b/dataactbroker/scripts/initialize.py
@@ -63,7 +63,7 @@ def createAdmin():
                 adminEmail, adminPass, Bcrypt(), permission=2)
             user = userDb.getUserByEmail(adminEmail)
             userDb.addUserInfo(user, "Admin", "SYS", "System Admin")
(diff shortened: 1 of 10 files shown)


jworcestBAH:
Has this been tested in sandbox to see if it resolves issues with d-file threads?

cmc333333:
It has not (I don't know how), but it does resolve a local threading issue (the local server spins up several threads). It still uses a static variable for non-flask work, so I don't think it'd resolve a connection issue in the celery worker. That said, it'd be pretty straight forward to wrap the worker function in a `app_context` to get the benefits here.

klrBAH:
[run tests]

klrBAH:
[run tests]

cmc333333:
This became part of #248

PR #246, opened 2016-09-06
add fiscal year to SF_133 index

bsweger:
Realized today that the unique index on the SF-133 is missing `fiscal_year`. We need to get that on there or the load file will fail in a few months.

Apologies for not noticing this until we had a few million rows loaded--the migration takes a few minutes to run.


cmc333333:
:+1: makes sense -- this'd fail in the next fiscal year as we'd have two SF133's with the same tas-period-line

PR #247, opened 2016-09-07
1097 plus database refactor

jworcestBAH:
This contains the changes to D file handling as well as some database connection refactors that reduce the frequency of the 500s we've been getting


jworcestBAH:
Completed the last test I was doing on sandbox, error message now correctly displayed when no data is available.

dataactcore/utils/jobQueue.py
```diff
@@ -1,12 +1,9 @@
...
 import requests
-from dataactvalidator.filestreaming.csvS3Writer import CsvS3Writer
-from dataactvalidator.filestreaming.csvLocalWriter import CsvLocalWriter
-from csv import reader
-from dataactcore.utils.jsonResponse import JsonResponse
-from dataactcore.utils.responseException import ResponseException
-from dataactcore.utils.statusCode import StatusCode
+from dataactcore.config import CONFIG_DB, CONFIG_SERVICES, CONFIG_JOB_QUEUE
+
+job_app = Flask(__name__)
```
cmc333333: This can be deleted as it's no longer used

dataactbroker/app.py
```diff
@@ -58,24 +56,14 @@ def createApp():
...
-            BaseInterface.interfaces = InterfaceHolder()
-        app.before_request(createInterfaces)
+
+        @app.teardown_appcontext
+        def teardown_appcontext(exception):
+            GlobalDB.close()
+
+        @app.before_request
+        def before_request():
+            GlobalDB.db()
```
cmc333333: I like the explicitness, but this isn't actually required (`db()` is called whenever retrieving one of the fields). Either way is good.

jworcestBAH:
[run tests]

dataactbroker/handlers/fileHandler.py
```diff
@@ -494,31 +498,112 @@ def startGenerationJob(self, submission_id, file_type):
...
             # TODO add generate calls for E and F
             jobDb.markJobStatus(job.job_id,"finished")
+
             pass
 
         return True, None
 
+    def get_xml_response_content(self, api_url):
+        """ Retrieve XML Response from the provided API url """
+        return requests.get(api_url, verify=False).text
```
cmc333333: Probably want to throw a timeout on this if we don't trust the API

dataactbroker/handlers/fileHandler.py
```diff
@@ -494,31 +498,112 @@ def startGenerationJob(self, submission_id, file_type):
...
+    def call_d_file_api(self, api_url):
+        """ Call D file API, return True if results found, False otherwise """
+        job_manager = self.interfaces.jobDb
+
+        xml_response = self.get_xml_response_content(api_url)
+        # Check for numFound = 0
+        if(xml_response.find("numFound='0'")!=-1):
+            # No results found, return False
+            return False
+        return True
```
cmc333333: NBD, but more concisely:
```python
return "numFound='0'" not in xml_response
```

dataactbroker/handlers/fileHandler.py
```diff
@@ -642,3 +727,17 @@ def getProtectedFiles(self):
 
         response["urls"] = self.s3manager.getFileUrls(bucket_name=CONFIG_BROKER["static_files_bucket"], path=CONFIG_BROKER["help_files_path"])
         return JsonResponse.create(StatusCode.OK, response)
+
+    def completeGeneration(self, generationId):
+        """ Retrieve the generated file, using the specified ID to look up submission and file type """
```
cmc333333: Not essential, but I guarantee we're going to have no idea what `generationId` and "generated file" mean when we look at this in a few months. Is there something more illuminating that we could note in this docstring? Maybe something like
> For files D1 and D2, we generate a "FileGenerationTask" with a unique identifier "generationId". A third party then posts data to us using that identifier. Handle that post by writing the relevant D file

?

dataactbroker/handlers/fileHandler.py
```diff
@@ -642,3 +727,17 @@ def getProtectedFiles(self):
...
+        if generationId is None:
+            return JsonResponse.error(ResponseException("Must include a generation ID",StatusCode.CLIENT_ERROR), StatusCode.CLIENT_ERROR)
+
+        # Pull url from request
+        safeDictionary = RequestDictionary(self.request)
+        url =  safeDictionary.getValue("href")
+        #Pull information based on task key
+        task = self.interfaces.jobDb.session.query(FileGenerationTask).options(joinedload(FileGenerationTask.file_type)).filter(FileGenerationTask.generation_task_key == generationId).one()
```
cmc333333: Note that this'll blow up if the submission id isn't in our database. Should it 400 instead?

dataactbroker/handlers/fileHandler.py
```diff
@@ -642,3 +727,17 @@ def getProtectedFiles(self):
...
+        job = self.interfaces.jobDb.getJobBySubmissionFileTypeAndJobType(task.submission_id, task.file_type.name, "file_upload")
+        self.load_d_file(url,job.filename,job.original_filename,job.job_id,self.isLocal)
+        return JsonResponse.create(StatusCode.OK,{"message":"File loaded successfully"})
```
cmc333333: Do we want to delete that task here? Maybe we're okay keeping them around forever?

dataactcore/models/jobModels.py
```diff
@@ -104,20 +104,12 @@ class FileType(Base):
...
-    status_id = Column(Integer, ForeignKey("job_status.job_status_id", name="fk_status_id"))
-    status = relationship("JobStatus", uselist=False)
-    error_message = Column(Text)
-    upload_file_name = Column(Text)
-    original_file_name = Column(Text)
-    is_submitted = Column(Boolean, default ="False", server_default="False")
-
-    __table_args__ = (UniqueConstraint('submission_id', 'type', name='_submission_type_uc'),)
+    file_generation_task_id = Column(Integer, primary_key=True)
+    generation_task_key = Column(Text, index=True, unique=True)
```
cmc333333: Not a big deal, but should this be the primary key?

cmc333333:
Lots of good stuff here; I added a bunch of suggestions but the only bits that I think are actually important:
* adding a timeout on the request. IIRC, `requests` will just hang waiting for an unresponsive API. Given the uncertainty with this API, we probably want to head that off at the pass
* accounting for an invalid "generationId"

dataactbroker/app.py
nmonga91: I would agree that I like the explicitness, it makes it very clear what's going on

dataactbroker/handlers/fileHandler.py
```diff
@@ -494,31 +498,112 @@ def startGenerationJob(self, submission_id, file_type):
...
                 job.start_date = datetime.strptime(start_date,"%m/%d/%Y").date()
                 job.end_date = datetime.strptime(end_date,"%m/%d/%Y").date()
                 valJob.start_date = datetime.strptime(start_date,"%m/%d/%Y").date()
                 valJob.end_date = datetime.strptime(end_date,"%m/%d/%Y").date()
+                # Generate random uuid and store generation task
+                task_key = uuid4()
+                task = FileGenerationTask(generation_task_key = task_key, submission_id = submission_id, file_type_id = jobDb.getFileTypeId(file_type_name))
+                jobDb.session.add(task)
+
                 jobDb.session.commit()
```
nmonga91: Might be cleaner to view if these job updates are abstracted out to a separate function

klrBAH:
[run tests]

jworcestBAH:
This PR should not be merged as it is preempted by the other 1097 PR, will close this once the comments have been addressed.

dataactcore/utils/jobQueue.py
jworcestBAH: Deleted in the new PR

dataactbroker/handlers/fileHandler.py
jworcestBAH: This part has always come back so far, but that's not a bad idea to add.  I'll set to 20 seconds for now.

dataactbroker/handlers/fileHandler.py
jworcestBAH: That is nicer, fixed.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Improved this docstring to provide more context

dataactbroker/handlers/fileHandler.py
jworcestBAH: Added some error handling, including returning a 400 for this case with a more specific error message

dataactbroker/handlers/fileHandler.py
jworcestBAH: I'm not sure here, if they re-call the route, should they get an error?  I've already found it useful in debugging to manually re-call these.  In general we've been keeping information for past submissions indefinitely, so I'd be tempted to do the same for these unless we have a reason these should be more temporary.

dataactcore/models/jobModels.py
jworcestBAH: I went with a surrogate key here, though I'm certainly willing to have this be the primary key if people prefer.  Paging @bsweger on this one.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Agreed, breaking this up into smaller functions

jworcestBAH:
Done with the comments here, so I'm closing this in favor of the newer 1097 PR without the extra connection refactoring

PR #248, opened 2016-09-07
Db 1097 generation callback

jworcestBAH:
This is the D file API changes along with the earlier DB connection hotfixes, but not including the start of the DB connection refactor the last PR had.

- [ ] PR reviewed
- [x] Flask debug removed


jworcestBAH:
Note that this branch currently has app.debug = True, we should remove that before merging this, I'll add a checkbox

dataactcore/models/jobModels.py
```diff
@@ -104,20 +104,12 @@ class FileType(Base):
...
-    status = relationship("JobStatus", uselist=False)
-    error_message = Column(Text)
-    upload_file_name = Column(Text)
-    original_file_name = Column(Text)
-    is_submitted = Column(Boolean, default ="False", server_default="False")
-
-    __table_args__ = (UniqueConstraint('submission_id', 'type', name='_submission_type_uc'),)
+    file_generation_task_id = Column(Integer, primary_key=True)
+    generation_task_key = Column(Text, index=True, unique=True)
+    submission_id = Column(Integer, ForeignKey("submission.submission_id", name = "fk_generation_submission"))
```
bsweger: Would it also be useful to track the job_id here (if the D files are still associated with a job...I might have lost the thread on that)?

dataactbroker/handlers/fileHandler.py
```diff
@@ -505,20 +510,89 @@ def startGenerationJob(self, submission_id, file_type):
...
                 jobDb.session.commit()
             except ValueError as e:
                 # Date was not in expected format
                 exc = ResponseException(str(e),StatusCode.CLIENT_ERROR,ValueError)
                 return False, JsonResponse.error(exc, exc.status, url = "", start = "", end = "",  file_type = file_type)
-            get_url = CONFIG_BROKER["".join([file_type_name, "_url"])].format(cgac_code, start_date, end_date)
-            jq.generate_d_file.delay(get_url, user_id, job.job_id, InterfaceHolder, timestamped_name, self.isLocal)
+            # Create file D API URL with dates and callback URL
+            callback = "https://{}:{}/v1/complete_generation/{}/".format(CONFIG_SERVICES["broker_api_host"], CONFIG_SERVICES["broker_api_port"],task_key)
+            get_url = CONFIG_BROKER["".join([file_type_name, "_url"])].format(cgac_code, start_date, end_date, callback)
```
bsweger: This is a small thing, and I can see it both ways, but it's hard to know what's going on here. Due to the combination of `.join` and `format` to construct `get_url`, it took me a while to realize that the `format` operating on parameters in the config file.

Does it make sense for readability to keep the API endpoints in the config file and add the parameters here? I.e., config.yml just has `https://data-usas.uat.workplace.gov/api/contracts`.

Honestly not sure what the best practice is here.

dataactcore/config_example.yml
```diff
@@ -37,12 +37,12 @@ broker:
     # File D1 API
     # Uses "{}" because it requires string formatting to insert query parameters
     # If the URL is provided, the file name provided will be used when uploading to S3
-    award_procurement_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
+    award_procurement_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}&callback={}
```
bsweger: See [earlier comment](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/248/commits/237935d17d58fd54bee733496d14d72df4a8813d#r77886425) about where we should detail the parameter names.

dataactcore/config_example.yml
```diff
@@ -37,12 +37,12 @@ broker:
...
     award_procurement_file_name: d1_data.csv
 
     # File D2 API
     # If the URL is provided, the file name provided will be used when uploading to S3
-    award_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}
+    award_url: https://data-usas.uat.workplace.gov/api/assistance?CGAC={}&date_signed_beginning={}&date_signed_end={}&callback={}
```
bsweger: See [earlier comment](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/248/commits/237935d17d58fd54bee733496d14d72df4a8813d#r77886425) about where we should detail the parameter names.

dataactbroker/handlers/fileHandler.py
```diff
@@ -505,20 +510,89 @@ def startGenerationJob(self, submission_id, file_type):
...
+    def get_lines_from_csv(self, file_path):
+        """ Retrieve all lines from specified CSV file """
+        lines = []
+        with open(file_path) as file:
+            for line in reader(file):
+                lines.append(line)
+        return lines
+
+    def load_d_file(self, url, timestamped_name, job_id, isLocal):
+        """ Pull D file from specified URL and write to S3 """
```
bsweger: Do we need to download the file before writing it to S3 (if I'm reading the code correctly, it doesn't look like we're doing anything with the local copy).

dataactbroker/handlers/fileHandler.py
jworcestBAH: Paging @nmonga91 in case he had an opinion from implementing this, I could go either way here.

dataactbroker/handlers/fileHandler.py
```diff
@@ -648,3 +722,10 @@ def completeGeneration(self, generationId):
         if generationId is None:
             return JsonResponse.error(ResponseException("Must include a generation ID",StatusCode.CLIENT_ERROR), StatusCode.CLIENT_ERROR)
 
+        # Pull url from request
+        safeDictionary = RequestDictionary(self.request)
+        url =  safeDictionary.getValue("href")
+        #Pull information based on task key
+        task = self.interfaces.jobDb.session.query(FileGenerationTask).options(joinedload(FileGenerationTask.file_type)).filter(FileGenerationTask.generation_task_key == generationId).one()
```
bsweger: Curious if there's a specific reason for doing an eager load here when joining these tables (as opposed to specifying a relationship in the model like we do for other foreign keys)?

bsweger:
@jworcestBAH done reviewing what's here now. @nmonga91 and @cmc333333 may want to take another peek.

dataactbroker/handlers/fileHandler.py
```diff
@@ -464,6 +471,7 @@ def startGenerationJob(self, submission_id, file_type):
             Tuple of boolean indicating successful start, and error response if False
 
         """
+        valJob = None
```
cmc333333: Looks like this isn't used?

dataactbroker/handlers/fileHandler.py
```diff
@@ -494,31 +500,146 @@ def startGenerationJob(self, submission_id, file_type):
...
+            # Create File object for this validation job
+            valFile = File(job_id = valJob.job_id, file_status_id = errorDb.getFileStatusId("complete"), filename = valJob.filename, row_errors_present = False)
+            errorDb.session.add(valFile)
+            errorDb.session.commit()
+            valJob.number_of_rows = 0
+            valJob.number_of_rows_valid = 0
+            valJob.file_size = 0
+            valJob.number_of_errors = 0
+            valJob.number_of_warnings = 0
+            valJob.filename = None
```
cmc333333: Do we need to `commit` after modifying the valJob?

dataactbroker/handlers/fileHandler.py
jworcestBAH: We do specify the relationship in the model, but those relationships are not populated unless this type of load is done.  Since the next lines use the file_type.name, this was simpler than doing another lookup to load the file_type name based on file_type_id.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Right, that was from before I pulled the D1/D2 setup into a different function, have now removed this.

dataactbroker/handlers/fileHandler.py
jworcestBAH: Yep, added this

dataactcore/models/jobModels.py
jworcestBAH: Updated to include job ID and simplified the loading to use that

dataactbroker/handlers/fileHandler.py
nmonga91: @bsweger Yes we do. When trying to write the contents of the response (from the file_url) directly to S3, the formatting was not what we wanted. So we download locally, then read from the local file, and write the contents to S3. 

We could try to format `response.text` the way we want, so an array of line, and that would eliminate having to download it locally, potentially.

dataactbroker/handlers/fileHandler.py
nmonga91: I'm indifferent. When I had first put that in there, there wasn't a `.join` happening so it was still concise. 

I'm fine either way. Just thought that not having to specify the parameter names here would be easier but it's the same either way.

dataactvalidator/config/awardFields.csv
```diff
@@ -1,61 +1,61 @@
...
-primaryplaceofperformancecountrycode,place_of_perform_country_c,TRUE,str,3,D1,
-primaryplaceofperformancecountyname,place_of_perform_county_na,FALSE,str,40,,
-primaryplaceofperformanceforeignlocationdescription,place_of_performance_forei,FALSE,str,100,D17,
-primaryplaceofperformancestatename,place_of_perform_state_nam,FALSE,str,50,,
-primaryplaceofperformancezip_4,place_of_performance_zip4a,FALSE,str,10,"D2, D16",
-recordtype,record_type,TRUE,int,1,D1,
-sai_number,sai_number,FALSE,str,50,,
-totalfundingamount,total_funding_amount,FALSE,float,20,,
-uri,uri,FALSE,str,70,"C9, D6, D7",
+actiondate,action_date,FALSE,str,,"D3, D4",
```
bsweger: lazy question w/o looking at the downstream code--what does it mean when the field length is left blank?

dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,14 +1,14 @@
 SELECT afa.row_number, afa.fain, afa.uri
 FROM award_financial_assistance AS afa
 WHERE afa.submission_id = {0}
-	AND afa.federal_action_obligation > 0
+	AND CAST(afa.federal_action_obligation as numeric) > 0
```
bsweger: If `federal_action_obligation` is described as a text data type in the schema and we do, in fact, get text input, this will break. 

Looks like Postgres doesn't have an `ISNUMERIC` function, so we'd have to futz with this. Ultimately, I don't think it's worth holding up the deploy for. I'd suggest:
1. adding a note to the C9 story on the board (I'll do that)
2. checking in the schema team--it doesn't make sense to have a numeric rule on a field that's defined as text.

dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,14 +1,14 @@
...
 	AND afa.row_number NOT IN (
 		SELECT afa.row_number
 		FROM award_financial_assistance AS afa
 			JOIN award_financial AS af
 				ON (afa.fain IS NOT DISTINCT FROM af.fain
 				AND afa.uri IS NOT DISTINCT FROM af.uri
 				AND afa.submission_id = af.submission_id)
-		WHERE afa.federal_action_obligation > 0
+		WHERE CAST(afa.federal_action_obligation as numeric) > 0
```
bsweger: Same comment as [above](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/248/files/667f336c8a9bdce5a52c212fdf20e558df84462f..b55aad54c683d3832d8066c08577270931f9b565#r78033874)

tests/integration/fileTests.py
```diff
@@ -449,9 +449,9 @@ def test_file_generation(self):
 
         self.assertEqual(response.status_code, 200)
         json = response.json
-        self.assertEqual(json["status"], "waiting")
+        self.assertIn(json["status"], ["waiting","finished"])
         self.assertEqual(json["file_type"], "D1")
```
bsweger: A small thing, but the docstring says that this is testing the D2 file generation.

tests/integration/fileTests.py
```diff
@@ -449,9 +449,9 @@ def test_file_generation(self):
...
-        self.assertIn("d1_data.csv", json["url"])
+        self.assertEqual("#", json["url"])
```
bsweger: Are we checking this because we're not expecting a real d1 file back in the integration tests?

bsweger:
Thanks, @jworcestBAH, for all the work you've done to get this down the home stretch. Done reviewing the additional changes. Some questions inline, but two more:

* Am I right in thinking that the .raw award file is a cut and paste from the latest schema docs (_i.e._, I didn't individually check each field against the IDD, but I can if that would be helpful)
* Saw some chatter about this on Slack, but is there a reason we no longer have the file d unit tests?

dataactvalidator/config/awardFields.csv
jworcestBAH: A blank field length means that no length check is applied

dataactvalidator/config/sqlrules/c9_award_financial.sql
jworcestBAH: So part of the motivation here is that the break will occur during the cross-file validation where there is error reporting to the user, rather than happening on D2 file validation which just causes a loop back to the beginning of the process without reporting anything to the user.  That said, the Smartronix system is supposed to be validating D2 for this type of thing, but if they do send data of the wrong type this will at least have the error occur on a step that is visible to the frontend.

tests/integration/fileTests.py
jworcestBAH: Fixed comments here

tests/integration/fileTests.py
jworcestBAH: The agency here is "SYS", so we expect the numFound = 0 handling to occur.

dataactvalidator/config/sqlrules/c9_award_financial.sql
bsweger: That makes sense. However, if we CAST a string to numeric, that will throw a SQL error, which presumably won't show up in a nice way on the error report. Again, I'm fine leaving this but think it merits more discussion. Will add to the question list.

tests/integration/fileTests.py
bsweger: Ah, thanks for clarifying that. Do we have some kind of test for the path where an actual D file comes back?

jworcestBAH:
The awardRaw file is originally from the IDD, but I've modified it here based on Kaitlin's request that we skip basic D file validations given that we have no way of reporting them on the frontend and they're supposed to be validated by Smartronix.

For the file D unit tests, the functions that were being tested no longer exist, so I removed the unit tests on those functions.  There is an integration test on the generate_file and check_generation_status routes.

bsweger:
@jworcestBAH thanks for the clarifications.

If we're skipping the individual file D validations entirely, we might want consider a future change that would bypass the file-level validations rather than having to make manual changes to the IDD, which will be hard to replicate when that changes. I say this, of course, not having the full context of what we're skipping.

My concern about the tests is that we no longer have any kind of coverage for the happy path where we get D files back from the API. Am I reading that right? @nmonga91 can you weigh in on this?

bsweger:
Going to merge this. Folks are working on re-instating a unit test in parallel to testing this set of changes on dev.

PR #249, opened 2016-09-07
Add logging

cmc333333:
~~@bsweger's GlobalDB commit was lost in the shuffle.~~ Also adds basic, configurable logging


bsweger:
@cmc333333 @jworcestBAH Sorry for the confusion around this, but that branch I made based on CM's db tweaks contains things intended for the current sprint weren't meant to deploy right now.

cmc333333:
Understood, @bsweger -- force-pushed to remove that changeset

bsweger:
@cmc333333 Thanks for setting this up. At your suggestion, I had swapped out some print statements with the logger in the domain values load script. Is the intent here to give a global means of using the Python logger instead of print statements? How would this work in conjunction with what we're already logging to logstash?

cmc333333:
I'm going to close this in favor of DB-1110; I think the `CloudLogger` is _so_ close to being a logging handler. Let's combine that with this code, soon.

PR #250, opened 2016-09-08
Set up file F endpoints

cmc333333:
This builds on #248([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1097-generation-callback...db-1039-connect-f-endpoints-take2)) to avoid merge conflicts and obviates #238.

While the generating the file from the UI "works" in the sense that it's generating a CSV with the appropriate headers, local testing only produces empty files. It's very possible that we're using the wrong field to do lookups. In any event, there's no harm merging this before that gets fixed.


cmc333333:
Brief update: looks like I hadn't imported enough matches yet. Now I'm seeing the worker die as it tries to process 398454 subawards from ~350 awards. Something's up, so let's not merge this just yet.

cmc333333:
And now that should all be fixed -- I was cross-joining. In any event, I'm more confident that this won't break anything now :)

cmc333333:
Closing in favor of a rebased version

PR #251, opened 2016-09-08
Db 1097 callback logging

nmonga91:
Updated CloudLogger to be more dynamic and added logging for the SMX callback


bsweger:
Thanks for the additional logging. Looks good--will merge when Jenkins is done.

PR #254, opened 2016-09-09
Use debug setting in config.yml when creating Flask apps

bsweger:
Setting the debug on app.run wasn't working. Do it when we create
the app. To turn off debug, set server_debug = false in config.yml.


jworcestBAH:
Looks good, merging

PR #257, opened 2016-09-09
Prepending https rather than http if port is 443

jworcestBAH:
Fix for D1/D2 callbacks on staging

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -547,7 +547,12 @@ class FileHandler:
             exc = ResponseException(str(e),StatusCode.CLIENT_ERROR,ValueError)
             return False, JsonResponse.error(exc, exc.status, url = "", start = "", end = "",  file_type = file_type)
         # Create file D API URL with dates and callback URL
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -547,7 +547,12 @@ def addJobInfoForDFile(self, upload_file_name, timestamped_name, submission_id,
             exc = ResponseException(str(e),StatusCode.CLIENT_ERROR,ValueError)
             return False, JsonResponse.error(exc, exc.status, url = "", start = "", end = "",  file_type = file_type)
         # Create file D API URL with dates and callback URL
-        callback = "http://{}:{}/v1/complete_generation/{}/".format(CONFIG_SERVICES["broker_api_host"], CONFIG_SERVICES["broker_api_port"],task_key)
+        if CONFIG_SERVICES["broker_api_port"] == 443:
+            # Use https
+            prefix = "https"
```
nmonga91: It's super nit-picky, but if we can just call it `protocol` instead of `prefix` it'd be very clear (even though people can see the value and determine that) but prefix is also used in other places, for example when doing stuff with S3.

jworcestBAH: Fixed

nmonga91:
Looks good, merging!

PR #258, opened 2016-09-09
Prepending https rather than http if port is 443

jworcestBAH:
Same for development


jworcestBAH:
Closing to wait for dev-stable

PR #263, opened 2016-09-09
Adding log messages to job queueing

jworcestBAH:
This reports when a job is sent to the job manager, when it arrives, and when the response from the validator comes back.

### DIFF
--- a/dataactcore/utils/jobQueue.py
+++ b/dataactcore/utils/jobQueue.py
@@ -36,6 +36,7 @@ class JobQueue:
             }
             response = requests.post(url, params)
             CloudLogger.log("Job {} has completed validation".format(str(jobID)))
+            CloudLogger.log("Validator response: {}".format(str(response.json())))
             return response.json()
 
         self.enqueue = enqueue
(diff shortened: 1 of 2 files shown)


klrBAH:
[run tests]

klrBAH:
[run tests]

klrBAH:
[run tests]

klrBAH:
[run tests]

dataactcore/utils/jobQueue.py
```diff
@@ -28,11 +29,13 @@ def __init__(self, job_queue_url="localhost"):
         @self.jobQueue.task(name='jobQueue.enqueue')
         def enqueue(jobID):
             # Don't need to worry about the response currently
+            CloudLogger.log("Adding job {} to the queue".format(str(jobID)))
```
bsweger: I'm not fully up to speed on the job manager, so asking as a general question--is there any other info that would be useful to log here? (`queueURL`, etc.)

Paging @nmonga91, who would have better insight into this.

bsweger:
@jworcestBAH done reviewing. Looks good. Since it takes a while to deploy to staging, would love @nmonga91's eyes to make sure we're logging every single piece of info that could help us track down the bottleneck.

dataactcore/utils/jobQueue.py
nmonga91: We should also output the response json on line 39 below. That would provide us with a more in-depth trace if any errors occur. The `queueURL` wouldn't be too helpful as it was just refer to itself (localhost). Validator URL may be helpful but that can just be derived from the config.

jworcestBAH: Added a log of response.json()

bsweger:
Looks good--merging!

PR #264, opened 2016-09-12
Dev stable

jworcestBAH:
Moving extra logging up to staging


bsweger:
Looks good--will merge when Jenkins is done!

PR #265, opened 2016-09-12
Dev stable

jworcestBAH:
And moving extra logging down to dev


bsweger:
Looks good--will merge when Jenkins is done!

PR #266, opened 2016-09-12
SQL Validations for C20 and C21

mab6bf:


mab6bf:
[run tests]

bsweger:
@mab6bf A question while I'm digging into these queries...will these new rules work with our current set of test files? We'll need to confirm we have working test files before merging to dev.

tests/unit/dataactvalidator/test_c21_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+
+    errors = number_of_errors(_FILE, database.stagingDb, models=[af1, af2, op])
+    assert errors == 0
+
+
+def test_failure(database):
+    """ Tests that the sum of financial elements in File C is not less than or equal
+         to the corresponding element in File B for the same TAS and Program Activity Code combination"""
+    submission_id = randint(1000, 10000)
```
bsweger: Another good place to use the dictionary pattern suggested in the C20 `test_success` comments.

tests/unit/dataactvalidator/test_c21_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+        ussgl497100_downward_adjus_cpe=ussgl497100_downward_adjus_cpe,
+        ussgl487200_downward_adjus_cpe=ussgl487200_downward_adjus_cpe,
+        ussgl497200_downward_adjus_cpe=ussgl497200_downward_adjus_cpe,
+        deobligations_recov_by_awa_cpe=deobligations_recov_by_awa_cpe,
+        tas=tas,
+        program_activity_code=program_activity_code,
+        submission_id=submission_id
+    )
+
+    op = ObjectClassProgramActivityFactory(
```
bsweger: Same question as the one I had for the `test_failure` method in the C20 unit test--should we make sure the `<` part of the rule is tested along with the `=` part.

I think the broader question, however, is: if any one of the comparisons fails, the entire rule fails. Is this really what we want (will say more in the PR comments).

tests/unit/dataactvalidator/test_c20_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number'}
+    actual = set(query_columns(_FILE, database.stagingDb))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Tests that the sum of financial elements in File C is less than or equal
+     to the corresponding element in File B for the same TAS and Object Class combination"""
+    submission_id = randint(1000, 10000)
```
mab6bf: Done

tests/unit/dataactvalidator/test_c20_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+
+    errors = number_of_errors(_FILE, database.stagingDb, models=[af1, af2, op])
+    assert errors == 0
+
+
+def test_failure(database):
+    """ Tests that the sum of financial elements in File C is not less than or equal
+         to the corresponding element in File B for the same TAS and Object Class combination"""
+    submission_id = randint(1000, 10000)
```
mab6bf: Done

tests/unit/dataactvalidator/test_c21_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number'}
+    actual = set(query_columns(_FILE, database.stagingDb))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Tests that the sum of financial elements in File C is less than or equal
+     to the corresponding element in File B for the same TAS and Program Activity Code combination"""
+    submission_id = randint(1000, 10000)
```
mab6bf: Done

tests/unit/dataactvalidator/test_c21_award_financial.py
mab6bf: Done

dataactvalidator/config/sqlrules/c20_award_financial.sql
```diff
@@ -0,0 +1,178 @@
...
+		SUM(deobligations_recov_by_awa_cpe) AS deobligations_recov_by_awa_cpe_sum,
+		tas,
+		object_class,
+		submission_id
+	FROM award_financial
+	WHERE submission_id = {0}
+	GROUP BY tas, object_class, submission_id
+	)
+SELECT
+	op.row_number
```
mab6bf: Fixed

dataactvalidator/config/sqlrules/c21_award_financial.sql
```diff
@@ -0,0 +1,178 @@
...
+		SUM(deobligations_recov_by_awa_cpe) AS deobligations_recov_by_awa_cpe_sum,
+		tas,
+		program_activity_code,
+		submission_id
+	FROM award_financial
+	WHERE submission_id = {0}
+	GROUP BY tas, program_activity_code, submission_id
+	)
+SELECT
+	op.row_number
```
mab6bf: Fixed

tests/unit/dataactvalidator/test_c20_award_financial.py
```diff
@@ -0,0 +1,327 @@
...
+        ussgl497100_downward_adjus_cpe=ussgl497100_downward_adjus_cpe,
+        ussgl487200_downward_adjus_cpe=ussgl487200_downward_adjus_cpe,
+        ussgl497200_downward_adjus_cpe=ussgl497200_downward_adjus_cpe,
+        deobligations_recov_by_awa_cpe=deobligations_recov_by_awa_cpe,
+        tas=tas,
+        object_class=object_class,
+        submission_id=submission_id
+    )
+
+    op = ObjectClassProgramActivityFactory(
```
mab6bf: Fixed

tests/unit/dataactvalidator/test_c21_award_financial.py
mab6bf: Fixed

dataactvalidator/config/sqlrules/c20_award_financial.sql
```diff
@@ -1,118 +1,77 @@
...
+	SUM(af.gross_outlays_undelivered_cpe) AS gross_outlays_undelivered_cpe_sum,
+	SUM(af.ussgl490200_delivered_orde_cpe) AS ussgl490200_delivered_orde_cpe_sum,
+	SUM(af.ussgl490800_authority_outl_fyb) AS ussgl490800_authority_outl_fyb_sum,
+	SUM(af.ussgl490800_authority_outl_cpe) AS ussgl490800_authority_outl_cpe_sum,
+	SUM(af.ussgl498200_upward_adjustm_cpe) AS ussgl498200_upward_adjustm_cpe_sum,
+	SUM(af.gross_outlays_delivered_or_fyb) AS gross_outlays_delivered_or_fyb_sum,
+	SUM(af.gross_outlays_delivered_or_cpe) AS gross_outlays_delivered_or_cpe_sum,
+	SUM(af.gross_outlay_amount_by_awa_fyb) AS gross_outlay_amount_by_awa_fyb_sum,
+	SUM(af.gross_outlay_amount_by_awa_cpe) AS gross_outlay_amount_by_awa_cpe_sum,
+	SUM(af.obligations_incurred_byawa_cpe) AS obligations_incurred_byawa_cpe_sum,
```
bsweger: Tiny thing--missing `-` between `by` and `awa`

bsweger: Disregard--`byawa` is actually the name in the db.

klrBAH:
[run tests]

bsweger:
We had some back-and-forth in our Slack channel...updating here for transparency.

At a high level, we expect rule C totals to be <= to their corresponding file B totals. Given this expectation, we're normally write a SQL rule to look for file C totals > file B, so we can return them as errors. 

However, because these values in Files B and C have been coming in as negative, we've reversed the comparator...in other words, we're looking for file C totals < file B totals and returning them as errors.

It seems that we can't always assume the both values are going to be negative. We also need to handle the case of mixed signs and two positive values. The schema team is taking a closer look at this, and we need some clarity there before merging this.

For example, there might need to be an additional SQL statement in these rules that checks for mixed signs between the B and C files.

bsweger:
For posterity, updating with the outcome of a call between the broker and schema teams. The schema folks confirmed that in the vast majority of cases, we can assume that both file B and file C will have negative values. Thus, the rule will work as written.

They also agreed that the amount of uncertainty around the C20 and C21 rules warrants moving them from a fatal error to a warning. So, with that change, I think we're good to go.

The schema team is also considering a re-word of the rules that will make it easier to draw a line between them and their technical implementation

PR #269, opened 2016-09-12
Increased to 7 uwsgi workers

jworcestBAH:
Uwsgi worker change to dev


bsweger:
@jworcestBAH should this PR be from `dev-stable` to `dev` (instead of from the hotfix branch)?

jworcestBAH:
Ah, right, I'll recreate it

PR #272, opened 2016-09-13
database connection refactor Part 1

bsweger:
This is the first part of [DB-1029](https://federal-spending-transparency.atlassian.net/browse/DB-1029), which involves a refactor of the way we handle database connections in the data broker.

Currently, we have a system of interface objects for each "database" (an artifact of the time when the broker's backend involved multiple databases). There are also interface "holders" that hold instantiated interfaces to pass around. We've run into a few problems with this method:

* complexity
* circular references
* connections not always being open/available when expected

As discussed, we're going to re-tool our approach with these high-level principles:

1. Database access within a Flask request will use db engine/session/connection info stashed in the application context. This db information is stored and torn down at the beginning and end of each Flask request.
2. Database access outside of Flask (e.g., database setup, integration test setup) will use a Python [context manager](https://www.python.org/dev/peps/pep-0343/) to get and tear down a a new db session/connection/engine.

This PR addresses number two on the above list. The idea is that when this is merged, we'll be able to focus on simplifying the database interfaces used by the Flask requests. Until that work is done, we'll have a few things running in parallel (for example, the session handling that CM added to BaseInterface). At the end of story, we'll be able to clean up the unused bits.

Taking a page from CM's playbook, I tried to chunk the commits into logical units of work with extended commit messages, so it might be easier to review by stepping through commit by commit.

- [x] PR comes out of WIP status
- [ ] Code review


dataactvalidator/scripts/loaderUtils.py
```diff
@@ -13,11 +13,7 @@ def padFunction(self, field, padTo, keepNull):
         """Pads field to specified length."""
         if isnull(field) or not str(field).strip():
             if keepNull:
```
nmonga91: Do we still need `keepNull` as a parameter? It looks like that option was removed above. Not sure if other places still rely on it.

bsweger: Yep, we need it. You're right that it's gone from the SF-133 load function, but it's still used in `loadProgramActivity.`

klrBAH:
[run tests\

klrBAH:
[run tests]

nmonga91:
Looks good, but will do one additional pass through a little later to ensure nothing was overlooked since there were a lot of updates.

nmonga91:
[run tests]

klrBAH:
[run tests]

nmonga91:
[run tests]

dataactcore/README.md
```diff
@@ -66,13 +66,6 @@ Database interfaces are defined for logical data functions in the data broker (e
 
 The `scripts/` folder contains various python scripts used in the DATA Act broker backend install process as well as various utility scripts.
 
-To clean out the databases for testing proposes, use the following scripts:
```
nmonga91: Kudos for remembering to update the documentation as well!

nmonga91:
Looks good, merging!

PR #283, opened 2016-09-19
Db 1177 d file exception

nmonga91:
When an exception occurs, the generation job is marked as "failed". Also added a Response exception to be raised in case the call to the SMX api times out.


dataactbroker/handlers/fileHandler.py
```diff
@@ -557,11 +557,17 @@ def addJobInfoForDFile(self, upload_file_name, timestamped_name, submission_id,
...
-        if not self.call_d_file_api(get_url):
-            self.handleEmptyResponse(job, valJob)
+        try:
+            CloudLogger.log("DEBUG: Calling D file API => " + str(get_url),
+                            log_type="debug",
+                            file_name=self.debug_file_name)
+            if not self.call_d_file_api(get_url):
+                self.handleEmptyResponse(job, valJob)
+        except Timeout as e:
+            exc = ResponseException(str(e), StatusCode.CLIENT_ERROR, Timeout)
```
bsweger: I didn't dig into `ResponseException`...will that make sure we log the timeout along with everything else we're logging about the D file process?

dataactbroker/handlers/fileHandler.py
```diff
@@ -557,11 +557,17 @@ def addJobInfoForDFile(self, upload_file_name, timestamped_name, submission_id,
             protocol = "http"
         callback = "{}://{}:{}/v1/complete_generation/{}/".format(protocol,CONFIG_SERVICES["broker_api_host"], CONFIG_SERVICES["broker_api_port"],task_key)
         get_url = CONFIG_BROKER["".join([file_type_name, "_url"])].format(cgac_code, start_date, end_date, callback)
-        CloudLogger.log("DEBUG: Calling D file API => " + str(get_url),
-                        log_type="debug",
-                        file_name=self.debug_file_name)
-        if not self.call_d_file_api(get_url):
-            self.handleEmptyResponse(job, valJob)
+        try:
+            CloudLogger.log("DEBUG: Calling D file API => " + str(get_url),
```
bsweger: Can the logging call come before the `try` block just to remove any ambiguity about what is causing an exception?

dataactbroker/handlers/fileHandler.py
nmonga91: Yup! that is correct. it's still logged, but instead wrapped in the ResponseException. I saw this happening in other places, but now that I'm looking back at it, I think it shouldn't be wrapped so the exact cause is known without any confusion.

dataactbroker/handlers/fileHandler.py
nmonga91: Done

dataactbroker/handlers/fileHandler.py
nmonga91: Based on my conversation with @jworcestBAH, we should be ok leaving it this way for consistency since it's also logged and is still traceable.

nmonga91:
@bsweger about your comment about the indirectly related request, I think that this story is a good opportunity to get that done. I will incorporate it.

bsweger:
@nmonga91 Thanks for being willing to work that in!

bsweger:
Looks good--thanks, @nmonga91...merging!

PR #284, opened 2016-09-19
Run SQL validation rules load outside of global database context.

bsweger:
Forget this validation setup script in the changes for PR #272. This is another piece of the work to use the session context for db access that occurs outside of a Flask request.


jworcestBAH:
Okay, changes look good, merging.

PR #285, opened 2016-09-19
Hotfix cross file status

jworcestBAH:
This adds a File object for cross-file jobs to allow the check_status route to report cross-file errors and warnings correctly

### DIFF
--- a/dataactvalidator/interfaces/validatorErrorInterface.py
+++ b/dataactvalidator/interfaces/validatorErrorInterface.py
@@ -26,7 +26,7 @@ class ValidatorErrorInterface(ErrorInterface):
         self.session.commit()
         return fileRec
 
-    def createFileIfNeeded(self, jobId, filename = ""):
+    def createFileIfNeeded(self, jobId, filename = None):
         """ Return the existing file object if it exists, or create a new one """
         try:
             fileRec = self.getFileByJobId(jobId)
@@ -73,7 +73,7 @@ class ValidatorErrorInterface(ErrorInterface):
         self.session.commit()
         return True
 
-    def markFileComplete(self, jobId, filename):
+    def markFileComplete(self, jobId, filename = None):
         """ Marks file's status as complete
 
         Args:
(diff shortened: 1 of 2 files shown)


dataactvalidator/interfaces/validatorErrorInterface.py
```diff
@@ -26,7 +26,7 @@ def createFile(self, jobId, filename):
         self.session.commit()
         return fileRec
 
-    def createFileIfNeeded(self, jobId, filename):
+    def createFileIfNeeded(self, jobId, filename = ""):
```
bsweger: Since we're trying to handle the case of no applicable filename, does it make sense to default this parameter to `None` instead of a blank string (to ensure that it inserts to the database as `NULL`)?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -533,6 +536,9 @@ def runCrossValidation(self, jobId, interfaces):
         if interfaces.jobDb.getSubmissionById(submissionId).number_of_errors == 0:
             interfaces.jobDb.setPublishableFlag(submissionId, True)
 
+        # Mark validation complete
+        interfaces.errorDb.markFileComplete(jobId, "")
```
bsweger: Should the `markFileComplete` method be updated to use a default for file name in the same way that `createFileIfNeeded` does? (see the question on that one re: `None` vs a blank string)?

dataactvalidator/interfaces/validatorErrorInterface.py
jworcestBAH: Good point, changed this to None

dataactvalidator/validation_handlers/validationManager.py
jworcestBAH: Yep, changed this to also use None as a default

bsweger:
@jworcestBAH thanks for those changes...looks good..merging!

PR #286, opened 2016-09-19
db connection refactor part 2: unit tests

bsweger:
This PR mainly updates the unit tests to get a database fixture using the functions in `dataactcore/interfaces/db.py` instead of baseInterface + interface holder.

A few miscellaneous items included:
* Update the FSRS loader (another one I forgot in PR #272)
* Migrations now use the common URI function in db.py instead of repeating the code to create a URI string

@mab6bf @jworcestBAH @nmonga91 @cmc333333: heads-up that once this is merged, the syntax for referencing the pytest database fixture will change slightly.

Old: `database.stagingDb`
New: `database`

### DIFF
--- a/dataactbroker/scripts/loadFSRS.py
+++ b/dataactbroker/scripts/loadFSRS.py
@@ -1,7 +1,7 @@
 import logging
 import sys
 
-from dataactcore.models.baseInterface import databaseSession
+from dataactcore.interfaces.db import databaseSession
 from dataactbroker.fsrs import (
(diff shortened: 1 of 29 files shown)


cmc333333:
Makes a ton of sense; the only issue I'm imagining is if we need access to other methods of `InterfaceHolder` or it's subfields. That said, if we're moving away from that style (yay!) this'll be a nice forcing function.

bsweger:
Thanks, @cmc333333. Yes, the plan is to move away from that style!

PR #287, opened 2016-09-20
File F endpoints

cmc333333:
I'll apologize up front for the size of this PR -- the changesets have been built over several weeks. I can definitely split this up into smaller PRs if reviewing is a problem. @bsweger as a lot of this was written before your db reworkings, please keep an eye out to flag any issues specifically around that.

This
* Generates file F data from the UI
* Fixes errors with FSRS XML
* Adds an database index to speed up queries


tests/unit/conftest.py
```diff
@@ -25,3 +26,23 @@ def database():
 
     db.close()
     dropDatabase(config['db_name'])
+
+
+@pytest.fixture()
```
bsweger: This is neat--thank you for modeling the power of fixtures.

dataactcore/utils/jobQueue.py
```diff
@@ -22,6 +30,32 @@ def enqueue(jobID):
...
+        try:
+            rows_of_dicts = fileF.generateFRows(job_manager.session,
+                                                submission_id)
+            header = [key for key in fileF.mappings]    # keep order
+            body = []
+            for row in rows_of_dicts:
+                body.append([row[key] for key in header])
+
+            write_csv(timestamped_name, is_local, header, body)
+            job_manager.markJobStatus(job_id, "finished")
```
bsweger: This is a good opportunity to think about the impending refactor of db access for code running as part of a Flask request.

The first pass we discussed was to take all the methods in the interface classes (_i.e._, `job_manager` in this case) and move them to one big file in `dataactcore.interfaces`. That will be an ugly-looking file, but will provide a starting point to putting things in better places (models, etc. @nmonga91 had some good ideas about this).

So in this case, you'd end up calling `markJobStatus(job_id, "finished")` without the need to get a `job_manager` first. 

Alternately, for these short functions, we should ask ourselves--do we really need to abstract to a function that is (or could be) 1-2 lines of code? Are there ways to tweak the models themselves that would support this effort?

If we opt for the latter, you could get ahead of the game by getting rid of the `markJobStatus` call and doing that inline.

Paging @nmonga91 and @jworcestBAH on this one!

dataactcore/utils/jobQueue.py
```diff
@@ -37,6 +71,8 @@ def __init__(self, job_queue_url="localhost"):
         self.jobQueue.config_from_object('celeryconfig')
 
         self.enqueue = self.jobQueue.task(name='jobQueue.enqueue')(enqueue)
+        self.generate_f_file = self.jobQueue.task(
```
bsweger: I don't understand the job queue very well and wasn't the D file reviewer, so forgive the question...what does this line of code do? Is the `generate_f_file` at the end some kind of callback?

dataactbroker/fsrs.py
```diff
@@ -30,6 +32,31 @@ def configValid():
     return bool(procWSDL) and bool(grantWSDL)
 
 
+class ControlFilter(MessagePlugin):
```
bsweger: Checking my understanding of how this works...this is a SUDS `MessagePlugin`, so it will intercept the incoming data and swap out the control spaces as directed, so long as the class is added as a plugin? This is what I surmise from reading the SUDS docs, but those are...sparse.

dataactcore/migrations/versions/51679e7bfe01_merge_dsrs_index_with_master.py
```diff
@@ -0,0 +1,36 @@
...
+
+Revision ID: 51679e7bfe01
+Revises: d074fd492181, dcbb5afa125e
+Create Date: 2016-09-20 14:46:56.470399
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '51679e7bfe01'
+down_revision = ('d074fd492181', 'dcbb5afa125e')
```
bsweger: oooh, cool--I've been handling alembic conflicts manually...good to know about this.

dataactcore/utils/jobQueue.py
cmc333333: Gotcha. I'll take a quick pass as option 2 (removing the `markJobStatus` function altogether) and if that fails, take the bucket-of-functions approach.

jworcestBAH: Putting these functions in the model classes seems like a good idea.  In the meantime, is there a benefit to moving them all into one file?  That seems like it would make it more difficult to find these functions, in the current organization they're at least split up by category of what they affect.  Maybe making them all static functions in the present locations could be a good first step, that way you wouldn't need an instance before calling the function.  Possibly some of the smaller functions could be eliminated and replaced with direct queries, however markJobStatus includes checking for jobs that have been marked as 'finished' and starting any dependent jobs that are ready.  We had talked about trying to have Celery manage dependencies, so if we get that working we could remove this function after that point.  Removing it prior to that would prevent validation and cross-file validation jobs from starting.

dataactcore/utils/jobQueue.py
cmc333333: It's a good question, and I'd argue it's rather messy right now. The `JobQueue` class creates a `Celery` connection in its constructor and then sets some convenience functions for using the two asynchronous functions (`enqueue` and now `generate_f_file`).  Celery [suggests](http://docs.celeryproject.org/en/latest/getting-started/first-steps-with-celery.html#application) tasks be set up like:

```python
app = Celery(...)

@app.task
def my_function():
   ...
```
I'm not too sure why we're currently setting all that up in a constructor. Maybe we needed to delay to run time for some reason?

This PR takes one step to simplifying that by moving the actual logic of those functions out of the initializer, but we can probably go further. I'll take a stab.

dataactbroker/fsrs.py
cmc333333: You got it! I'll add a better docstring. Definitely agreed on poor SUDS docs.

dataactcore/migrations/versions/51679e7bfe01_merge_dsrs_index_with_master.py
cmc333333: Sure thing! It's very [analogous](http://alembic.zzzcomputing.com/en/latest/branches.html#merging-branches) to git.

dataactcore/utils/jobQueue.py
```diff
@@ -1,46 +1,80 @@
...
-            url = ''.join([validatorUrl, '/validate/'])
-            params = {
-                'job_id': jobID
-            }
-            response = requests.post(url, params)
-            CloudLogger.log("Job {} has completed validation".format(str(jobID)))
-            CloudLogger.log("Validator response: {}".format(str(response.json())))
-            return response.json()
-
-        self.enqueue = enqueue
```
jworcestBAH: This may require a change to the last line of dataactcore.models.jobTrackerInterface.checkJobDependencies that currently queues validation and cross-file validation jobs that are ready to be started, could you check if that line is consistent with the new way of queueing?  I think the way it's written now is dependent on enqueue being part of JobQueue.

dataactcore/utils/jobQueue.py
jworcestBAH: Also, it appears that the usage of the interface holder above isn't consistent with the recent database changes, it may be a good idea to merge dev into this branch to make sure things still work before completing the PR?

dataactcore/utils/jobQueue.py
cmc333333: So far, there's no interface change, so we should be fine. Thanks for the heads up, though; I'll be sure to update callers when testing out the `JobQueue`-less approach

dataactcore/utils/jobQueue.py
cmc333333: Dev is merged in and things are working locally -- what doesn't look consistent?

bsweger: @jworcestBAH Thanks for weighing in here... all good ideas and concerns.

Re: changing everything to static functions in their present locations...my .02 is that the current organization makes the functions hard to find because they are scattered between the broker, core, and validator. I did toss around the idea of creating a single file of static functions for each of the old db names and putting those in `core/interfaces`. 

When I started playing with that in practice, I realized that it doesn't solve the circular reference issue, because the separate files would all still have to import each other. I also come back to that fact that these names (_e.g._, error, job, etc.) aren't meaningful now that we've consolidated the databases, so I don't see an advantage to keeping that distinction in the code base.

I *really* agree with you that a big old file with a bunch of database functions is not a good final goal. I see it as a starting point--a common place to hold everything as we get it the individual pieces moved out to the models, replaced with direct queries, celery dependencies, etc.

Thanks for pointing out the additional logic embedded in `markJobStatus`. That's something I missed when scanning the code.

cmc333333: After moving about 250 lines out of JobTrackerInterface, I hit a dependency on JobQueue that I don't think is worth trying to resolve. I'm going to punt and say that conversion is outside the scope of this PR.

bsweger: 👍 We'll handle this in the broader refactor--thanks for taking a stab at it!

dataactcore/models/jobTrackerInterface.py
```diff
@@ -6,18 +6,12 @@
...
+from dataactcore.utils.jobQueue import enqueue
 from dataactvalidator.validation_handlers.validationError import ValidationError
 
 
 class JobTrackerInterface(BaseInterface):
     """Manages all interaction with the job tracker database."""
-
-    def __init__(self):
-        self.jobQueue = JobQueue(job_queue_url=CONFIG_JOB_QUEUE['url'])
-        super(JobTrackerInterface, self).__init__()
```
bsweger: Do we lose anything crucial by no longer invoking the `__init__` of the parent class here (_i.e._, `BaseInterface`). I'm asking because I see that `BaseInterface` sets `self.dbName`, which is maaaybe used in the `baseTestAPI.py` and `baseTestValidator.py` teardown functions. 

That said, everything seems to be working here, test-wise, so perhaps I got lost in the interfaces.

cmc333333: This constructor was setting up `self.jobQueue` and then calling it's parent class' constructor. By removing the `__init__` override, we'll continue to call the parent class' constructor (due to class inheritance) -- we're just not doing the "special" thing of setting `self.jobQueue`. Does that make sense?

dataactbroker/fsrs.py
```diff
@@ -35,7 +35,8 @@ def configValid():
 class ControlFilter(MessagePlugin):
     """Suds (apparently) doesn't know how to decode certain control characters
     like ^V (synchronous idle) and ^A (start of heading). As we don't really
-    care about these characters, swap them out for spaces"""
+    care about these characters, swap them out for spaces. MessagePlugins are
+    Suds's mechanism to transform SOAP content before it gets parsed."""
```
bsweger: Thanks for adding this to make things easier for the casual code reader!

bsweger:
Looks good--merging!

PR #288, opened 2016-09-20
Db 1102 rule b18

jworcestBAH:
Adds B18 along with unit tests


jworcestBAH:
Note that this includes the file B factory, if a separate PR is created for that we should merge that in from dev into this branch first.

dataactvalidator/config/sqlrules/b18_object_class_program_activity_2.sql
```diff
@@ -0,0 +1,9 @@
+SELECT
+    op.row_number,
+    op.object_class,
+    op.by_direct_reimbursable_fun
+FROM object_class_program_activity AS op
+WHERE op.submission_id = {} AND length(op.object_class) = 4 AND NOT COALESCE(op.by_direct_reimbursable_fun,'') = ''
+AND NOT (SUBSTRING(op.object_class,1,1) = '1'
+AND lower(op.by_direct_reimbursable_fun) = 'd') AND NOT (SUBSTRING(op.object_class,1,1) = '2'
+AND lower(op.by_direct_reimbursable_fun) = 'r')
```
cmc333333: Does this query need some `OR`s? I'm not sure how this is working now -- I don't think any record would have `lower(op.by_direct_reimbursable_fun)` be both `d` _and_ `r`? Maybe `NOT` is binding in an unexpected way? Related: Is `<>` preferable?

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
```diff
@@ -0,0 +1,45 @@
+from factory import fuzzy
+from tests.unit.dataactcore.factories.staging import ObjectClassProgramActivityFactory
+from tests.unit.dataactvalidator.utils import number_of_errors, query_columns
+
+
+_FILE = 'b18_object_class_program_activity_1'
+
+not_req_dict = {
+        "submission_id":fuzzy.FuzzyInteger(9999),
+        "object_class":fuzzy.FuzzyInteger(1000,9999),
```
cmc333333: I'm very confused by the use of `fuzzy.FuzzyInteger` here. Are you wanting them to be random values? If so, you probably want to use `randint` instead; `FuzzyInteger` is a class. However, if you're looking for random values, why specify them at all? FactoryBoy'll generate random values for you.

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
```diff
@@ -0,0 +1,45 @@
...
+req_absent_dict = {
+        "submission_id":fuzzy.FuzzyInteger(9999),
+        "object_class":fuzzy.FuzzyInteger(999),
+        "by_direct_reimbursable_fun":""
+}
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "object_class", "by_direct_reimbursable_fun"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
```
cmc333333: Curious, why `<=` rather than `==` here?

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_success(database):
+    """ Test that a four digit object class with no flag is a success, and a three digit object class with a flag is a success"""
+    not_req_ocpa = ObjectClassProgramActivityFactory(**not_req_dict)
+    req_present_ocpa = ObjectClassProgramActivityFactory(**req_present_dict)
```
cmc333333: This may just be a style thing, but one of the goals of factory methods is that folks can read the tests with minimal distractions/lookups. Why not remove the module-level variables and spell everything out here? I think you can do:
```python
not_req_ocpa = ObjectClassProgramActivityFactory(
    by_direct_reimbursable_fun='', object_class=str(randint(1000, 10000)))
req_present_ocpa = ObjectClassProgramActivityFactory(
    by_direct_reimbursable_fun='d', object_class=str(randint(100, 1000)))
```

tests/unit/dataactvalidator/test_b18_object_class_program_activity_2.py
```diff
@@ -0,0 +1,66 @@
...
+def test_column_headers(database):
+    expected_subset = {"row_number", "object_class", "by_direct_reimbursable_fun"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
+
+def test_success(database):
+    """ Test that a four digit object class with no flag is a success, and a three digit object class with a flag is a success"""
+    ocpa_list = []
+    for value_dict in [skip_three_digit, skip_blank_flag, match_d_value, match_r_value]:
+        ocpa_list.append(ObjectClassProgramActivityFactory(**value_dict))
```
cmc333333: NBD, but if these are separate success cases, it might make sense to create separate tests for each.

tests/unit/dataactvalidator/test_b18_object_class_program_activity_2.py
```diff
@@ -0,0 +1,66 @@
...
+
+skip_three_digit = {
+        "submission_id":fuzzy.FuzzyInteger(9999),
+        "object_class":fuzzy.FuzzyInteger(999),
+        "by_direct_reimbursable_fun":"d"
+}
+
+skip_blank_flag = {
+        "submission_id":fuzzy.FuzzyInteger(9999),
+        "object_class":fuzzy.FuzzyInteger(9999),
```
cmc333333: We want to lower bound this with 1000, yes?

dataactcore/utils/cloudLogger.py
```diff
@@ -46,7 +46,7 @@ def logError(message,exception,traceback):
                 os.makedirs(path)
             localFile = os.path.join(path, "error.log")
             with open(localFile, "a") as file:
-                file.write("\n\n".join(["\n\n", message,
+                file.write(" ".join(["\n\n", message,
```
jworcestBAH: Agreed, that is cleaner, switched to this.

dataactvalidator/config/sqlrules/b18_object_class_program_activity_2.sql
jworcestBAH: So the goal here is to allow either the pairing (1,d) or the pairing (2,r).  The two clauses are basically saying 
NOT (1 and d) AND NOT (2 and r)
so it returns any records that don't match either pattern.  I've rearranged the line breaks to make the groupings a bit clearer:
'AND NOT (SUBSTRING(op.object_class,1,1) = '1' AND lower(op.by_direct_reimbursable_fun) = 'd')'
'AND NOT (SUBSTRING(op.object_class,1,1) = '2' AND lower(op.by_direct_reimbursable_fun) = 'r')'
In theory I could use <> by rewriting this part to be something like 'AND '((SUBSTRING(op.object_class,1,1) = '1' AND lower(op.by_direct_reimbursable_fun) <> 'd') OR (SUBSTRING(op.object_class,1,1) = '2' AND lower(op.by_direct_reimbursable_fun) <> 'r') OR
(SUBSTRING(op.object_class,1,1) NOT IN ['1','2']))'

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
jworcestBAH: I needed different ranges than the defaults to test specifically 3-digit (0,999) or 4-digit (1000,9999) object classes.  I chose fuzzy over randint here just for consistency with the rest of the factory object.  I'll switch back to randint.

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
jworcestBAH: I was basing this on the existing unit tests that check this as a subset, I'll ask if we expect the result to have any other columns or if this should be the complete result.

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
jworcestBAH: Fixed in this file, the second test still uses the dicts since it's looping through them.

tests/unit/dataactvalidator/test_b18_object_class_program_activity_2.py
jworcestBAH: Good catch, fixed

tests/unit/dataactvalidator/test_b18_object_class_program_activity_2.py
jworcestBAH: Okay broke this out into six tests, which also allowed me to get rid of the dicts defining the test cases

dataactvalidator/config/sqlrules/b18_object_class_program_activity_2.sql
cmc333333: Ah, thanks, I definitely wasn't picking up on the parens there due to line spacing

tests/unit/dataactvalidator/test_b18_object_class_program_activity_1.py
cmc333333: :+1: WFM

PR #289, opened 2016-09-21
Load multiple configs

jworcestBAH:
This allows optional loading from multiple config files for a base config, environment-specific config, and a secrets config.  Local installs would not be affected as the base config would continue to hold everything in that case.


dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+    return existing_config
+
+first_config = True
+log_message = ""
+for config_path in path_list:
+    try:
+        with open(config_path) as c:
+            CONFIG_ALL = yaml.load(c)
+    except IOError:
+        if first_config:
```
bsweger: Another decision point we could eliminate by standardizing the config file structure.

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+    ENV_PATH = os.path.join(dirname(abspath(__file__)), '{}_config.yml'.format(env))
+    SECRET_PATH = os.path.join(dirname(abspath(__file__)), '{}_secrets.yml'.format(env))
+    path_list = [CONFIG_PATH, ENV_PATH, SECRET_PATH]
+else:
+    # If no environment set, just use the base config
+    path_list = [CONFIG_PATH]
+# set the location of the Alembic config file
+ALEMBIC_PATH = os.path.join(dirname(abspath(__file__)), 'alembic.ini')
+MIGRATION_PATH = os.path.join(dirname(abspath(__file__)), 'migrations')
+CONFIG_BROKER = {}
```
bsweger: See comments below about not needing lines 31-35

.gitignore
```diff
@@ -40,3 +40,4 @@ dataactcore/scripts/dropMany.py
 runDynamo.py
 dataactvalidator/config/sf_133*.csv
 runBothTestPackages.py
+dataactcore/dev_config.yml
```
bsweger: Is the plan also to have a `prod_config`, `staging_config` (maybe even `local_config`), etc? If so, should those be in here too?

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+    CONFIG_SERVICES["protocol"] = "http"
+
+# Log some values from config
+for category_name in VALUES_TO_LOG:
+    category = globals()[category_name]
+    category_message = "### {}".format(category_name)
+    for key in VALUES_TO_LOG[category_name]:
+        value = category[key]
+        category_message = ", ".join([category_message, "{}:{}".format(key,value)])
+    log_message = " ".join([log_message, category_message])
```
bsweger: Apologies for not being up to speed enough to understand how the logging is happening here (I know you and @cmc333333 worked through some import things in the channel). Is setting `log_message` is enough to log something?

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+
+if CONFIG_SERVICES["broker_api_port"] == 443:
+    # Use https
+    CONFIG_SERVICES["protocol"] = "https"
+else:
+    CONFIG_SERVICES["protocol"] = "http"
+
+# Log some values from config
+for category_name in VALUES_TO_LOG:
+    category = globals()[category_name]
```
bsweger: Is the `globals()` reference needed here?

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+    category_message = "### {}".format(category_name)
+    for key in VALUES_TO_LOG[category_name]:
+        value = category[key]
+        category_message = ", ".join([category_message, "{}:{}".format(key,value)])
```
bsweger: I'd recommend sticking to one form of string formatting here (_i.e._, either `join` or `format` but not both), perhaps something like `category_message = "{} {}: {}".format(category_message, key, value)`

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
+import os
+import os.path
+from os.path import expanduser, normpath, dirname, abspath
+import yaml
+import re
+
+VALUES_TO_LOG = {"CONFIG_BROKER":["path","use_aws","local",'aws_bucket', 'aws_role', 'aws_region', 'aws_create_temp_credentials',
```
bsweger: Good call on explicitly listing what to log. My first thought was "why not just use a short list of what _not_ to log" and then realized that's a security issue waiting to happen.

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+                                  "rss_file", "generated_files_folder", "sf_133_folder", "sf_133_file"],
+                 "CONFIG_SERVICES":["error_report_path","broker_api_host","validator_host","broker_api_port","validator_port",
+                                    "protocol", "rest_trace", "server_debug"],
+                 "CONFIG_DB":["dynamo_host","dynamo_port","base_db_name","scheme"],
+                 "CONFIG_LOGGING":["log_files","use_logstash","logstash_host","logstash_port"],
+                 "CONFIG_JOB_QUEUE":["url", "port", "broker_scheme"]}
+
+# set the location of the DATA Act broker config files
+CONFIG_PATH = os.path.join(dirname(abspath(__file__)), 'config.yml')
+if "env" in os.environ:
```
jworcestBAH: Changed to this approach, there's actually no downside to doing it this way since missing files just generate a warning and use the default, so this gives users the option of having a separate secrets file or using a single file without needing the environment variable locally.

.gitignore
jworcestBAH: This was added to the gitignore while I experimented with an env specific config, given the changes above I'll change this to local_config and use that instead

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+    CONFIG_SERVICES['error_report_path'] = error_report_path
+
+    # if not using AWS and no DynamoDB settings specified, default
+    # to localhost and port 8000
+    if not CONFIG_DB['dynamo_host']:
+        # TODO: dynamo host/port don't do anything; code assumes localhost/8000
+        CONFIG_DB['dynamo_host'] = '127.0.0.1'
+    if not CONFIG_DB['dynamo_port']:
+        CONFIG_DB['dynamo_port'] = 8000
+    # TODO: can we test that local dynamo is up and running? if not, route calls hang
```
jworcestBAH: Removed

dataactcore/read_config.py
jworcestBAH: Ah good point, locals() makes more sense here, switched to that

dataactcore/read_config.py
jworcestBAH: Fixed

dataactcore/read_config.py
klrBAH: Can we potentially move this list itself into the config (or a not-to-log)? That way we won't have to modify config.py every time we add an item to config.yml. We'll just have to make a change to the config file we're already in.

bsweger: @klrBAH oh, that sounds like a good idea. What do you think @jworcestBAH?

jworcestBAH: Sure, I can do that.  Only difficulty is if we want to allows things to be marked for logging in multiple files.  As long as we can maintain the fields_to_log dict in one config this is a simple change, but if we want to combine different dicts from multiple files that would be inconsistent with the dict methods proposed below.  Would the assumption here be that if you want to log a field from the environment specific config, you'll either add it to the base config, or overwrite the list in the base config by relisting all fields to be logged in the environment-specific config?  Either of those is consistent with the overwriting approach we've been taking so far, I just want to make sure we're not expecting a different behavior for this field.

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+            # First config file must be present, the rest are optional
+            raise IOError('Error reading the config file. Please make sure this file exists'
+               ' before starting the DATA Act broker: {}'.format(config_path))
+        else:
+            log_message = " ".join([log_message, "### Did not load config file from {}".format(config_path), "warning"])
+            continue
+    # File successfully loaded, mark as no longer first
+    first_config = False
+
+    CONFIG_BROKER = merge_config(CONFIG_BROKER, CONFIG_ALL['broker']) if "broker" in CONFIG_ALL else CONFIG_BROKER
```
jworcestBAH: Switched to using update and did some related simplifications, merge_config is now gone

dataactcore/read_config.py
jworcestBAH: These lines got simplified to use CONFIG_CATEGORIES, but still need to exist in some form so update has a dict to run on for the first file

dataactcore/read_config.py
```diff
@@ -0,0 +1,165 @@
...
+# set the location of the Alembic config file
+ALEMBIC_PATH = os.path.join(dirname(abspath(__file__)), 'alembic.ini')
+MIGRATION_PATH = os.path.join(dirname(abspath(__file__)), 'migrations')
+CONFIG_BROKER = {}
+CONFIG_SERVICES = {}
+CONFIG_DB = {}
+CONFIG_LOGGING = {}
+CONFIG_JOB_QUEUE = {}
+
+def merge_config(existing_config, new_config):
```
jworcestBAH: merge_config is now gone

dataactcore/read_config.py
jworcestBAH: log_message gets imported to config.py, which then sends it to the logger.  I could move this part out to config.py if desired, but that wouldn't be possible for the part that puts the warnings for missing files into log_message, so we'd still have to import log_message in that case.

dataactcore/read_config.py
```diff
@@ -4,7 +4,7 @@
 import yaml
 import re
 
-VALUES_TO_LOG = {"CONFIG_BROKER":["path","use_aws","local",'aws_bucket', 'aws_role', 'aws_region', 'aws_create_temp_credentials',
+"""VALUES_TO_LOG = {"CONFIG_BROKER":["path","use_aws","local",'aws_bucket', 'aws_role', 'aws_region', 'aws_create_temp_credentials',
```
bsweger: Thanks for incorporating @klrBAH's suggestion re: putting this info into the actual log file. Can this string come out of `read_config` now?

jworcestBAH: Whoops good catch, removed that.

dataactcore/read_config.py
```diff
@@ -29,18 +31,9 @@
...
-
-def merge_config(existing_config, new_config):
-    if type(existing_config) is not dict or type(new_config) is not dict:
-        raise ValueError("Both args to merge_config must be dicts")
-    for key in new_config.keys():
-        existing_config[key] = new_config[key]
-    return existing_config
+
+for value in CONFIG_CATEGORIES.values():
+    locals()[value] = {}
```
bsweger: What does `locals()` do here? Does just referencing `value` directly not work?

dataactcore/read_config.py
```diff
@@ -58,12 +51,8 @@ def merge_config(existing_config, new_config):
...
     # File successfully loaded, mark as no longer first
     first_config = False
-
-    CONFIG_BROKER = merge_config(CONFIG_BROKER, CONFIG_ALL['broker']) if "broker" in CONFIG_ALL else CONFIG_BROKER
-    CONFIG_SERVICES = merge_config(CONFIG_SERVICES, CONFIG_ALL['services']) if "services" in CONFIG_ALL else CONFIG_SERVICES
-    CONFIG_DB = merge_config(CONFIG_DB, CONFIG_ALL['db']) if "db" in CONFIG_ALL else CONFIG_DB
-    CONFIG_LOGGING = merge_config(CONFIG_LOGGING, CONFIG_ALL['logging']) if "logging" in CONFIG_ALL else CONFIG_LOGGING
-    CONFIG_JOB_QUEUE = merge_config(CONFIG_JOB_QUEUE, CONFIG_ALL['job-queue']) if "job-queue" in CONFIG_ALL else CONFIG_JOB_QUEUE
+    for category_name in CONFIG_CATEGORIES:
+        locals()[CONFIG_CATEGORIES[category_name]].update(CONFIG_ALL.get(category_name, {}))
```
bsweger: Sorry all of the questions about `local()`. I feel like I must be missing something about what's happening here, but I'm not sure why it's needed.

dataactcore/read_config.py
jworcestBAH: So CONFIG_CATEGORIES is holding the names of the various config dictionaries, this is a way of creating the variable with the name stored in CONFIG_CATEGORIES.  If there's a better way to do that, please let me know, I'm also using this down in the logging section to get these variables.  If this were within a class I'd be using setattr(), is there a more direct equivalent of that for a current namespace?

dataactcore/read_config.py
bsweger: Would something like this work? (Assuming that we wouldn't need the `get` for the `CONFIG_ALL` lookup if we're requiring all three config files?)

```python
for category_name, category_dict in CONFIG_CATEGORIES.items():
        category_dict.update(CONFIG_ALL[category_name])
```

jworcestBAH:
Added a couple of alternative implementations, my preference would be for the second alternative where the CONFIG_CATEGORIES holds the actual dicts.  The only downside is that in the logging this uses names of the form "broker" rather than "CONFIG_BROKER", so the categories are less visually distinct, possibly we could use some other formatting to make the category names clearer in that.

klrBAH:
[run tests]

dataactcore/read_config.py
```diff
@@ -4,8 +4,13 @@
 import yaml
```
jworcestBAH: This diff got overly complicated by comparing against the first alternative.  The relevant differences from the original are lines 7-13, 39, and 130-136

jworcestBAH:
[run tests]

dataactcore/read_config.py
```diff
@@ -97,49 +98,45 @@
...
+if "values_to_log" in CONFIG_LOGGING:
     # If no values specified, don't do logging
-    for category_yaml_name in CONFIG_DICT["CONFIG_LOGGING"]["values_to_log"]:
-        category_name = CONFIG_CATEGORIES[category_yaml_name]
-        category = CONFIG_DICT[category_name]
-        category_message = "### {}".format(category_name)
-        for key in CONFIG_DICT["CONFIG_LOGGING"]["values_to_log"][category_yaml_name]:
+    for category_yaml_name in CONFIG_LOGGING["values_to_log"]:
+        category = CONFIG_CATEGORIES[category_yaml_name]
+        category = category
```
bsweger: Is this left-over from a find/replace? 😄

jworcestBAH:
[run tests]

dataactcore/read_config.py
jworcestBAH: I'd like to use that excuse, but sadly that was actually me manually replacing locals()[category_name] with the new category variable from line 133.  I just didn't actually look at the left side of the equation.  Removing that now.

klrBAH:
[run tests]

klrBAH:
[run tests]

bsweger:
@jworcestBAH Looks good. @klrBAH made one final tweak on her end, and tests ran successfully after that change, so I'm merging. Posted a note in Slack about needing to create new files.

PR #290, opened 2016-09-21
B14 and B15 updated to sum for each tas, unit tests updated to match

jworcestBAH:
These rules were updated to check the sum for all records within a TAS against the sf-133 amount for that TAS rather than checking each record against the amount individually.

### DIFF
--- a/tests/unit/dataactvalidator/test_b15_object_class_program_activity.py
+++ b/tests/unit/dataactvalidator/test_b15_object_class_program_activity.py
@@ -11,6 +11,46 @@ def test_success(database):
     """ Tests that SF 133 amount sum for line 2104 matches the calculation from Appropriation based on the fields below
         for the specified fiscal year and period """
     tas = "".join([_TAS, "_success"])
+    tas2 = "".join([_TAS, "_other_tas"])
+
+    # This uses the default submission created in utils for 10/2015 which is period 1 of FY 2016
+    sf = SF133(line=2104, tas=tas, period=1, fiscal_year=2016, amount=15, agency_identifier="sys",
+               main_account_code="000", sub_account_code="000")
+
+    op = ObjectClassProgramActivity(job_id=1, row_number=1, tas=tas, by_direct_reimbursable_fun='r',
+                                    ussgl480100_undelivered_or_cpe=1, ussgl480100_undelivered_or_fyb=1,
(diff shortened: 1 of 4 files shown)


tests/unit/dataactvalidator/test_b15_object_class_program_activity.py
```diff
@@ -23,16 +25,34 @@ def test_success(database):
...
+    op2 = ObjectClassProgramActivity(job_id=1, row_number=2, tas=tas, by_direct_reimbursable_fun='r',
+                                    ussgl480100_undelivered_or_cpe=2, ussgl480100_undelivered_or_fyb=2,
+                                    ussgl480200_undelivered_or_cpe=2, ussgl480200_undelivered_or_fyb=2,
+                                    ussgl488100_upward_adjustm_cpe=2, ussgl488200_upward_adjustm_cpe=2,
+                                    ussgl490100_delivered_orde_cpe=2, ussgl490100_delivered_orde_fyb=2,
+                                    ussgl490200_delivered_orde_cpe=2, ussgl490800_authority_outl_cpe=2,
+                                    ussgl490800_authority_outl_fyb=2, ussgl498100_upward_adjustm_cpe=2,
+                                    ussgl498200_upward_adjustm_cpe=2)
+
+    # Record for other TAS should not be included in sum
```
nmonga91: Great check to include in the test!

tests/unit/dataactvalidator/test_b15_object_class_program_activity.py
```diff
@@ -11,8 +11,10 @@ def test_success(database):
     """ Tests that SF 133 amount sum for line 2104 matches the calculation from Appropriation based on the fields below
         for the specified fiscal year and period """
     tas = "".join([_TAS, "_success"])
+    tas2 = "".join([_TAS, "_other_tas"])
```
nmonga91: For simple string concatenation, do we still want to use `join` or is `+` simple enough to use? I know it has to deal with flexibility/efficiency, but since this is just 2 simple strings, I'm curious if there's a difference.

jworcestBAH: I don't really have strong preferences here, but I know in the past we've changed away from using '+' for string concatenation so I've generally been leaning towards either format or join.

nmonga91:
Looks good, merging!

PR #291, opened 2016-09-21
Handle empty values in non-padded TAS fields.

bsweger:
On our latest SF-133 file from GTAS, a few fields that had
previously been filled with spaces are now coming in as empty
values. We'll speak to the team about  not
altering the file format month-to-month, but it seemed
reasonable to make this small fix to handle either case.

### DIFF
--- a/dataactvalidator/scripts/loadFile.py
+++ b/dataactvalidator/scripts/loadFile.py
@@ -145,11 +145,14 @@ def loadSF133(filename, fiscal_year, fiscal_period, force_load=False):
              "main_account_code": {"pad_to_length": 4},
              "sub_account_code": {"pad_to_length": 3},
              # next 3 lines handle the TAS fields that shouldn't
-             # be padded but should still be empty spaces,
-             # at least temporarily, so that downstream formatting
-             # and cleaning functions work correctly. this works
-             # around the fact that sometimes the incoming data
-             # is a single space and sometimes it's blank.
+             # be padded but should still be empty spaces rather
+             # than NULLs. this ensures that the downstream pivot & melt
+             # (which insert the missing 0-value SF-133 lines)
+             # will work as expected (values used in the pivot
+             # index cannot be NULL).
+             # the "pad_to_length: 0" works around the fact
+             # that sometimes the incoming data for these columns
+             # is a single space and sometimes it is blank/NULL.
              "beginning_period_of_availa": {"pad_to_length": 0},
              "ending_period_of_availabil": {"pad_to_length": 0},
              "availability_type_code": {"pad_to_length": 0},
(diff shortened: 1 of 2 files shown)


dataactvalidator/scripts/loadFile.py
```diff
@@ -144,6 +144,15 @@ def loadSF133(filename, fiscal_year, fiscal_period, force_load=False):
...
              "sub_account_code": {"pad_to_length": 3},
+             # next 3 lines handle the TAS fields that shouldn't
+             # be padded but should still be empty spaces,
+             # at least temporarily, so that downstream formatting
+             # and cleaning functions work correctly. this works
+             # around the fact that sometimes the incoming data
+             # is a single space and sometimes it's blank.
+             "beginning_period_of_availa": {"pad_to_length": 0},
+             "ending_period_of_availabil": {"pad_to_length": 0},
+             "availability_type_code": {"pad_to_length": 0},
```
jworcestBAH: Should these explicitly specify the 'keep_null' option, since that seems to be the handling we're trying to get here?  More generally, since stripping extra whitespace and converting blanks to nulls is something we wanted to always do independently of padding, should this functionality happen in LoaderUtils.cleanData before the options are applied?  FieldCleaner.cleanRow does this for the values in the files provided by users, so we may want to apply the same handling to domain value loading to ensure they match.

bsweger: That is a fair question and a long story.

**Short story**
We don't want to keep_null..the new sf-133 file had some nulls where it previously had spaces, which causes a downstream problem. This fix pads to 0 to change incoming nulls to spaces (everything is set back to null before the db load).

It would also be fair to ask, "why doesn't the SF-133 load have unit tests if it's so complicated?" If you want to hold the merge until said tests are written, I would not blame you.

**Long story**
The requirement to interpret missing rows and insert them as 0 values makes the SF-133 load different than the other domain loaders.

Here's the code at the crux of the problem: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/hotfix-sf133-load/dataactvalidator/scripts/loadFile.py#L174

Lines 174-75 are a shortcut to fill in the 0 values. Line 174 uses `pivot_table` to explode the data out horizontally, creating one row for each tas/fiscal year/period, with columns for each SF-133 line. The `fill_value` parameter puts `0` into any SF-133 cell with a missing value for that tas/fiscal year/period.

Once we have the zeroes filled in, line 175 uses the `melt` function to re-arrange the data back to it's normal format of one row per tas/fiscal year/period.

It was the shortest and most efficient way I could think of to put in the missing zeroes. There's a catch, though--if any value that we pivot on in line 174 is null, that row is silently dropped by pandas. I lost several hours of my life tracking this down.

Which is a long way of saying...we do need the incoming TAS components to be blank values instead of NULLS to ensure that no lines are dropped in the pivot. Once we have the 0 value lines, we replace those empty strings with None before inserting to the db: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/hotfix-sf133-load/dataactvalidator/scripts/loadFile.py#L196

jworcestBAH: Thanks for the explanation, that does make it much clearer.  I do wonder if we really need to have the pivot index include that many columns, shouldn't 'tas' + 'period' be sufficient to guarantee uniqueness when combined with 'line' as the columns?  Reducing to those would have the side benefit that none of those should be null, which might allow some steps of this process to be simplified.  Or are the extra fields just part of the pivot index to keep them around for the final structure after the melt, and they can't be shifted to values because fill_value can't be limited to a specific value?

bsweger: @jworcestBAH yep, that last thing you said. This thread is a good note that I should expand the comments in the SF-133 script to be more clear on what is happening with the pivot. Will do that and push.

bsweger:
@jworcestBAH Just pushed some additional comments, which will hopefully be useful to the next person who looks at this code--thanks for the good questions and the nudge.

FWIW, this load is slated to be replaced when the GTAS team implements an API, and I look forward to that day.

Again, I wouldn't blame you for holding this for lack of unit tests. I might not get to those for a few days, but I think that's fine.

jworcestBAH:
Okay, looks good, merging now.

PR #293, opened 2016-09-22
Update db connection integration test

bsweger:
This PR is a first stab at the `function_bag` and using using `flask.g` to store on db information. It's a small piece of work that removes the `userHandler` dependency when setting up test users in `baseTestAPI`.

@cmc333333 you're probably the best candidate to take a look, since you've been helping me think how all of this might work. But @jworcestBAH and @nmonga91 should feel free to weigh in too, since this will impact how we code db access going forward.

Apologies--there's also one miscellaneous change that made it into the PR:
* I'd missed a `userHandler` dependency in `dataactbroker/scripts/initialize.py` during a previous PR. That's set up to use `databaseSession`...the plan is go back and change all of that to use `GlobalDB` instead if it works well in the integration tests code.


tests/integration/baseTestAPI.py
```diff
@@ -73,84 +73,88 @@ def setUpClass(cls):
...
+            sess.add(statusChangedUser)
+            cls.status_change_user_id = statusChangedUser.user_id
+
+            # set up deactivated user
+            deactivated_user = sess.query(User).filter(User.email == test_users['inactive_email']).one()
+            deactivated_user.last_login_date = time.strftime("%c")
+            deactivated_user.is_active = False
+            sess.add(deactivated_user)
+
+            sess.commit()
```
cmc333333: Nice, single `commit`

dataactcore/interfaces/function_bag.py
```diff
@@ -17,15 +19,16 @@
...
-    sess = GlobalDB.db().session
-    status = sess.query(UserStatus).filter(UserStatus.name == 'approved').one()
-    user = User(email=email, user_status=status, permissions=permission,
-                cgac_code=cgac_code)
-    pwd = getPasswordHash(password, bcrypt)
-    user.salt = pwd[0]
-    user.password_hash = pwd[1]
-    sess.add(user)
-    sess.commit()
+    with createApp().app_context():
```
cmc333333: We probably want to create this context outside of `createUserWithPassword` -- to make it as unobtrusive as possible, we'd want to push the context set up as close to the "start" of the application as possible. Maybe at the test setup?

bsweger:
Woot, thanks for for the thorough review @cmc333333. I'll take you up on all suggestions/fixes but will focus on the application context question first, so the team can have better clarity on the blueprint for this.

bsweger:
[run-tests]

dataactbroker/scripts/initialize.py
```diff
@@ -53,17 +53,13 @@ def createAdmin():
...
-    except ResponseException as e:
-
-        if type(e.wrappedException) is NoResultFound:
-            userDb.createUserWithPassword(
-                adminEmail, adminPass, Bcrypt(), permission=2)
-            user = userDb.getUserByEmail(adminEmail)
-            userDb.addUserInfo(user, "Admin", "SYS", "System Admin")
-    userDb.close()
+    with databaseSession() as sess:
+        user = sess.query(User).filter(User.email==adminEmail).one_or_none()
```
cmc333333: Never mind! There's no need for a case insensitive check here since the email comes from the config file.

dataactbroker/scripts/initialize.py
```diff
@@ -53,17 +53,13 @@ def createAdmin():
...
+        sess.close()
+    if not user:
+        user = createUserWithPassword(
+            adminEmail, adminPass, Bcrypt(), permission=2)
```
bsweger: Good point. Since the `createUserWithPassword` is already set up to hard-code the agency code as `CGAC`, I'll update that function to set the name and title to something appropriately admin-y sounding.

dataactcore/interfaces/dbInterface.py
```diff
@@ -0,0 +1,40 @@
...
+# the db access within Flask requests, fire up an ad-hoc db session in
+# these transitional functions.
+
+
+def createUserWithPassword(email, password, bcrypt, permission=1, cgac_code="SYS"):
+    """Convenience function to set up fully-baked user (used for setup/testing only)."""
+    with databaseSession() as sess:
+        status = sess.query(UserStatus).filter(UserStatus.name == 'approved').one()
+        user = User(email=email, user_status=status, permissions=permission,
+                    cgac_code=cgac_code)
```
bsweger: Yes, agreed! I'm going to leave this as part of our future work to re-think our overall strategy to dealing with helper table values.

dataactcore/interfaces/dbInterface.py
```diff
@@ -0,0 +1,40 @@
...
+
+    return user
+
+
+def getPasswordHash(password, bcrypt):
+    """Generate password hash."""
+    # TODO: handle password hashing/lookup in the User model
+    salt = uuid.uuid4().hex
+    # number 12 below iw the number of rounds for bcrypt
+    hash = bcrypt.generate_password_hash(password + salt, 12)
```
bsweger: Yes, good call. I'll put it into a constant and will punt the config idea until we can get all of this password hashing stuff happening in the user model instead of here in the function bag.

bsweger:
@cmc333333 Thanks again for the review and for working through the db handling with me. I made a few minor changes too, based on your feedback. This is ready for a final look whenever you're ready!

cmc333333:
Nice work, @bsweger !

PR #294, opened 2016-09-22
Db 944 new file a validations

nmonga91:
- [x] @bsweger's sign off that PR is ready to be merged


dataactvalidator/config/sqlrules/a36_appropriations.sql
```diff
@@ -0,0 +1,13 @@
...
+    sf.amount as sf_133_amount
+FROM appropriation as approp
+    INNER JOIN sf_133 as sf ON approp.tas = sf.tas
+    INNER JOIN submission as sub ON approp.submission_id = sub.submission_id AND
+        sf.period = sub.reporting_fiscal_period AND
+        sf.fiscal_year = sub.reporting_fiscal_year
+WHERE approp.submission_id = {} AND
+    sf.line = 1000 AND
+    sf.amount <> 0 AND
+    approp.budget_authority_unobligat_fyb IS NULL
```
bsweger: Apologies, because I'm pretty sure I missed this convo on Slack, but is it possible that the appropriation file would arrive with a 0 value? If so, does 0 count as not being "populated"?

dataactvalidator/config/sqlrules/a34_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    approp.row_number,
+    approp.budget_authority_unobligat_fyb,
+    sf.amount as sf_133_amount
```
bsweger: Does it make sense to add the SF-133 line number as part of the column alias to make it more explicit for folks tracking down errors? Something like `sf_133_2490_amount`? Or do you think the rule text itself is enough of a clue?

tests/unit/dataactvalidator/test_a34_appropriations.py
```diff
@@ -0,0 +1,33 @@
...
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Tests that SF-133 amount for line 2490 for the end of the last fiscal year
+    equals Appropriation budget_authority_unobligat_fyb """
+
+    sf = SF133Factory(line=2490, tas=_TAS, period=12, fiscal_year=2015, amount=1)
+    ap = AppropriationFactory(tas=_TAS, budget_authority_unobligat_fyb=1)
```
bsweger: Do we also need a submission factory to make sure the year -1 part of this rule is working as intended? Honestly, I'm not sure of the best thing to do here, when we're checking a prior year.

tests/unit/dataactvalidator/test_a34_appropriations.py
```diff
@@ -0,0 +1,33 @@
...
+
+    assert number_of_errors(_FILE, database, models=[sf, ap]) == 0
+
+
+def test_failure(database):
+    """ Tests that SF-133 amount for line 2490 for the end of the last fiscal year
+    does not equal Appropriation budget_authority_unobligat_fyb """
+
+    sf = SF133Factory(line=2490, tas=_TAS, period=12, fiscal_year=2015, amount=1)
+    ap = AppropriationFactory(tas=_TAS, budget_authority_unobligat_fyb=0)
```
bsweger: Same question here about involving the submission's FY as part of the test

tests/unit/dataactvalidator/test_a34_appropriations.py
```diff
@@ -0,0 +1,33 @@
...
+
+_FILE = 'a34_appropriations'
+_TAS = 'a34_appropriations_tas'
+
+
+def test_column_headers(database):
+    expected_subset = {'row_number', 'budget_authority_unobligat_fyb', 'sf_133_amount'}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
```
bsweger: Am loving these unit tests for the rules...so easy to read! Do you think we need a test case for the scenario of a new tas/new fiscal year where we expect the rule to pass where there's no corresponding SF-133 record? Is that taking things too far?

tests/unit/dataactvalidator/test_a35_cross_file.py
```diff
@@ -0,0 +1,69 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number', 'deobligations_recoveries_r_cpe', 'ussgl487100_downward_adjus_cpe_sum',
+                       'ussgl497100_downward_adjus_cpe_sum', 'ussgl487200_downward_adjus_cpe_sum',
+                       'ussgl497200_downward_adjus_cpe_sum'}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Tests that, for entries with the matching TAS, Appropriations deobligations_recoveries_r_cpe equals the sum of
```
bsweger: Thanks for this detailed docstring--really handy not to reference the rule text to understand what's happening here.

tests/unit/dataactvalidator/test_a35_cross_file.py
```diff
@@ -0,0 +1,69 @@
...
+
+    ap = AppropriationFactory(tas=_TAS, deobligations_recoveries_r_cpe=8)
+
+    op_1 = ObjectClassProgramActivityFactory(tas=_TAS, ussgl487100_downward_adjus_cpe=1, ussgl497100_downward_adjus_cpe=1,
+                                             ussgl487200_downward_adjus_cpe=1, ussgl497200_downward_adjus_cpe=1)
+
+    op_2 = ObjectClassProgramActivityFactory(tas=_TAS, ussgl487100_downward_adjus_cpe=1, ussgl497100_downward_adjus_cpe=1,
+                                             ussgl487200_downward_adjus_cpe=1, ussgl497200_downward_adjus_cpe=1)
+
+    op_3 = ObjectClassProgramActivityFactory(tas=tas_ignore, ussgl487100_downward_adjus_cpe=1, ussgl497100_downward_adjus_cpe=1,
```
bsweger: Oh, very nice! Love that you put in an extra tas to check the `group by`!

dataactvalidator/config/sqlrules/a36_appropriations.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    approp.row_number,
+    approp.budget_authority_unobligat_fyb,
+    sf.amount as sf_133_amount
```
bsweger: Similar question here as for the A34 rule..should we make the column alias more explicit?

tests/unit/dataactvalidator/test_a36_appropriations.py
```diff
@@ -0,0 +1,38 @@
...
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Tests that if the SF-133 amount for line 1000 for the same fiscal year and period is populated, then
+    Appropriation budget_authority_unobligat_fyb is also populated"""
+
+    sf = SF133Factory(line=1000, tas=_TAS, period=1, fiscal_year=2016, amount=1)
+    ap = AppropriationFactory(tas=_TAS, budget_authority_unobligat_fyb=1)
+
```
bsweger: Same question as above re: mocking up the submission date on these SF-133 validations. If we haven't been doing it thus far, I'm okay to move forward.

dataactvalidator/config/sqlrules/a36_appropriations.sql
nmonga91: I looked up the field in appropFields.csv and there it said it wasn't required, from that I assumed that they could leave it blank, which would translate to NULL. But I do remember that if the field is required, 0 could be interpreted as not being "populated".

dataactvalidator/config/sqlrules/a34_appropriations.sql
nmonga91: The rule text has the line # in there, so I think that would be enough to track down the error. In previous rules, we haven't included the line number in the column alias, so if we decide to go that route, those would need to be updated as well.

tests/unit/dataactvalidator/test_a34_appropriations.py
nmonga91: We can achieve that test without needing a Submission factory. I can change the year and expect a failure which would sufficiently test that. What do you think?

tests/unit/dataactvalidator/test_a34_appropriations.py
nmonga91: I'm not too sure that's necessary. It's just going to return 0 errors since the fiscal year/tas won't match so it wouldn't do much benefit, unless there is some inherent benefit which i'm not seeing (please correct me if i'm wrong, i'm not too sure about this).

dataactvalidator/config/sqlrules/a36_appropriations.sql
bsweger: @nmonga91 Thanks for clarifying! The required/non-required distinction is good to know.

dataactvalidator/config/sqlrules/a34_appropriations.sql
bsweger: Agreed that the rule text is enough context for people to track this down, so we're all good here.

In general, though, I don't think we need to prioritize consistency of past code over making improvements. I'd hate for the notion that we have to retrofit every enhancement to prevent us from doing incremental improvements!

tests/unit/dataactvalidator/test_a34_appropriations.py
bsweger: Heh, here's where I'm getting lost, and it might be because I haven't written of the new unit tests yet...

How does the test know how to check the year w/o having a submission object? Thank you for helping a wayward unit test soul understand this magic.

dataactvalidator/config/sqlrules/a34_appropriations.sql
nmonga91: I agree with the latter statement. Just because it's how it's been done previously doesn't mean it's still a good solution.

tests/unit/dataactvalidator/test_a34_appropriations.py
nmonga91: So in `number_of_errors` it automatically inserts a default submission that has a fiscal year of 2016 and fiscal period of 1. So i've been writing tests against that default submission altering the fiscal year/period for the different Files (A/B/C) to test.

bsweger: Ah--thanks for explaining that. Yeah, if we could test a bad/unexpected year since this is an unusual rule, that would be great.

nmonga91: I will add those tests

bsweger: Thanks for the additional check!

bsweger:
Looks good--mergine!

PR #295, opened 2016-09-22
SQL Validations and Unit Tests for A25, A26, A27, A28, and A31 (A/B/C)

mab6bf:


tests/unit/dataactvalidator/test_a25_appropriations.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_success(database):
+    """ Tests that BorrowingAuthorityAmountTotal_CPE is provided if TAS has borrowing authority value
+    provided in GTAS """
+    tas = "".join([_TAS, "_success"])
+
+    sf1 = SF133(line=1340, tas=tas, period=1, fiscal_year=2016, amount=1, agency_identifier="sys",
+               main_account_code="000", sub_account_code="000")
+    sf2 = SF133(line=1440, tas=tas, period=1, fiscal_year=2016, amount=1, agency_identifier="sys",
+               main_account_code="000", sub_account_code="000")
```
jworcestBAH: Should this be using the factory in tests.dataactcore.factories.domain so you don't need to specify the values you're not using?

tests/unit/dataactvalidator/test_a25_appropriations.py
```diff
@@ -0,0 +1,43 @@
...
+
+    ap = Appropriation(job_id=1, row_number=1, tas=tas, borrowing_authority_amount_cpe=1)
```
jworcestBAH: Could also use the factory for this just to have a more complete model rather than having everything else being null, not too important here but since we have them we may as well have the tests be one step closer to realistic.

dataactvalidator/config/sqlrules/a25_appropriations.sql
```diff
@@ -0,0 +1,12 @@
...
+    approp.borrowing_authority_amount_cpe
+FROM appropriation AS approp
+    INNER JOIN sf_133 AS sf ON approp.tas = sf.tas
+    INNER JOIN submission AS sub ON approp.submission_id = sub.submission_id
+        AND sf.period = sub.reporting_fiscal_period
+        AND sf.fiscal_year = sub.reporting_fiscal_year
+WHERE approp.submission_id = {}
+    AND (sf.line = 1340 OR sf.line = 1440)
+    AND sf.amount > 0
+    AND (approp.borrowing_authority_amount_cpe = 0 OR approp.borrowing_authority_amount_cpe IS NULL);
```
mab6bf: Good calls - implementing

dataactvalidator/config/sqlrules/a25_appropriations.sql
```diff
@@ -0,0 +1,12 @@
+SELECT DISTINCT
+    approp.row_number,
+    approp.borrowing_authority_amount_cpe
```
mab6bf: Implemented

tests/unit/dataactvalidator/test_a25_appropriations.py
mab6bf: Implemented

tests/unit/dataactvalidator/test_a25_appropriations.py
mab6bf: Good catch - implemented

jworcestBAH:
Looks good, merging

PR #297, opened 2016-09-23
Apply rule C24 requirements to existing rule C8

bsweger:
Per rule C24, validation rule C8 should not be applied to File C rows
that have a valid CGAC code in the allocation transfer agency column.


tests/unit/dataactvalidator/test_c8_award_financial.py
```diff
@@ -0,0 +1,188 @@
...
+    tas = _TAS
+    af = AwardFinancialFactory(
+        tas=tas, fain=None, uri=None, allocation_transfer_agency=None)
+    afa = AwardFinancialAssistanceFactory(
+        tas=tas, submission_id=af.submission_id, fain=None, uri='abc')
+
+    errors = number_of_errors(_FILE, database, models=[af, afa])
+    assert errors == 1
+
+def test_unequal_uri_null(database):
```
mab6bf: Is this a duplication of test_unequal_uri_null in lines 108-118?

bsweger: Yes, yes it is. Removed--thank you!

PR #298, opened 2016-09-23
Adding SQL and unit tests for Rules C11 and C12

nmonga91:


dataactcore/models/stagingModels.py
```diff
@@ -477,4 +477,4 @@ def __init__(self, **kwargs):
         # broker is set up to ignore extra columns in submitted data
         # so get rid of any extraneous kwargs before instantiating
         cleanKwargs = {k: v for k, v in kwargs.items() if hasattr(self, k)}
-        super(Appropriation, self).__init__(**cleanKwargs)
+        super(AwardProcurement, self).__init__(**cleanKwargs)
```
cmc333333: Ouch! Good catch!

dataactvalidator/config/sqlrules/c11_cross_file.sql
```diff
@@ -0,0 +1,13 @@
...
+	af.piid,
+    af.parent_award_id
+FROM award_financial AS af
+WHERE af.submission_id = {}
+	AND NOT EXISTS (
+		SELECT 1
+		FROM award_procurement AS ap
+		WHERE ap.submission_id = af.submission_id
+		    AND ap.piid = af.piid
+		    AND ap.parent_award_id = af.parent_award_id
```
cmc333333: Does this need to compare reporting period, too? Rule is

> Unique PIID, ParentAwardId from file C exists in file D1 during the same reporting period.

dataactvalidator/config/sqlrules/c12_cross_file.sql
```diff
@@ -0,0 +1,16 @@
...
+WHERE ap.submission_id = {}
+    AND (ap.federal_action_obligation IS NOT NULL
+        AND ap.federal_action_obligation <> ''
+        AND ap.federal_action_obligation <> '0')
+    AND NOT EXISTS (
+		SELECT 1
+		FROM award_financial AS af
+		WHERE af.submission_id = ap.submission_id
+		    AND af.piid = ap.piid
+		    AND af.parent_award_id = ap.parent_award_id
```
cmc333333: Similar question re: reporting period.

bsweger:
@nmonga91 I know that @cmc333333 is reviewing this, but was the plan to incorporate the C24 changes into these rules? Or should I take a second pass through and add that bit + tests?

tests/unit/dataactvalidator/test_c11_cross_file.py
```diff
@@ -0,0 +1,41 @@
...
+
+    af = AwardFinancialFactory(piid='some_piid', parent_award_id='some_parent_award_id')
+    ap_1 = AwardProcurementFactory(piid='some_piid', parent_award_id='some_parent_award_id')
+    ap_2 = AwardProcurementFactory(piid='some_piid', parent_award_id='some_parent_award_id')
+
+    assert number_of_errors(_FILE, database, models=[af, ap_1, ap_2]) == 0
+
+
+def test_failure(database):
+    def test_success(database):
```
cmc333333: Function within a function -- I don't think py.test is calling the inner one

nmonga91: good catch! that was a mistake, will fix

dataactvalidator/config/sqlrules/c11_cross_file.sql
nmonga91: After talking to Kaitlin (if i remember correctly), same reporting period is satisfied by making sure the submission IDs match, since the submission is what records the reporting period.

cmc333333: Ah, good point

PR #299, opened 2016-09-25
Update docs to match environment-specific broker config.

bsweger:
Create two new config samples (local_config_example.yml and
local_secrets_example.yml) to match the changes to the broker's
config handling. Also update the install docs to explain how
these files are used.

Took my best guess at a reasonable way of splitting up values between the three sample files. Happy to go back and forth on this. Ultimately, though, I think the sample files matching our particular use is less important than making sure that folks reading the install docs know what to do.

### DIFF
--- a/dataactcore/local_config_example.yml
+++ b/dataactcore/local_config_example.yml
@@ -1,0 +1,44 @@
+#######################################################
+### Sample Local Data Act Broker Configuration.     ###
+### Values here will override values in config.yml. ###
+#######################################################
+
+broker:
+
+    use_aws: false
+    full_url: http://localhost:3000
+    reply_to_email: valid.developer.email@domain.com
+    broker_files: /full/path/to/local/broker/files
+    d_file_storage_path: /full/path/to/local/d/file/storage
+
+services:
+
+    broker_api_host: 127.0.0.1
+    broker_api_port: 3333
+    validator_host: 127.0.0.1
+    validator_port: 3334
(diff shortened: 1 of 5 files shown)


doc/CONTRIBUTING.md
```diff
@@ -159,28 +159,24 @@ If you are using virtualenvwrapper-powershell in Windows, the following command
...
-        * `broker_files`
-    * under _services_:
-        * `error_report_path`
-    * under _db_:
-        * `username`
-        * `password`
-    * under _logging_:
-        * `log_files`
-4. Save your changes to config.yml
+### Create Broker Config Files
```
nmonga91: +1 for making sure the instructions are also up to date!

dataactcore/local_config_example.yml
```diff
@@ -0,0 +1,44 @@
...
+    validator_port: 3334
+    error_report_path: error_report_path: /full/path/to/local/error/reports
+
+db:
+
+    base_db_name: postgres
+    scheme: postgres
+    host: localhost
+    port: 5432
+    base_db_name: postgres #This is the default db on the instance.
```
nmonga91: `base_db_name` is present twice. I think the one with the comment should stay and the first occurrence may be the one to be removed.

bsweger:
@nmonga91 Thanks for the review. I pushed a change to remove the duplicate `base_db_name` param--thanks for catching that!

PR #300, opened 2016-09-26
Add c24 to c11

bsweger:
Update validation rule C11 to incorporate the requirements of rule C24. In other words, C11 should never fail for records that have a valid allocation transfer agency specified.


nmonga91:
Looks good, merging!

PR #301, opened 2016-09-26
Update file F's write_csv

cmc333333:
As the parameters to write_csv have changed, we need to modify their usage in file F generation, too. This also adds back a unit test which had been accidentally deleted to catch this sort of error in the future.


jworcestBAH:
Looks good, merging

PR #302, opened 2016-09-26
Make sure alembic is aware of all models in use.

bsweger:
I had removed some of these model imports  when making  unrelated changes
to env.py b/c the IDE was marking them as unused. They do need to be
there, however, else Alembic won't know which models to check when
autogenerating migrations.


cmc333333:
:+1: sorry, I should have caught that during review!

PR #303, opened 2016-09-27
remove db interface from filetests setup

bsweger:
- [x] Dev stable branch from Sprint 22 is created
- [x] Code reviewer approves

**Update:** can be merged when code review is complete

This is a continuation of the db refactor work. I'm going through the integration tests suites to remove the interface and interface holder references. This series of commits:

* Creates new foreign keys on the job table, which we weren't able to do when the databases were separate.
* Removes interfaces from the fileTests.py setup, replacing it with the new global db object/flask application context.
* Creates lookups info for things like file types and job status to the base test setup method and stores them as class variables. Ultimately, we want a better way to handle these lookups, but this is a first step that at least puts the values in a single place and eliminates magic numbers from the test setup.

Still to do for fileTests is to remove the interface dependencies from the actual test code.


tests/integration/baseTestAPI.py
```diff
@@ -159,6 +162,21 @@ def setUpClass(cls):
...
+            cls.jobTypeDict = {j.name: j.job_type_id for j in jt.all()}
+            ft = sess.query(FileType)
+            cls.fileTypeDict = {f.name: f.file_type_id for f in ft.all()}
+            fs = sess.query(FileStatus)
+            cls.fileStatusDict = {f.name: f.file_status_id for f in fs.all()}
+            sev = sess.query(RuleSeverity)
+            cls.ruleSeverityDict = {s.name: s.rule_severity_id for s in sev.all()}
+            et = sess.query(ErrorType)
+            cls.errorTypeDict = {e.name: e.error_type_id for e in et.all()}
+
```
jworcestBAH: Should this instead use the dicts attached to the model objects?  For example, JobStatus has JOB_STATUS_DICT.  These can also be used with the BaseInterface.getIdFromDict function, which will use the dictionary when present and populate it if it is empty (i.e. on the first call).  That function is also used by the various helper functions in the interface objects that provide these values, like JobTrackerInterface.getJobStatusId (also has getJobTypeId and getFileTypeId).  I understand that the goal here is to use a different way of getting the session object, but I'm wondering if our approach should be to have a way to provide that session to the interface objects rather than duplicating those functions whenever they're needed outside the app.  Not too big a deal on this one, but may be worth thinking about.

dataactcore/models/errorModels.py
```diff
@@ -43,8 +43,11 @@ class ErrorMetadata(Base):
...
+    file_type = relationship("FileType", foreign_keys=[file_type_id])
     # Second file type id is used in cross file errors
-    target_file_type_id = Column(Integer)
-    original_rule_label = Column(Text, nullable = True)
-    severity_id = Column(Integer)
+    target_file_type_id = Column(Integer, ForeignKey("file_type.file_type_id", name="fk_target_file_type_file_status_id"))
+    target_file_type = relationship("FileType", foreign_keys=[target_file_type_id])
+    original_rule_label = Column(Text, nullable=True)
+    severity_id = Column(Integer, ForeignKey("rule_severity.rule_severity_id", name="fk_error_severity_id"))
+    severity = relationship("RuleSeverity")
```
jworcestBAH: Thanks for adding these, it's good to start getting more use out of the consolidated DB.

tests/integration/baseTestAPI.py
bsweger: @jworcestBAH Agreed that we need a better overall approach to handling these types of lookups. I didn't move `getIdFromDict` to the function bag because I would like to phase it out in favor of a lookup method that has less cognitive overhead, if you will, even if saves some extra db calls on these small tables.

Your comment, however, made me think harder about a way of having it both ways--minimize the db lookups while having a more straightforward way of looking up values. 

@cmc333333 pointed out that because our lookup values don't change as the app runs, we could create a `lookups.py` that has the values for each helper table. We could reference `lookups.py` in the db setup scripts, and the app itself could also use it.

What do you think?



Ideally, we could use postgres enums for these types of values, but I'm not sure how hard it is to implement those post-schema design, on an existing body of data.

bsweger: @jworcestBAH I gave this idea a try over in PR #330. It worked very well for doing the initial db setup scripts. Once that's merged, I'd like to try it out over here and replace this code that dynamically creates lookup dictionaries by querying the lookup tables.

bsweger:
@jworcestBAH just pushed an update to replace the dynamically-created lookup dictionaries with the new constants in lookup.py.

jworcestBAH:
Looks good, merging

PR #304, opened 2016-09-27
Db 925 d1 basic validation

jworcestBAH:
This loads the D1 file into the staging table so C-D1 rules can be run


dataactvalidator/config/awardProcurementFields.csv
```diff
@@ -1,194 +1,194 @@
...
-other_minority_owned_business,other_minority_owned_busin,FALSE,str,1,,
-for_profit_organization,for_profit_organization,FALSE,str,1,,
-nonprofit_organization,nonprofit_organization,FALSE,str,1,,
-other_not_for_profit_organization,other_not_for_profit_organ,FALSE,str,1,,
-u.s._local_government,us_local_government,FALSE,str,1,,
-referenced_idv_modification_number,referenced_idv_modificatio,FALSE,str,25,,
-undefinitized_action,undefinitized_action,FALSE,str,1,,
-domestic_or_foreign_entity,domestic_or_foreign_entity,FALSE,str,1,,
+piid,piid,FALSE,str,,,
+awardingsubtieragencycode,awarding_sub_tier_agency_c,FALSE,str,,,
```
bsweger: Are the changes made to this file a way to skip basic D1 validations? I vaguely remember something like this on the D2 file, maybe. Will this make it harder for us to update the IDD specs as they change, having to remember the manual edits we did to these files?

I don't think we have time at this point to rework, but I'd like to think about an alternate way to meet the requirements w/o having to manually edit the schema .csv. This seems like something that could come flummox our future selves or our future counterparts.

bsweger:
@jworcestBAH Done reviewing. I'm game to merge now, so we can proceed with the accelerated deploy schedule.

Would be great to remove that hard-coded dictionary of file types to model names if we can, but not necessary to merge. Ditto the conversation about the best way to skip the basic validations for the D files. Is there a way we can bypass those without change the schema itself...I'm worried about keeping everything in sync when the schema is updated.

dataactvalidator/config/awardProcurementFields.csv
jworcestBAH: Yes, this is the same change that was made to D2, this makes all fields optional strings with no length limits.  Ideally an updated version of the IDD would include these changes as well.  Alternatively, we could refactor schemaLoader to overwrite these values for award_financial_assistance and award_procurement with (false, str, None) as it loads each field.  I'll add a story to JIRA proposing that change.

dataactvalidator/interfaces/validatorStagingInterface.py
```diff
@@ -1,12 +1,12 @@
...
-from dataactcore.models.stagingModels import Appropriation, ObjectClassProgramActivity, AwardFinancial, AwardFinancialAssistance
+from dataactcore.models.stagingModels import Appropriation, ObjectClassProgramActivity, AwardFinancial, AwardFinancialAssistance, AwardProcurement
 from dataactcore.utils.responseException import ResponseException
 from dataactcore.utils.statusCode import StatusCode
 
 
 class ValidatorStagingInterface(ValidationInterface):
     """Manages all interaction with the staging tables in the validation database."""
-    MODEL_MAP = {"award":AwardFinancialAssistance,"award_financial":AwardFinancial,"appropriations":Appropriation,"program_activity":ObjectClassProgramActivity}
+    MODEL_MAP = {"award":AwardFinancialAssistance,"award_financial":AwardFinancial,"appropriations":Appropriation,"program_activity":ObjectClassProgramActivity,"award_procurement":AwardProcurement}
```
jworcestBAH: I did some experimenting with this and got this mapping tablenames to models:
```
MODEL_MAP = {model[1].__tablename__:model[1] for model in inspect.getmembers(stagingModels, inspect.isclass)
                 if str(model[1].__module__) == "dataactcore.models.stagingModels"}
```

However, it turns out the labels being used here do not all correspond to the table name, the mismatches are:
award->award_financial_assistance
appropriations->appropriation
program_activity->object_class_program_activity

Changing these in setupJobTrackerDB.py led to quite a few errors, so this would need a broader effort to go through and find everywhere we're specifying one of those file types.

PR #305, opened 2016-09-27
Fixing write_csv function for using s3

jworcestBAH:
This fixes the s3 writing for D files


cmc333333:
Simple enough, sorry for the problem. Merging before CI for demo

PR #308, opened 2016-09-28
Db 925 c23

jworcestBAH:
Added rule c23 with unit tests.


dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -93,3 +93,7 @@ C17,TransactionObligatedAmount required on rows that do not include values for U
...
+C23,"For each unique award ID (PIID/ParentAwardId for procurement, FAIN/URI for financial assistance) in File C, the sum of each TransactionObligatedAmount submitted in the reporting period should match (in inverse) the sum of the FederalActionObligation (in either D1 or D2) or OriginalLoanSubsidyCost (only in D2) amounts reported in D1 or D2 for the same timeframe, regardless of modifications. ","For each unique award ID (PIID/ParentAwardId for procurement, FAIN/URI for financial assistance) in File C, the sum of each TransactionObligatedAmount submitted in the reporting period should match (in inverse) the sum of the FederalActionObligation (in either D1 or D2) or OriginalLoanSubsidyCost (only in D2) amounts reported in D1 or D2 for the same timeframe, regardless of modifications. ",y,award_financial,warning,award_procurement,c23_award_financial_4
```
cmc333333: Since the rule is split into four checks, should the descriptions of the check explain that "PIID failed" vs "FAIN failed" vs ...? If not, the user could receive multiple "different" errors with the same error message, hey?

dataactvalidator/config/sqlrules/c23_award_financial_1.sql
```diff
@@ -0,0 +1,18 @@
...
+JOIN award_procurement AS ap
+		ON af.piid = ap.piid
+	  AND af.submission_id = ap.submission_id
+WHERE af.submission_id = {0}
+GROUP BY af.piid
+HAVING
+		(SELECT COALESCE(SUM(sub_af.transaction_obligated_amou::numeric),0) AS transaction_sum
+		FROM award_financial as sub_af WHERE submission_id = {0} AND sub_af.piid = af.piid) <>
+		-1*(SELECT COALESCE(SUM(sub_ap.federal_action_obligation::numeric),0) AS obligation_sum
+		FROM award_procurement as sub_ap WHERE submission_id = {0} AND sub_ap.piid = af.piid)
```
cmc333333: I think this `HAVING` clause can be simplified by using the aliases above:
```sql
HAVING transaction_obligated_amou_sum <> -1 * federal_action_obligation_sum
```

tests/unit/dataactvalidator/test_c23_award_financial_1.py
```diff
@@ -0,0 +1,51 @@
...
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
+
+def test_success(database):
+    """ Test that a four digit object class with no flag is a success, and a three digit object class with a flag is a success"""
+    # Create a 12 character random piid
+    piid = ''.join(choice(ascii_uppercase + ascii_lowercase + digits) for i in range(12))
+    piid_two = ''.join(choice(ascii_uppercase + ascii_lowercase + digits) for i in range(12))
+    first_piid_row_one = AwardFinancialFactory(object_class = randint(1100,1999),
+        by_direct_reimbursable_fun = "d", transaction_obligated_amou = 1234, piid = piid)
```
cmc333333: Are the `object_class` and `by_direct_reimbursable_fun` fields relevant to the test? If not, can we use the defaults? The less readers have to think about when reading the test, the easier they can understand it

tests/unit/dataactvalidator/test_c23_award_financial_1.py
```diff
@@ -0,0 +1,51 @@
...
+    first_piid_row_two = AwardFinancialFactory(object_class = randint(1000,1099),
+        by_direct_reimbursable_fun = "d", transaction_obligated_amou = 2345, piid = piid)
+    # And add a row for a different piid
+    second_piid_row_one = AwardFinancialFactory(object_class = randint(1100,1999),
+        by_direct_reimbursable_fun = "d", transaction_obligated_amou = 9999, piid = piid_two)
+    first_ap_row = AwardProcurementFactory(piid = piid, federal_action_obligation = -2468)
+    second_ap_row = AwardProcurementFactory(piid = piid, federal_action_obligation = -1100)
+    third_ap_row = AwardProcurementFactory(piid = piid, federal_action_obligation = -11)
```
cmc333333: It took me a bit to realize that 1234+2345 = 2468 + 1100 + 11. Not a big deal, but it there a way to make the equality more obvious? Tweaking the numbers may help, but even something simple like:
```python
# first_piid and three aps amounts are equal:
assert 1234 + 2345 = 2468 + 1100 + 11
```
may help

tests/unit/dataactvalidator/test_c23_award_financial_2.py
```diff
@@ -0,0 +1,51 @@
...
+        by_direct_reimbursable_fun = "d", transaction_obligated_amou = 2345, parent_award_id = parent_award_id)
+    # And add a row that shouldn't be included
+    second_parent_award_id_row_one = AwardFinancialFactory(object_class = randint(1100,1999),
+        by_direct_reimbursable_fun = "d", transaction_obligated_amou = 9999, parent_award_id = parent_award_id_two)
+    first_ap_row = AwardProcurementFactory(parent_award_id = parent_award_id, federal_action_obligation = -2468)
+    second_ap_row = AwardProcurementFactory(parent_award_id = parent_award_id, federal_action_obligation = -1000)
+    other_parent_award_id_ap_row = AwardProcurementFactory(parent_award_id = parent_award_id_two, federal_action_obligation = -1111)
+
+    errors = number_of_errors(_FILE, database, models=[first_parent_award_id_row_one, first_parent_award_id_row_two, second_parent_award_id_row_one, first_ap_row, second_ap_row, other_parent_award_id_ap_row])
+    assert errors == 2
```
cmc333333: Another small thing: given how similar the first two test files are (they vary only in field name), it might be worth considering combining them. Keeping legibility high would be the challenge with this approach.

dataactvalidator/config/sqlrules/c23_award_financial_1.sql
jworcestBAH: I tried this, but HAVING actually happens before SELECT, so this didn't work (http://www.bennadel.com/blog/70-sql-query-order-of-operations.htm).  Also tried putting the alias in the having clause, but it turns out that's not a thing.

tests/unit/dataactvalidator/test_c23_award_financial_1.py
jworcestBAH: Removed these.  I had them there with separate ranges because a repeat of object class, program activity, that flag, and award ID is an error, but since that error is defined in a different rule it shouldn't affect this test, so no reason to specify them here.

tests/unit/dataactvalidator/test_c23_award_financial_2.py
jworcestBAH: I thought about this, but I wasn't sure where to put the common code in the new pytest setup.  Adding a function in utils.py to be used twice seems the wrong approach, would it be reasonable to import a function from test_c23_award_financial_1 into the _2 file (and the same from _3 to _4)?  Alternatively I could have a c23_utils module that stores the common functions.

cmc333333: I agree that adding the function to utils.py would make things more messy. What do you think about combining the files into, e.g. `test_c23_award_financial_1_2`. Then the utility function would live in the same file as its tests. Again, though, not a big deal -- tests are always a bit messier than app code

dataactvalidator/config/sqlrules/c23_award_financial_1.sql
cmc333333: Sad pandas, but understood

tests/unit/dataactvalidator/test_c23_award_financial_1.py
jworcestBAH: Changed the numbers being used to be clearer, now it's 1100 + 11 compared to 1100 + 10 + 1

dataactvalidator/config/sqlrules/sqlRules.csv
jworcestBAH: Revised the error messages to be more specific to each case

bsweger:
@jworcestBAH I haven't looked closely, but does this implementation of the rule include the C24 caveat (i.e., we shouldn't generate these warnings for C records that have a valid _allocation_transfer_agency_ value?

jworcestBAH:
I brought this up in standup today, Kristine and I are going to talk this afternoon about how C24 might be applied to this.  The issue is that C24 is a check on a single row, while C23 is checking sums across the entire file.  Leaving out rows with an ATA might make the sum fail when it shouldn't, or possibly we could skip C23 if any lines for that award ID have an ATA?

cmc333333:
This is in a good state to merge now and we'll follow up with the C24 exception once the business logic's resolved. These unit tests are getting better and better; nice work @jworcestBAH

PR #309, opened 2016-09-29
Added C24 exception to C23

jworcestBAH:
Added clause to exclude award IDs that have an ATA


cmc333333:
Very complicated logic, but after discussing some scenarios w/ @jworcestBAH I'm convinced. Nice work

PR #310, opened 2016-09-30
Adding null piid check and using IS NOT DISTINCE FROM for null equality checks

nmonga91:


jworcestBAH:
Okay, looks good, will merge once Jenkins finishes

PR #315, opened 2016-09-30
Changed TAS loader to no longer ignore the first line

jworcestBAH:
Test file also changed to match new format without top line


jworcestBAH:
[run tests]

PR #324, opened 2016-10-03
Changed A34 to warning

jworcestBAH:


bsweger:
@jworcestBAH thanks! looks good--will merge when Jenkins is done

PR #327, opened 2016-10-03
Staging

jworcestBAH:
Staging up to master for A34 fix


bsweger:
Look good--thanks for taking care of this, @jworcestBAH!

PR #328, opened 2016-10-03
Fixed B18 test to fill object classes up to three digits

jworcestBAH:
This fixes object class handling for B18 to avoid failures that were occurring for object classes under 100


cmc333333:
:+1: This'll help reduce those random errors

PR #329, opened 2016-10-03
Clear database after each unittest

cmc333333:
The `database` fixture will now cycle through each of the model types and
delete any present in the database, meaning that every test gets a fresh
start. This adds a bit of overhead (2s on 10s of tests); we might be able to
target it further in the future


bsweger:
Looks good--thanks for tracking this one down!

PR #330, opened 2016-10-03
Create a file with lookup values used by the broker

bsweger:
This is a companion to PR #303. Based on suggestions from @jworcestBAH and @cmc333333, this moves our static helper values out of the database load scripts and into a common file. This way, they can be referenced as needed, which will:

1. make the in-application surrogate lookups more straightforward than the current methods in `baseInterface.py`
2. still allow the database load scripts to populate the helper tables

If this looks good, I'll cleanup up the integration test code in #303 to use these constants instead of dynamically populating dictionaries from the database (see: https://github.com/fedspendingtransparency/data-act-broker-backend/pull/303#discussion_r81338600)

### DIFF
--- a/dataactcore/models/lookups.py
+++ b/dataactcore/models/lookups.py
@@ -1,0 +1,90 @@
+# This file defines a series of constants that represent the values used in the
+# broker's "helper" tables. Rather than define the values in the db setup scripts
+# and then make db calls to lookup the surrogate keys, we'll define everything
+# here, in a file that can be used by the db setup scripts *and* the application
+# code.
+# todo: replace getIdFromDict and getNameFromDict baseInterface functions with these constants
+
+from collections import namedtuple
+
+LookupType = namedtuple('LookupType', ['id', 'name', 'desc'])
+LookupFileType = namedtuple('LookupFileType', ['id', 'name', 'desc', 'letter', 'order'])
+
+FILE_STATUS = []
+FILE_STATUS.append(LookupType(1, 'complete', 'File has been processed'))
+FILE_STATUS.append(LookupType(2, 'header_error', 'The file has errors in the header row'))
+FILE_STATUS.append(LookupType(3, 'unknown_error', 'An unknown error has occurred with this file'))
+FILE_STATUS.append(LookupType(4, 'single_row_error', 'Error occurred in job manager'))
+FILE_STATUS.append(LookupType(5, 'job_error', 'File has not yet been validated'))
+FILE_STATUS.append(LookupType(6, 'incomplete', 'File has not yet been validated'))
+FILE_STATUS_DICT = {item.id: item.name for item in FILE_STATUS}
+
+ERROR_TYPE = []
+ERROR_TYPE.append(LookupType(1, 'type_error', 'The value provided was of the wrong type'))
+ERROR_TYPE.append(LookupType(2, 'required_error', 'A required value was not provided'))
+ERROR_TYPE.append(LookupType(3, 'value_error', 'The value provided was invalid'))
+ERROR_TYPE.append(LookupType(4, 'read_error', 'Could not parse this record correctly'))
+ERROR_TYPE.append(LookupType(5, 'write_error', 'Could not write this record into the staging table'))
+ERROR_TYPE.append(LookupType(6, 'rule_failed', 'A rule failed for this value'))
+ERROR_TYPE.append(LookupType(7, 'length_error', 'Value was longer than allowed length'))
+ERROR_TYPE_DICT = {item.id: item.name for item in ERROR_TYPE}
+
+JOB_STATUS = []
+JOB_STATUS.append(LookupType(1, 'waiting', 'check dependency table'))
+JOB_STATUS.append(LookupType(2, 'ready', 'can be assigned'))
+JOB_STATUS.append(LookupType(3, 'running', 'job is currently in progress'))
+JOB_STATUS.append(LookupType(4, 'finished', 'job is complete'))
+JOB_STATUS.append(LookupType(5, 'invalid', 'job is invalid'))
+JOB_STATUS.append(LookupType(6, 'failed', 'job failed to complete'))
+JOB_STATUS_DICT = {item.id: item.name for item in JOB_STATUS}
+
+JOB_TYPE = []
+JOB_TYPE.append(LookupType(1, 'file_upload', 'file must be uploaded to S3'))
+JOB_TYPE.append(LookupType(2, 'csv_record_validation', 'do record level validation and add to staging table'))
+JOB_TYPE.append(LookupType(3, 'db_transfer', 'information must be moved from production DB to staging table'))
+JOB_TYPE.append(LookupType(4, 'validation', 'new information must be validated'))
+JOB_TYPE.append(LookupType(5, 'external_validation', 'new information must be validated against external sources'))
+JOB_TYPE_DICT = {item.id: item.name for item in JOB_TYPE}
+
+PUBLISH_STATUS = []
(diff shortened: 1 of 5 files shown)


dataactcore/models/lookups.py
```diff
@@ -0,0 +1,90 @@
...
+LookupType = namedtuple('LookupType', ['id', 'name', 'desc'])
+LookupFileType = namedtuple('LookupFileType', ['id', 'name', 'desc', 'letter', 'order'])
+
+FILE_STATUS = []
+FILE_STATUS.append(LookupType(1, 'complete', 'File has been processed'))
+FILE_STATUS.append(LookupType(2, 'header_error', 'The file has errors in the header row'))
+FILE_STATUS.append(LookupType(3, 'unknown_error', 'An unknown error has occurred with this file'))
+FILE_STATUS.append(LookupType(4, 'single_row_error', 'Error occurred in job manager'))
+FILE_STATUS.append(LookupType(5, 'job_error', 'File has not yet been validated'))
+FILE_STATUS.append(LookupType(6, 'incomplete', 'File has not yet been validated'))
```
cmc333333: Obviously no big deal, but just wondering, why use a bunch of `append`s as opposed to defining the list all at once?

dataactcore/scripts/setupErrorDB.py
```diff
@@ -1,38 +1,29 @@
...
-    statusList = [(1, 'complete', 'File has been processed'),
-        (2, 'header_error', 'The file has errors in the header row'),
-        (3, 'unknown_error', 'An unknown error has occurred with this file'),
-        (4, 'single_row_error', 'CSV file must have a header row and at least one record'),
-        (5, 'job_error', 'Error occurred in job manager'),
-        (6, 'incomplete', 'File has not yet been validated')]
-    for s in statusList:
-        status = FileStatus(file_status_id=s[0], name=s[1], description=s[2])
+    for s in lookups.FILE_STATUS:
+        status = FileStatus(file_status_id=s.id, name=s.name, description=s.desc)
```
cmc333333: I'm not sure if this would be more readable, but I wanted to throw it out in case you liked it:
```python
for status_attrs in lookups.FILE_STATUS:
    status = FileStatus(*status_attrs)
```

dataactcore/scripts/setupValidationDB.py
```diff
@@ -1,47 +1,34 @@
...
-        (4, 'award', 'award file',4),
-        (5, 'award_procurement', 'award procurement file', 5)
-        ]
-    for f in fileTypeList:
-        fileType = FileTypeValidation(file_id=f[0], name=f[1], description=f[2], file_order = f[3])
-        sess.merge(fileType)
+    # insert validation file type
+    # todo: combine this table with file_types table
+    for f in lookups.FILE_TYPE:
+        if f.order:
```
cmc333333: Small detail, but we probable want `if f.order is not None`, lest we add zero at some point and get confused.

dataactcore/models/lookups.py
```diff
@@ -0,0 +1,90 @@
...
+# broker's "helper" tables. Rather than define the values in the db setup scripts
+# and then make db calls to lookup the surrogate keys, we'll define everything
+# here, in a file that can be used by the db setup scripts *and* the application
+# code.
+# todo: replace getIdFromDict and getNameFromDict baseInterface functions with these constants
+
+from collections import namedtuple
+
+LookupType = namedtuple('LookupType', ['id', 'name', 'desc'])
+LookupFileType = namedtuple('LookupFileType', ['id', 'name', 'desc', 'letter', 'order'])
```
jworcestBAH: Thanks for adding this file, it's nice to have all these values in one location.  Minor question on FILE_TYPE, I'm wondering about the purpose of the "order" element here, it doesn't seem to be used when loading the file_type table.  Related question, if we are adding an explicit ordering, should D1 be before D2 in that ordering?

dataactcore/scripts/setupValidationDB.py
jworcestBAH: This would seem to prevent E and F from being added?  Is that the intention or am I missing something here?

dataactcore/models/lookups.py
bsweger: This bit of confusion goes back to the fact that we have two file type tables (an artifact from the days of multiple dbs). I'm using `LookupFileType` as a way to load both (`file_type` and `file_type_validation`) rather than pull the thread or merging those (which we should do at some point).

dataactcore/scripts/setupValidationDB.py
bsweger: E and F aren't in the current `file_type_validation` table, so I didn't load them here. Not fully understanding the ramifications of messing with the two file type tables, I tried to leave them as is

dataactcore/models/lookups.py
bsweger: Oh, no good reason at all. I'll tweak that.

dataactcore/scripts/setupErrorDB.py
bsweger: I like that very much, and thanks for pointing it out. Didn't even think of that as I went through and updated the existing code.

dataactcore/models/lookups.py
jworcestBAH: Ah okay that makes sense, agreed that long-term we should merge these two tables.

dataactcore/scripts/setupErrorDB.py
bsweger: Well, except that the id columns are named differently in the various models. Will keep this as is.

bsweger:
@jworcestBAH and @cmc333333 thanks both for your comments--made some of the suggested changes.

dataactcore/scripts/setupErrorDB.py
cmc333333: I don't think field names would be a problem as that only uses a single star, but field _order_ would be.

bsweger: Yeah, I would have thought so too. I'm sure there's a good explanation for the behavior I'm seeing below, but I'm also fine leaving this as with the explicit column names and frying some other fish.

```python
class JobStatus(Base):
    __tablename__ = "job_status"
    JOB_STATUS_DICT = None

    job_status_id = Column(Integer, primary_key=True)
    name = Column(Text)
    description = Column(Text)
```
```
In [2]: from dataactcore.models.jobTrackerInterface import JobStatus

In [3]: from dataactcore.models.lookups import JOB_STATUS

In [4]: JOB_STATUS
Out[4]: 
[LookupType(id=1, name='waiting', desc='check dependency table'),
 LookupType(id=2, name='ready', desc='can be assigned'),
 LookupType(id=3, name='running', desc='job is currently in progress'),
 LookupType(id=4, name='finished', desc='job is complete'),
 LookupType(id=5, name='invalid', desc='job is invalid'),
 LookupType(id=6, name='failed', desc='job failed to complete')]

In [5]: test = JobStatus(*JOB_STATUS[0])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-d1f5210372ed> in <module>()
----> 1 test = JobStatus(*JOB_STATUS[0])

TypeError: __init__() takes 1 positional argument but 4 were given
```

cmc333333: D'oh! That makes sense, the constructor's definition is probably something like:
```python
def __init__(self, **kwargs):
```
Sorry for not testing this first!

bsweger: Please, no apology necessary. 'Twas a good thought!

PR #331, opened 2016-10-03
File E generation

cmc333333:
Functions to grab file E data from the SAM SOAP endpoints, convert that into the fields we care about, add an asynchronous task for dealing with this processing in the job queue, and connecting all of that to the HTTP `fileHandler` endpoint.

This'll probably be review changeset by changeset.

### DIFF
--- a/dataactcore/utils/fileE.py
+++ b/dataactcore/utils/fileE.py
@@ -1,0 +1,101 @@
+from collections import namedtuple
+import logging
+from operator import attrgetter
+
+from suds.client import Client
+
+from dataactcore.config import CONFIG_BROKER
+
+
+logger = logging.getLogger(__name__)
+
+
+def configValid():
+    """Does the config have the necessary bits for talking to the SAM SOAP
+    API"""
+    sam = CONFIG_BROKER.get('sam') or {}
+    hasWsdl = bool(sam.get('wsdl'))
+    hasUser = bool(sam.get('username'))
+    hasPass = bool(sam.get('password'))
+    return hasWsdl and hasUser and hasPass
+
+
+def createAuth(client):
+    auth = client.factory.create('userAuthenticationKeyType')
+    auth.userID = CONFIG_BROKER['sam']['username']
+    auth.password = CONFIG_BROKER['sam']['password']
+    return auth
+
+
+def createSearch(client, duns):
+    search = client.factory.create('entitySearchCriteriaType')
+    search.DUNSList = client.factory.create('DUNSList')
+    search.DUNSList.DUNSNumber = duns
+    return search
+
+
+def getEntities(client, duns):
+    """Hit the SAM SOAP API, searching for the provided DUNS numbers. Return
+    the results as a list of Suds objects"""
+    params = client.factory.create('requestedData')
+    params.coreData.value = 'Y'
+
+    result = client.service.getEntities(
+        createAuth(client), createSearch(client, duns), params)
+
+    if result.listOfEntities:
+        return result.listOfEntities.entity
+    else:
+        return []
+
+
+Row = namedtuple('Row', (
+    'AwardeeOrRecipientUniqueIdentifier',
+    'UltimateParentUniqueIdentifier',
+    'UltimateParentLegalEntityName',
+    'HighCompOfficer1Name',
+    'HighCompOfficer1Amount',
+    'HighCompOfficer2Name',
+    'HighCompOfficer2Amount',
+    'HighCompOfficer3Name',
+    'HighCompOfficer3Amount',
+    'HighCompOfficer4Name',
+    'HighCompOfficer4Amount',
+    'HighCompOfficer5Name',
+    'HighCompOfficer5Amount'))
+
+
+def sudsToRow(sudsObj):
+    """Convert a Suds result object into a Row tuple. This accounts for the
+    presence/absence of top-paid officers"""
+    comp = sudsObj.coreData.listOfExecutiveCompensationInformation
+    officers = []
+    officer_data = getattr(comp, 'executiveCompensationDetail', [])
+    officer_data = sorted(officer_data, key=attrgetter('compensation'),
+                          reverse=True)
+    for data in officer_data:
+        officers.append(data.name)
+        officers.append(data.compensation)
+
+    # Only top 5
+    officers = officers[:10]
+    # Fill in any blanks
+    officers.extend([''] * (10 - len(officers)))
+
+    return Row(
+        sudsObj.entityIdentification.DUNS,
+        sudsObj.coreData.DUNSInformation.globalParentDUNS.DUNSNumber,
+        sudsObj.coreData.DUNSInformation.globalParentDUNS.legalBusinessName,
+        *officers
+    )
+
+
+def retrieveRows(duns):
+    """Soup-to-nuts creates a list of Row tuples from a set of DUNS
+    numbers."""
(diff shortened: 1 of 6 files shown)


dataactcore/utils/jobQueue.py
```diff
@@ -42,27 +45,16 @@ def enqueue(jobID):
...
-            header = [key for key in fileF.mappings]    # keep order
-            body = []
-            for row in rows_of_dicts:
-                body.append([row[key] for key in header])
-
-            write_csv(timestamped_name, upload_file_name, is_local, header,
-                      body)
+            yield job_manager
             job_manager.markJobStatus(job_id, "finished")
         except Exception as e:
```
jworcestBAH: After doing a bit of searching on this I'm still not quite sure, will exceptions thrown inside a with block using this context manager land here?  All the examples I'm seeing are putting the error handling outside the with block.  We only want to mark the job as "finished" rather than "failed" if the associated generate function completes without throwing an error, and ideally any exceptions being thrown will hit the below line that marks the job as "failed" so the frontend shows an error rather than hanging.

dataactcore/utils/jobQueue.py
```diff
@@ -73,5 +65,40 @@ def generate_f_file(submission_id, job_id, interface_holder_class,
...
+def generate_e_file(submission_id, job_id, interface_holder_class,
+                    timestamped_name, upload_file_name, is_local):
+    """Write file E to an appropriate CSV. See generate_file_file for an
+    explanation of interface_holder_class"""
+    with job_context(interface_holder_class, job_id) as job_manager:
+        results = job_manager.session.\
+            query(AwardProcurement.awardee_or_recipient_uniqu).\
+            filter(AwardProcurement.submission_id == submission_id).\
+            distinct()
+        duns = [r.awardee_or_recipient_uniqu for r in results]
```
jworcestBAH: I believe we may also need to pull DUNS numbers out of D2 (AwardFinancialAssistance), we should check with Kristine on this.

tests/unit/dataactcore/test_job_queue.py
```diff
@@ -31,13 +25,68 @@ def test_generate_f_file(monkeypatch, mock_broker_config_paths):
...
+         'HighCompOfficer2Name', 'HighCompOfficer2Amount',
+         'HighCompOfficer3Name', 'HighCompOfficer3Amount',
+         'HighCompOfficer4Name', 'HighCompOfficer4Amount',
+         'HighCompOfficer5Name', 'HighCompOfficer5Amount'],
+        ['a', 'b', 'c', '1a', '1b', '2a', '2b', '3a', '3b', '4a', '4b',
+         '5a', '5b'],
+        ['A', 'B', 'C', '1A', '1B', '2A', '2B', '3A', '3B', '4A', '4B',
+         '5A', '5B']
+    ]
+    assert read_file_rows(file_path) == expected
```
jworcestBAH: I'm a bit confused about what's being tested on this line.  It looks like we're mocking the result of the entire file E code, and then checking that the mock result is written to the csv, so the file writing and the Row tuple are the parts that are tested here?  I get that we don't want to test by calling the soap API, but perhaps we could mock something lower like getEntities so that the retrieveRows functionality gets tested?  Alternatively we could add a retrieveRows test to the test_utils_fileE below with the other fileE.py tests.

tests/unit/dataactcore/test_utils_fileE.py
```diff
@@ -0,0 +1,94 @@
...
+        Mock(compensation=i*11.11) for i in range(1, 10)]
+    for idx, person in enumerate(info.executiveCompensationDetail):
+        # Can't do this in the constructor as "name" is a reserved word
+        person.name = "Person {}".format(idx + 1)
+
+    row = fileE.sudsToRow(sudsObj)
+    assert row == fileE.Row(
+        'B Duns', 'Par DunsB', 'Par NameB',
+        'Person 9', 99.99, 'Person 8', 88.88, 'Person 7', 77.77,
+        'Person 6', 66.66, 'Person 5', 55.55)
```
jworcestBAH: Nice to have these, some very thorough testing going on here.  I'm assuming that for createAuth, createSearch, and getEntities we don't want to mock suds.client.Client in order to test these?

dataactcore/utils/fileE.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def getEntities(client, duns):
+    """Hit the SAM SOAP API, searching for the provided DUNS numbers. Return
+    the results as a list of Suds objects"""
+    params = client.factory.create('requestedData')
+    params.coreData.value = 'Y'
+
+    result = client.service.getEntities(
+        createAuth(client), createSearch(client, duns), params)
```
cmc333333: It does, yes; the `result` contains a sequence of of results in its `listOfEntities` field.

dataactcore/utils/fileE.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if result.listOfEntities:
+        return result.listOfEntities.entity
```
cmc333333: Indeed, this is the same as line 33 -- `entity` is a sequence.

dataactcore/utils/fileE.py
```diff
@@ -0,0 +1,101 @@
...
+    else:
+        return []
+
+
+Row = namedtuple('Row', (
+    'AwardeeOrRecipientUniqueIdentifier',
+    'UltimateParentUniqueIdentifier',
+    'UltimateParentLegalEntityName',
+    'HighCompOfficer1Name',
```
cmc333333: That's been discussed out of band; the IDD's going to be changed as the web service doesn't break up names that way.

dataactcore/utils/jobQueue.py
cmc333333: Yes, the exception will percolate up. Here's a quick demonstration:
```python
from contextlib import contextmanager


@contextmanager
def catch_anything():
    try:
        yield
    except Exception:
        print("Caught inside")


with catch_anything():
    int("not-an-int")

try:
    int("not-an-int")
except Exception:
    print("Caught outside")
```

tests/unit/dataactcore/test_job_queue.py
cmc333333: Indeed, this test is only meant to cover `generate_e_file`, which generates a CSV based on `fileE.Row`s. I'll see if there's not some more useful tests to write at a lower level, though.

tests/unit/dataactcore/test_utils_fileE.py
cmc333333: Unfortunately, mocking out suds responses won't do us much good -- they're dynamically generated based on the WSDL, so we'd basically be passing in a `Mock` with a few fields set and then verifying that those fields were set. I'll try out a bigger integration test, though, so we can verify that at least the parts work together (even if suds does not).

dataactcore/utils/jobQueue.py
cmc333333: Good call! Kristine says the D2 DUNS should indeed be included.

cmc333333:
@jworcestBAH updated, thanks for the review; I think the changes address everything you mentioned. Re testing: this works locally after including fedspendingtransparency/data-act-broker-web-app#316 and making fake D1/D2 calls, but lmk if anything looks suspicious on the sandbox.

jworcestBAH:
Changes look good, merging now.

PR #333, opened 2016-10-04
Db 1104 first nine

jworcestBAH:
This adds unit tests for many of the rules that did not have them


tests/unit/dataactvalidator/test_c5_award_financial_1.py
```diff
@@ -0,0 +1,38 @@
...
+    expected_subset = {'row_number', "gross_outlay_amount_by_awa_cpe", "gross_outlays_undelivered_cpe",
+	    "gross_outlays_delivered_or_cpe"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c5_award_financial_1.py
```diff
@@ -0,0 +1,38 @@
...
+    value_two = Decimal(randint(0,100000)) / Decimal(100)
+    award_fin = AwardFinancialFactory(gross_outlay_amount_by_awa_cpe = value_one+value_two,
+                                             gross_outlays_undelivered_cpe = value_one,
+                                             gross_outlays_delivered_or_cpe = value_two)
+    award_fin_null = AwardFinancialFactory(gross_outlay_amount_by_awa_cpe = value_one,
+                                                  gross_outlays_undelivered_cpe = None,
```
bsweger: As always, thanks for including the NULL case!

tests/unit/dataactvalidator/test_c5_award_financial_2.py
```diff
@@ -0,0 +1,38 @@
...
+    expected_subset = {'row_number', "gross_outlay_amount_by_awa_fyb", "gross_outlays_undelivered_fyb",
+	    "gross_outlays_delivered_or_fyb"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c6_award_financial_1.py
```diff
@@ -0,0 +1,45 @@
...
+    expected_subset = {'row_number', "gross_outlays_undelivered_cpe", "ussgl480200_undelivered_or_cpe",
+	    "ussgl483200_undelivered_or_cpe", "ussgl487200_downward_adjus_cpe", "ussgl488200_upward_adjustm_cpe"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c6_award_financial_2.py
```diff
@@ -0,0 +1,32 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number', "gross_outlays_undelivered_fyb", "ussgl480200_undelivered_or_fyb"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c7_award_financial_1.py
```diff
@@ -0,0 +1,45 @@
...
+    expected_subset = {'row_number', "gross_outlays_delivered_or_cpe", "ussgl490200_delivered_orde_cpe",
+	    "ussgl490800_authority_outl_cpe", "ussgl497200_downward_adjus_cpe", "ussgl498200_upward_adjustm_cpe"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c7_award_financial_2.py
```diff
@@ -0,0 +1,32 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number', "gross_outlays_delivered_or_fyb", "ussgl490800_authority_outl_fyb"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test that calculation passes with equal values and with a null """
+
+    value_one = Decimal(randint(0,100000)) / Decimal(100)
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -0,0 +1,38 @@
...
+    award_fin_fain = AwardFinancialFactory(uri = None)
+    award_fin_uri = AwardFinancialFactory(fain = None)
+    afa_fain = AwardFinancialAssistanceFactory(fain = award_fin_fain.fain, uri = None, federal_action_obligation = randint(1,1000))
+    afa_uri = AwardFinancialAssistanceFactory(fain = None, uri = award_fin_uri.uri, federal_action_obligation = randint(1,1000))
+    # Test with both fain and uri
+    award_fin = AwardFinancialFactory()
+    afa = AwardFinancialAssistanceFactory(fain = award_fin.fain, uri = award_fin.uri, federal_action_obligation = randint(1,1000))
+    # Test with zero federal action obligation
+    afa_zero = AwardFinancialAssistanceFactory(federal_action_obligation = 0)
+
```
bsweger: This might be another case where we missed a rule revision, but do we also need to check that the original loan subsidy cost = 0 in addition to the federal action obligation = 0?

Will ask about this one in Slack too.

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -0,0 +1,38 @@
...
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test case where all fain and uri in D2 are present in C """
+
+    # Test with only one of fain/uri
+    award_fin_fain = AwardFinancialFactory(uri = None)
+    award_fin_uri = AwardFinancialFactory(fain = None)
+    afa_fain = AwardFinancialAssistanceFactory(fain = award_fin_fain.fain, uri = None, federal_action_obligation = randint(1,1000))
```
bsweger: Same question re: randomly-generated numbers vs. using an explicit value.

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -0,0 +1,38 @@
...
+
+def test_failure(database):
+    """ Test fain and uri present in D2 but not in C """
+    award_fin = AwardFinancialFactory(fain = None, uri = None)
+    afa_fain = AwardFinancialAssistanceFactory(uri = None, federal_action_obligation = randint(1,1000))
+    afa_uri = AwardFinancialAssistanceFactory(fain = None, federal_action_obligation = randint(1,1000))
+    # Test with both fain and uri
+    afa = AwardFinancialAssistanceFactory(federal_action_obligation = randint(1,1000))
+
+    assert number_of_errors(_FILE, database, models=[award_fin, afa_fain, afa_uri, afa]) == 3
```
bsweger: Nice roundup of failure cases here!

tests/unit/dataactvalidator/utils.py
```diff
@@ -11,7 +11,7 @@ def insert_submission(db, submission):
     return submission.submission_id
 
 
-def error_rows(rule_file, staging_db, submission=None, models=None):
+def error_rows(rule_file, staging_db, submission=None, models=None, assert_num = None):
```
bsweger: I don't remember seeing the new `assert_num` parameter being used anywhere--do we still need this?

tests/unit/dataactvalidator/utils.py
```diff
@@ -39,8 +42,8 @@ def error_rows(rule_file, staging_db, submission=None, models=None):
     return result
 
 
-def number_of_errors(rule_file, staging_db, submission=None, models=None):
-    return len(error_rows(rule_file, staging_db, submission, models))
+def number_of_errors(rule_file, staging_db, submission=None, models=None, assert_num = None):
```
bsweger: Same question here re: `assert_num` parameter.

dataactvalidator/config/sqlrules/c14_award_financial.sql
```diff
@@ -1,4 +1,4 @@
-SELECT row_number, fain, uri
+SELECT row_number, fain, uri, piid
```
bsweger: Very nice catch!

dataactvalidator/config/sqlrules/c17_award_financial.sql
```diff
@@ -2,23 +2,23 @@ SELECT row_number, transaction_obligated_amou
...
-		AND ussgl490200_delivered_orde_cpe IS NULL
-		AND ussgl490800_authority_outl_cpe IS NULL
-		AND ussgl490800_authority_outl_fyb IS NULL
-		AND ussgl493100_delivered_orde_cpe IS NULL
-		AND ussgl497100_downward_adjus_cpe IS NULL
-		AND ussgl497200_downward_adjus_cpe IS NULL
-		AND ussgl498100_upward_adjustm_cpe IS NULL
-		AND ussgl498200_upward_adjustm_cpe IS NULL);
\ No newline at end of file
+	AND (COALESCE(ussgl480100_undelivered_or_cpe,0) = 0
```
bsweger: I know for files like SF-133, we're supposed to interpret NULL as 0. Just confirming that's the case here as well. If so, good catch!

tests/unit/dataactvalidator/test_c9_award_financial.py
bsweger: Nevermind! I see that C9 revisions are on the board for the current sprint, so I'll assume that this test will be updated as part of that work.

tests/unit/dataactvalidator/test_a1_appropriations.py
```diff
@@ -0,0 +1,58 @@
...
+from tests.unit.dataactcore.factories.domain import TASFactory
+from tests.unit.dataactvalidator.utils import number_of_errors, query_columns
+
+_FILE = 'a1_appropriations'
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "allocation_transfer_agency", "agency_identifier", "beginning_period_of_availa",
+        "ending_period_of_availabil", "availability_type_code", "main_account_code", "sub_account_code"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
```
jworcestBAH: I based this on the existing tests that have this test_column_headers.  That said, I haven't seen a case where I was expecting it to be a subset, so I'll go ahead and change these to test for equality.

tests/unit/dataactvalidator/test_a1_appropriations.py
```diff
@@ -0,0 +1,58 @@
...
+    tas = TASFactory()
+    tas_null = TASFactory(allocation_transfer_agency = None, availability_type_code = None, sub_account_code = None)
+    approp = AppropriationFactory(allocation_transfer_agency = tas.allocation_transfer_agency,
+                                  agency_identifier = tas.agency_identifier,
+                                  beginning_period_of_availa = tas.beginning_period_of_availability,
+                                  ending_period_of_availabil = tas.ending_period_of_availability,
+                                  availability_type_code = tas.availability_type_code,
+                                  main_account_code = tas.main_account_code,
+                                  sub_account_code = tas.sub_account_code)
+    approp_null = AppropriationFactory(allocation_transfer_agency = None,
```
jworcestBAH: Fixed

tests/unit/dataactvalidator/test_a1_award_financial.py
```diff
@@ -0,0 +1,58 @@
...
+    tas = TASFactory()
+    tas_null = TASFactory(allocation_transfer_agency = None, availability_type_code = None, sub_account_code = None)
+    approp = AwardFinancialFactory(allocation_transfer_agency = tas.allocation_transfer_agency,
+                                  agency_identifier = tas.agency_identifier,
+                                  beginning_period_of_availa = tas.beginning_period_of_availability,
+                                  ending_period_of_availabil = tas.ending_period_of_availability,
+                                  availability_type_code = tas.availability_type_code,
+                                  main_account_code = tas.main_account_code,
+                                  sub_account_code = tas.sub_account_code)
+    approp_null = AwardFinancialFactory(allocation_transfer_agency = None,
```
jworcestBAH: Good catch, fixed those

tests/unit/dataactvalidator/test_a1_object_class_program_activity.py
```diff
@@ -0,0 +1,58 @@
...
+
+_FILE = 'a1_object_class_program_activity'
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "allocation_transfer_agency", "agency_identifier", "beginning_period_of_availa",
+        "ending_period_of_availabil", "availability_type_code", "main_account_code", "sub_account_code"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
+
+def test_success(database):
```
jworcestBAH: Fixed

tests/unit/dataactvalidator/test_a20_appropriations_1.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "allocation_transfer_agency"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
+
+def test_success(database):
+    """ Test that agency codes are matched against cgac correctly """
+    approp = AppropriationFactory()
+    approp_null = AppropriationFactory(allocation_transfer_agency = None)
```
jworcestBAH: The comparison between approp and cgac is for a non-null allocation_transfer_agency, the AppropriationFactory generates a random string and cgac reuses that string.  Is that the case you're referring to or is there another case I'm not thinking of here?

tests/unit/dataactvalidator/test_a20_appropriations_1.py
```diff
@@ -0,0 +1,32 @@
...
+    cgac = CGACFactory(cgac_code = approp.allocation_transfer_agency)
+
+    errors = number_of_errors(_FILE, database, models=[approp, approp_null, cgac])
+    assert errors == 0
+
+def test_failure(database):
+    """ Test a cgac not present in cgac table """
+    # These cgacs are different lengths to avoid being equal
```
jworcestBAH: So the reason I've been leaning towards randomly generated data throughout these tests is that it eventually tests a broader range of possibilities.  If there are strange edge cases that occur for certain types of values, the random data gives us a better chance of finding those.  Does that make sense, or is the extra readability of the specific values worth dropping that?

tests/unit/dataactvalidator/test_a21_appropriation.py
```diff
@@ -0,0 +1,24 @@
...
+_FILE = 'a21_appropriations'
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "availability_type_code"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset <= actual
+
+def test_success(database):
+    """ Test that availability type code is either x or absent """
+    approp = AppropriationFactory(availability_type_code = 'X')
```
jworcestBAH: Added these cases

tests/unit/dataactvalidator/test_a21_award_financial.py
```diff
@@ -0,0 +1,24 @@
...
+    """ Test that availability type code is either x or absent """
+    approp = AwardFinancialFactory(availability_type_code = 'X')
+    approp_null = AwardFinancialFactory(availability_type_code = None)
+
+    errors = number_of_errors(_FILE, database, models=[approp, approp_null])
+    assert errors == 0
+
+def test_failure(database):
+    """ Test an incorrect availability_type_code """
+    approp = AwardFinancialFactory(availability_type_code = 'z')
```
jworcestBAH: Yep fixed here and for ocpa

tests/unit/dataactvalidator/test_c14_award_financial.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_success(database):
+    """ Test cases with different combinations of fain, uri, and piid """
+
+    # Test with only one present
+    award_fin_fain = AwardFinancialFactory(uri = None, piid = None)
+    award_fin_uri = AwardFinancialFactory(fain = None, piid = None)
+    award_fin_piid = AwardFinancialFactory(fain = None, uri = None)
+    # Test with all three
+    award_fin = AwardFinancialFactory()
```
jworcestBAH: It does not fail this rule, which just requires at least one of the three to be present.  Revisions to C8, C9, and this rule cover the possible combinations (fain+uri will be a warning, piid with either fain or uri will be an error).  I created story DB-1244 to update these three rules to include the revisions made in the Validation_Rules_For_Broker sheet.

bsweger:
@jworcestBAH FYI, don't worry about the C14 inconsistencies. We need to go back and update that rule as a separate story (it's on the board as [DB-1247](https://federal-spending-transparency.atlassian.net/browse/DB-1247)).

tests/unit/dataactvalidator/test_c14_award_financial.py
```diff
@@ -0,0 +1,33 @@
...
+    # Test with one missing
+    award_fin_no_fain = AwardFinancialFactory(fain = None)
```
jworcestBAH: When we do the updates from the validation rules sheet, I'm thinking we'll probably add a c14_award_financial_2 file that will catch that error, which would then also get its own unit test.

tests/unit/dataactvalidator/test_a1_appropriations.py
bsweger: Ah, nope--was misreading. Carry on!

tests/unit/dataactvalidator/test_a20_appropriations_1.py
bsweger: Ah, nope--was misreading. Carry on!

tests/unit/dataactvalidator/test_a20_appropriations_1.py
bsweger: If there are broader cases we need to test, I think it makes sense to explicitly test them rather than relying on randomly-generated numbers to hit the edge cases. Again, I wouldn't hold up the PR for this point, but I think it's worth considering in the larger group as we move forward with unit tests.

tests/unit/dataactvalidator/test_c17_award_financial.py
```diff
@@ -0,0 +1,149 @@
...
+                                             ussgl490100_delivered_orde_fyb = None,
+                                             ussgl490200_delivered_orde_cpe = None,
+                                             ussgl490800_authority_outl_cpe = None,
+                                             ussgl490800_authority_outl_fyb = randint(1,1000),
+                                             ussgl493100_delivered_orde_cpe = None,
+                                             ussgl497100_downward_adjus_cpe = None,
+                                             ussgl497200_downward_adjus_cpe = None,
+                                             ussgl498100_upward_adjustm_cpe = None,
+                                             ussgl498200_upward_adjustm_cpe = None)
+    award_fin_three = AwardFinancialFactory(transaction_obligated_amou = None,
```
jworcestBAH: Good point, I changed one of these cases to be 0 instead of null, the point of the multiple cases here is just to check a few different randomly chosen fields, plus one with a group of non-null fields.

tests/unit/dataactvalidator/test_c9_award_financial.py
jworcestBAH: Thanks for pointing that out, I didn't see that story.  I see you created DB-1247 for the C14 changes, so DB-1244 is no longer needed if you have the delete permission on JIRA.

dataactvalidator/config/sqlrules/c17_award_financial.sql
jworcestBAH: I asked Kristine about this after my unit test failed on the case with zeros, and she confirmed that zeros here should be considered unpopulated for this rule.

bsweger: Very cool--another great example of things we're flushing out by adding the tests.

tests/unit/dataactvalidator/test_a21_appropriation.py
```diff
@@ -6,14 +6,15 @@
 def test_column_headers(database):
     expected_subset = {"row_number", "availability_type_code"}
     actual = set(query_columns(_FILE, database))
-    assert expected_subset <= actual
+    assert expected_subset == actual
 
 def test_success(database):
     """ Test that availability type code is either x or absent """
     approp = AppropriationFactory(availability_type_code = 'X')
+    approp_lower = AppropriationFactory(availability_type_code = 'x')
```
bsweger: Thanks for adding this!

bsweger:
Thanks for the updates @jworcestBAH. This is a great addition to the code base, and I'm so happy that the very act of writing these tests surfaced some things we need to look at. The question of randomly-generated numbers isn't worth holding up the PR, so I'm going to merge.

PR #334, opened 2016-10-04
Remove db interfaces from the fileTests suite of integration tests

bsweger:
This PR completes the removal of the old db interfaces from the fileTests.py suite of integration tests.

The changes might make more sense when reviewing commit by commit.

### DIFF
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -77,8 +77,7 @@ def sumNumberOfErrorsForJobList(submissionId, errorType='fatal'):
 def checkNumberOfErrorsByJobId(jobId, errorType='fatal'):
     """Get the number of errors for a specified job and severity."""
     sess = GlobalDB.db().session
-    errors = sess.query(
-        func.sum(ErrorMetadata.occurrences)).join(
-        ErrorMetadata.severity).filter(
-        ErrorMetadata.job_id == jobId, RuleSeverity.name == errorType).scalar()
+    errors = sess.query(func.sum(ErrorMetadata.occurrences)).\
+        join(ErrorMetadata.severity).\
+        filter(ErrorMetadata.job_id == jobId, RuleSeverity.name == errorType).scalar()
     return errors or 0
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -68,7 +68,6 @@ def sumNumberOfErrorsForJobList(submissionId, errorType='fatal'):
             job.number_of_errors = jobErrors
         elif errorType == 'warning':
             job.number_of_warnings = jobErrors
-        sess.add(job)
         errorSum += jobErrors
     sess.commit()
     return errorSum
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -79,4 +79,8 @@ def checkNumberOfErrorsByJobId(jobId, errorType='fatal'):
     errors = sess.query(func.sum(ErrorMetadata.occurrences)).\
         join(ErrorMetadata.severity).\
         filter(ErrorMetadata.job_id == jobId, RuleSeverity.name == errorType).scalar()
+    # error_metadata table tallies total errors by job/file/field/error type. jobs that
+    # don't have errors or warnings won't be in the table at all. thus, if the above query
+    # returns an empty value that means the job didn't have any errors that matched
+    # the specified severity type, so return 0
     return errors or 0
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -1,6 +1,11 @@
 import uuid
 
+from sqlalchemy.sql import func
+
+from dataactcore.models.errorModels import ErrorMetadata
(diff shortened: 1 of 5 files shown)


tests/integration/fileTests.py
```diff
@@ -122,23 +120,24 @@ def test_file_submission(self):
...
-            self.assertIsInstance(jobId, int)
-            # Check that original filenames were stored in DB
-            originalFilename = self.interfaces.jobDb.getOriginalFilenameById(jobId)
-            self.assertEquals(originalFilename,self.filenames[key])
-        # check that submission got mapped to the correct user
-        submissionId = responseDict["submission_id"]
-        self.file_submission_id = submissionId
-        submission = self.interfaces.jobDb.getSubmissionById(submissionId)
+        fileKeys = ["program_activity", "award_financial", "appropriations"]
+        with createApp().app_context():
```
cmc333333: I'm not sure how to do this, but there's probably a way to call this within the class's `setUp` and `tearDown` functions (or using a py.test fixture) rather than in the various tests. It's context that kind of muddies up individual tests

tests/integration/baseTestAPI.py
```diff
@@ -167,6 +167,7 @@ def setUpClass(cls):
         cls.fileStatusDict = lookups.FILE_STATUS_DICT
         cls.ruleSeverityDict = lookups.RULE_SEVERITY_DICT
         cls.errorTypeDict = lookups.ERROR_TYPE_DICT
+        cls.publishStatusDict = lookups.PUBLISH_STATUS_DICT
```
cmc333333: Curious, why set this as a class variable? Seems simple enough to access `PUBLISH_STATUS_DICT` directly?

dataactcore/interfaces/function_bag.py
```diff
@@ -41,3 +46,39 @@ def getPasswordHash(password, bcrypt):
...
+    sess = GlobalDB.db().session
+    errorSum = 0
+    jobs = sess.query(Job).filter(Job.submission_id == submissionId).all()
+    for job in jobs:
+        jobErrors = checkNumberOfErrorsByJobId(job.job_id, errorType)
+        if errorType == 'fatal':
+            job.number_of_errors = jobErrors
+        elif errorType == 'warning':
+            job.number_of_warnings = jobErrors
+        sess.add(job)
```
bsweger: Nope--good call...will remove.

dataactcore/interfaces/function_bag.py
```diff
@@ -41,3 +46,39 @@ def getPasswordHash(password, bcrypt):
...
+
+
+def checkNumberOfErrorsByJobId(jobId, errorType='fatal'):
+    """Get the number of errors for a specified job and severity."""
+    sess = GlobalDB.db().session
+    errors = sess.query(
+        func.sum(ErrorMetadata.occurrences)).join(
+        ErrorMetadata.severity).filter(
+        ErrorMetadata.job_id == jobId, RuleSeverity.name == errorType).scalar()
+    return errors or 0
```
bsweger: Agreed. I took a quick peek at the code that writes the error messages. I'd love to look deeper, but am reluctant to get into a pattern of digging every time we hit a question like this as we recreate these functions in the `function_bag`.

It's a fair question indeed, but my preference is to table for now.

bsweger: (oh, should have said--I did remove the `or 0` and got errors, so you were right...I'm just trying to sort through which things to change at the moment)

dataactcore/models/errorInterface.py
```diff
@@ -57,14 +57,7 @@ def getFileStatusLabelByJobId(self, jobId):
...
     def checkNumberOfErrorsByJobId(self, jobId, valDb, errorType = "fatal"):
-        """ Get the total number of errors for a specified job
-
-        Args:
-            jobId: job to get errors for
-
-        Returns:
-            Number of errors for specified job
-        """
+        """Deprecated: moved to function_bag.py."""
```
bsweger: That's a good idea..will do.

dataactcore/interfaces/function_bag.py
bsweger: Dig a bit more into this. The ErrorMetadata table actually tracks error/warning count by combination of job/filename/fieldname, and error type. So jobs w/o errors and warnings won't be represented, which is why the code returns a 0 when the query doesn't find anything.

To your great point about the new `sum` function potentially breaking on NULL values--the field being summed is not nullable. I also checked the code that inserts ErrorMetadata records...it always starts counting at 1. For these reasons, the refactor should work the same way as the old code.

dataactcore/models/errorInterface.py
bsweger: Hmmm, tests failed all over the place when I did this, with sqlalchemy/postgres errors. I'm not inclined to spend time debugging when it's a temporary situation so I'd prefer to keep the duplicated code/deprecated message and revisit when doing the larger task of moving interface code into the function_bag.

bsweger:
@cmc333333 As always, thanks for the thorough review. If you're still thinking this is ok to merge, I think it's time to move on to the rest of the refactor, even though there are always more improvements to make.

cmc333333:
Yep, all's well. Nice work, @bsweger !

PR #335, opened 2016-10-05
Db 1104 file b unit tests

nmonga91:
Adding unit tests 13-18


tests/unit/dataactvalidator/test_b12_award_financial_1.py
```diff
@@ -0,0 +1,75 @@
...
+    af = AwardFinancialFactory()
+    assert number_of_errors(_FILE, database, models=[af]) == 0
+
+
+def test_failure(database):
+    """ Test for USSGL 48XX & 49XX (except 487X & 497X) if any one is provided then
+    by_direct_reimbursable_fun is empty """
+
+    af = AwardFinancialFactory(ussgl480100_undelivered_or_fyb=None, by_direct_reimbursable_fun=None)
+    assert number_of_errors(_FILE, database, models=[af]) == 1
```
cmc333333: I'm not sure that that these tests match the docstring and SQL -- should we set all of the fields _except_ one to `None` instead? Also, for tests like this (where we're just running through a bunch of different values and asserting the same thing, consider using [`parametrize`](http://docs.pytest.org/en/latest/parametrize.html)

tests/unit/dataactvalidator/test_b12_award_financial_2.py
```diff
@@ -0,0 +1,37 @@
...
+    assert number_of_errors(_FILE, database, models=[af]) == 0
+
+    af = AwardFinancialFactory(by_direct_reimbursable_fun='r')
+    assert number_of_errors(_FILE, database, models=[af]) == 0
+
+    af = AwardFinancialFactory(by_direct_reimbursable_fun='D')
+    assert number_of_errors(_FILE, database, models=[af]) == 0
+
+    af = AwardFinancialFactory(by_direct_reimbursable_fun='d')
+    assert number_of_errors(_FILE, database, models=[af]) == 0
```
cmc333333: Not sure if it matters, but would it make sense to explicitly test the empty string, too? Also, this might be another use case for `parametrize`.

tests/unit/dataactvalidator/test_b12_object_class_program_activity_1.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_success(database):
+    """ Test for USSGL 48XX & 49XX (except 487X & 497X) if any one is provided then
+    by_direct_reimbursable_fun is not empty """
+
+    op = ObjectClassProgramActivityFactory()
+    assert number_of_errors(_FILE, database, models=[op]) == 0
+
+    op = ObjectClassProgramActivityFactory(object_class=1234)
```
cmc333333: Should this be setting `by_direct_reimbursable_fun` to `None`? Otherwise, it's not only testing the 4-digit `object_class`.

tests/unit/dataactvalidator/test_b12_object_class_program_activity_1.py
```diff
@@ -0,0 +1,94 @@
...
+    assert number_of_errors(_FILE, database, models=[op]) == 0
+
+
+def test_failure(database):
+    """ Test for USSGL 48XX & 49XX (except 487X & 497X) if any one is provided then
+    by_direct_reimbursable_fun is empty """
+
+    op = ObjectClassProgramActivityFactory(ussgl480100_undelivered_or_fyb=None, by_direct_reimbursable_fun=None,
+        object_class = 123)
+    assert number_of_errors(_FILE, database, models=[op]) == 1
```
cmc333333: Similar question as in `b12_award_financial_1` -- do we need to set all of the fields except one to `None`?

tests/unit/dataactvalidator/test_b12_object_class_program_activity_2.py
```diff
@@ -0,0 +1,37 @@
...
+
+    op = ObjectClassProgramActivityFactory(by_direct_reimbursable_fun='r')
+    assert number_of_errors(_FILE, database, models=[op]) == 0
+
+    op = ObjectClassProgramActivityFactory(by_direct_reimbursable_fun='D')
+    assert number_of_errors(_FILE, database, models=[op]) == 0
+
+    op = ObjectClassProgramActivityFactory(by_direct_reimbursable_fun='d')
+    assert number_of_errors(_FILE, database, models=[op]) == 0
+
```
cmc333333: Similar question re: the empty string.

tests/unit/dataactvalidator/test_b13_object_class_program_activity.py
```diff
@@ -0,0 +1,47 @@
...
+def test_column_headers(database):
+    expected_subset = {'row_number', 'deobligations_recov_by_pro_cpe', 'ussgl487100_downward_adjus_cpe',
+                       'ussgl487200_downward_adjus_cpe', 'ussgl497100_downward_adjus_cpe',
+                       'ussgl497200_downward_adjus_cpe'}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ DeobligationsRecoveriesRefundsOfPriorYearByProgramObjectClass_CPE in File B = USSGL(4871+ 4872 + 4971 + 4972)
```
cmc333333: Good use of the docstring to spell out the rule

tests/unit/dataactvalidator/test_b9_award_financial.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # program activity name/code as null
+    af = AwardFinancialFactory(row_number=1, beginning_period_of_availa=2016, agency_identifier='test_wrong',
+                               allocation_transfer_agency='test', main_account_code='test',
+                               program_activity_name=None, program_activity_code=None)
+
+    pa = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
+                                account_number='test')
+
+    assert number_of_errors(_FILE, database, models=[af, pa]) == 1
```
cmc333333: The SQL seems to imply that we're comparing each `award_financial` with _other_ `award_financial`s -- should there be a test that includes creating multiple `award_financials` with different `row_number`s?

tests/unit/dataactvalidator/test_b9_object_class_program_activity.py
```diff
@@ -0,0 +1,71 @@
...
+    pa = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
+                                account_number='test', program_activity_name='test', program_activity_code='test')
+
+    assert number_of_errors(_FILE, database, models=[op, pa]) == 1
+
+
+    # program activity name/code as null
+    op = ObjectClassProgramActivityFactory(row_number=1, beginning_period_of_availa=2016, agency_identifier='test_wrong',
+                               allocation_transfer_agency='test', main_account_code='test',
+                               program_activity_name=None, program_activity_code=None)
```
cmc333333: Why is the name/code as null case important? I don't see anything exceptional about them in the SQL.

tests/unit/dataactvalidator/test_b9_object_class_program_activity.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # program activity name/code as null
+    op = ObjectClassProgramActivityFactory(row_number=1, beginning_period_of_availa=2016, agency_identifier='test',
+                               allocation_transfer_agency='test', main_account_code='test',
+                               program_activity_name=None, program_activity_code=None)
+
+    pa = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
+                                account_number='test')
+
+    assert number_of_errors(_FILE, database, models=[op, pa]) == 1
```
cmc333333: Same question re: do we want to test with multiple `object_class_program_activity`s?

tests/unit/dataactvalidator/test_b12_award_financial_1.py
nmonga91: I think the doc string needs to be reworded. Looking at the SQL, I can't check any of the SGL fields unless the `by_direct_reimbursable_fun` is `None`. I do like the idea of parameterization, I'll look into setting that up for tests similar to this as well.

cmc333333: If I understand what's going on, the rule is triggered when `by_direct_reimbursable_fun` is null and _any_ of a set of other fields are not null. In each of these assertions, we've got all of those fields set except for one -- I'm wondering if it should be the other way around: all of the fields are _null_ except for one.

tests/unit/dataactvalidator/test_b12_award_financial_2.py
nmonga91: Added test for empty string and also used `parameterize`

tests/unit/dataactvalidator/test_b9_object_class_program_activity.py
nmonga91: Same situation as b9 for award_financial

tests/unit/dataactvalidator/test_b9_object_class_program_activity.py
nmonga91: Good catch. I used that portion from the b9 test for `award_financial` and is not necessary for `object_class_program_activity`. It has been removed.

nmonga91:
@cmc333333 I've made updates based on our discussion and your feedback. Please let me know if there's any additional changes that need to be accounted for.

tests/unit/dataactvalidator/test_b12_award_financial_1.py
```diff
@@ -18,58 +18,52 @@ def test_success(database):
...
+               'ussgl483200_undelivered_or_cpe': None, 'ussgl488200_upward_adjustm_cpe': None,
+               'ussgl490200_delivered_orde_cpe': None, 'ussgl490800_authority_outl_fyb': None,
+               'ussgl490800_authority_outl_cpe': None, 'ussgl498200_upward_adjustm_cpe': None}
+
+    keys = ['ussgl480100_undelivered_or_fyb', 'ussgl480100_undelivered_or_cpe', 'ussgl483100_undelivered_or_cpe',
+            'ussgl488100_upward_adjustm_cpe', 'ussgl490100_delivered_orde_fyb', 'ussgl490100_delivered_orde_cpe',
+            'ussgl493100_delivered_orde_cpe', 'ussgl498100_upward_adjustm_cpe', 'ussgl498100_upward_adjustm_cpe',
+            'ussgl480200_undelivered_or_fyb', 'ussgl480200_undelivered_or_cpe', 'ussgl483200_undelivered_or_cpe',
+            'ussgl488200_upward_adjustm_cpe', 'ussgl490200_delivered_orde_cpe', 'ussgl490800_authority_outl_fyb',
+            'ussgl490800_authority_outl_cpe', 'ussgl498200_upward_adjustm_cpe']
```
cmc333333: NBD, but you could have created keys from the dictionary:
```python
keys = af_dict.keys()
```
or a dictionary from the keys:
```python
ad_dict = {key: None for key in keys}
```

tests/unit/dataactvalidator/test_b12_award_financial_1.py
```diff
@@ -18,58 +18,52 @@ def test_success(database):
...
+
+    for i in range(len(keys)):
+        af_dict_copy = copy.deepcopy(af_dict)
+        af_dict_copy.pop(keys[i])
```
cmc333333: Again, no biggie, but this'd be a little more pythonic:
```python
for key in keys:
    af_dict_copy = copy.deepcopy(af_dict)
    af_dict_copy.pop(key)
```

Also, I don't know how widely used this practice is, but if you want to make a shallow copy (i.e. just another set of keys pointing to the same values), you can do
```python
af_dict_copy = dict(af_dict)
```

tests/unit/dataactvalidator/test_b12_award_financial_2.py
```diff
@@ -11,22 +11,14 @@ def test_column_headers(database):
     assert (actual & expected_subset) == expected_subset
 
 
-def test_success(database):
+@pytest.mark.parametrize('af', [AwardFinancialFactory(by_direct_reimbursable_fun=None),
+                                AwardFinancialFactory(by_direct_reimbursable_fun=''),
+                                AwardFinancialFactory(by_direct_reimbursable_fun='R'),
+                                AwardFinancialFactory(by_direct_reimbursable_fun='r'),
+                                AwardFinancialFactory(by_direct_reimbursable_fun='D'),
+                                AwardFinancialFactory(by_direct_reimbursable_fun='d')])
```
cmc333333: I like parametrize here, but wonder about creating whole objects -- we want to be as concise as possible. Would it make sense to have the parameter just be `[None, '', 'R', 'r'...]` and then create the `AwardFinancial` in the test?

PR #336, opened 2016-10-05
Cross file validation now uses error message instead of description

jworcestBAH:
Fixes a bug where rule descriptions were displayed as error messages for cross-file validations


jworcestBAH:
[run tests]

cmc333333:
:+1: Makes sense to me

PR #337, opened 2016-10-06
Db 1220 handle missing d files

jworcestBAH:
Improved error handling of bad download links for D files


cmc333333:
The logic here makes sense to me; if it's not super time sensitive, would you mind adding a unittest for this path of `load_d_file`?

jworcestBAH:
I have a branch where I've been experimenting with unit tests for loading D files, the issue is that the session-scope fixture uses a validator interface holder, and I haven't been able to find a way to do a fixture with a broker interface holder that doesn't break that.  It looks like Becky has a current PR to combine the interface holders, so once that gets in I can revive that branch and get those tests working.

cmc333333:
:+1:

PR #338, opened 2016-10-06
Refactor userTests database connections

bsweger:
This PR removes the old database interfaces and interface holders from userTests.py, in favor of using the GlobalDB object.

This is the last refactor in the series of tests that inherit from baseTestAPI, so the PR also updates that class to remove the last vestiges of `InterfaceHolder()`.

Note that I moved some user token functions over to the function_bag as part of this work but didn't do a lot work on them, knowing that we're getting ready to move our user auth to MAX.gov.


nmonga91:
Everything looks good! Just one quick question, what dictates what will go into `function_bag.py`?

bsweger:
@nmonga91 good question...the intention of `function_bag.py` is to serve as a parking lot for all of the methods in the current db interface classes (with the exception of a few one-liners that may be dropped in favor of doing them inline).

Most of the move will happen when we refactor the app itself (see stories 1113, 1199, 1200, 1201). Not much is there currently--just things that got moved over out of necessity as I refactored the db setup code and integration tests.

PR #339, opened 2016-10-06
Updated C14 to check for combination of piid with either fain or uri

jworcestBAH:
Also adds unit test for new part and updates integration tests for the change from warning to fatal


jworcestBAH:
[run tests]

tests/unit/dataactvalidator/test_c14_award_financial_2.py
```diff
@@ -0,0 +1,32 @@
...
+    expected_subset = {'row_number', "fain", "uri", "piid"}
+    actual = set(query_columns(_FILE, database))
+    assert (actual & expected_subset) == expected_subset
+
+
+def test_success(database):
+    """ Test cases with different combinations of fain, uri, and piid """
+
+    # Test with only one present
+    award_fin_fain = AwardFinancialFactory(uri = None, piid = None)
```
bsweger: Great job hitting all the edge cases in this test!

dataactvalidator/config/sqlrules/c14_award_financial_2.sql
```diff
@@ -0,0 +1,5 @@
+SELECT row_number, fain, uri, piid
```
bsweger: Even though the original C14 was pretty simple, I like the idea of having this logic as a separate query to enable a more specific error message.

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -88,7 +88,8 @@ C8,Unique FAIN/URI from file C exists in file D2,Unique FAIN/URI from file C exi
 C9,"Unique FAIN/URI from file D2 exists in file C, except D2 records with zero FederalActionObligation","Unique FAIN/URI from file D2 exists in file C, except D2 records with zero FederalActionObligation",y,award,warning,award_financial,c9_award_financial
 C11,"Unique PIID, ParentAwardId from file C exists in file D1 during the same reporting period.","Unique PIID, ParentAwardId from file C exists in file D1 during the same reporting period.",y,award_financial,warning,award_procurement,c11_cross_file
 C12,"Unique PIID, ParentAwardId from file D1 exists in file C during the same reporting period, except D1 records with zero FederalActionObligation","Unique PIID, ParentAwardId from file D1 exists in file C during the same reporting period, except D1 records with zero FederalActionObligation",y,award_procurement,warning,award_financial,c12_cross_file
-C14,If FAIN is not provided then provide URI,If FAIN is not provided then provide URI,n,award_financial,warning,,c14_award_financial
+C14,If FAIN is not provided then provide URI,"Must provide either FAIN, URI, or PIID",n,award_financial,fatal,,c14_award_financial_1
```
bsweger: I realize this text pre-dates your PR, but the `rule_description` here (2nd column) isn't right: `If FAIN is not provided then provide URI`

Since we now have D1 records in the mix, would it make more sense to re-word as `FAIR, URI, or PIID must be provided` (or just copy the current wording from the validation rules .csv, which is maybe what we did to begin with, and now it needs updated).

I know someone will likely take a final pass at the various error messages, but I want to make sure we're not confusing agencies in the meantime.

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -88,7 +88,8 @@ C8,Unique FAIN/URI from file C exists in file D2,Unique FAIN/URI from file C exi
...
+C14,If FAIN is not provided then provide URI,PIID may not be provided if FAIN or URI are present,n,award_financial,fatal,,c14_award_financial_2
```
bsweger: Same question here about updating the text in `rule_description`

bsweger: Also, thanks for making the `rule_error_message` specific to the part of the rule being checked in the new query!

dataactvalidator/config/sqlrules/sqlRules.csv
jworcestBAH: Updated to a newer description from the validator rules sheet that includes piid.

bsweger:
Thanks for updating that description! Merging....

PR #340, opened 2016-10-06
Retry file E on exceptions

cmc333333:
The core here in the last changeset, which adds the retry logic. Along the way, there are several PRs to add better test coverage, some new factories, and a new fixture. This is likely easiest to review changeset by changeset.


cmc333333:
Perhaps the easiest way to test this locally would be to raise an exception in `dataactcore.utils.fileE`'s `retrieveRows` and watch the logs in the `jobQueue` -- you should see it retry 3 times (4 total exceptions) and then give up.

PR #341, opened 2016-10-07
Combine interface holders

bsweger:
To help fix the `InterfaceHolder` type mix-up that is causing the errors in PR #338, this PR combines the broker and the validator interface holder classes into a single class that lives in `dataactcore`.

Basically, it moves the methods from the classes in `dataactvalidator/interfaces` to their corresponding interface classes in `dataactcore/models`. Then, the new `InterfaceHolder` in `dataactcore/models` can use the `dataactbroker` series of interfaces, which inherit from those in `dataactcore`.

It's not overly-pretty, but having a single `InterfaceHolder` simplifies things a bit. All of the interfaces and holders are deprecated--we just need the tests to continue working in the meantime.

I did run my local #338 branch using `py.test` with these changes and had success, so fingers crossed.


dataactvalidator/interfaces/validatorErrorInterface.py
```diff
@@ -1,154 +1,9 @@
-from sqlalchemy.orm.exc import NoResultFound,MultipleResultsFound
-from dataactcore.utils.responseException import ResponseException
-from dataactcore.models.errorModels import File, ErrorMetadata
 from dataactcore.models.errorInterface import ErrorInterface
-from dataactvalidator.validation_handlers.validationError import ValidationError
 
 class ValidatorErrorInterface(ErrorInterface):
```
jworcestBAH: Is there a reason for these classes to still exist?  I thought the idea in this PR was to remove the separate interfaces for the validator, and just have the validator directly use the interfaces in the core.  If this is being kept as a way of not needing to change the existing uses, would it be more direct to just have the import statement be "from dataactcore.models.errorInterface import ErrorInterface as ValidatorErrorInterface" and skip the creation of the subclass?

jworcestBAH: Doing some searching, the only one of these that is imported directly rather than through the InterfaceHolder is the ValidatorValidationInterface, which is used in resetAlembicVersion and csvAbstractReader.  So the JobTracker, Error, and Staging versions could probably be removed.

bsweger: Good call--I didn't dig that deeply. Will remove these and push the change. Thanks!

bsweger:
[run-tests]

bsweger:
@jworcestBAH thanks for the suggestion--I went ahead and removed all of the broker and validator /interfaces files to prevent future confusion. Good suggestion!

jworcestBAH:
Looks good, merging

PR #342, opened 2016-10-07
WIP: Changed from random values to constant values

jworcestBAH:
This is the change recommended in the earlier DB-1104 PR, @bsweger could you take a look?  Marked as WIP since we're past code freeze.

- [x] Sprint 23 started
- [x] Code review complete


bsweger:
@jworcestBAH Thanks for taking the time to make these changes. I think this makes the intent of the tests more explicit, especially for the decimal values.

Good to merge once the next sprint is underway!

bsweger:
This was reviewed last week--we can merge now that sprint 23 is done.

PR #343, opened 2016-10-07
Db 1233 d file skip basic validations

jworcestBAH:
This adds a check to skip basic validations for D files without being dependent on the schema.

- [x] Sprint 23 has started


jworcestBAH:
Also tested this with the D file schema setting all fields to required integers, did not run into any errors with that.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -375,7 +375,12 @@ def runValidation(self, jobId, interfaces):
                     # second phase of validations: do basic schema checks
                     # (e.g., require fields, field length, data type)
                     #
-                    passedValidations, failures, valid = Validator.validate(record,csvSchema,fileType,interfaces)
+                    if fileType in ["award", "award_procurement"]:
+                        # Skip basic validations for D files, set as valid to trigger write to staging
```
bsweger: I really like the simplicity of this approach and that it will allow us to keep the D file schemas intact and consistent with the IDD as much as possible. One minor thing we might want to think about here is an expanded comment for future teams that explains why we're skipping basic validations for the D files (_e.g._, "the broker gets D files from upstream systems (ASP and FPDS) that perform their own basic validations, so we don't repeat those" or something like that?).

jworcestBAH: Added a comment for this

jworcestBAH:
Reversed the changes to D file schemas since those changes are no longer needed

bsweger:
Thanks for the updates, @jworcestBAH--looks good!

PR #344, opened 2016-10-08
Change file E column names and fix limit bug

cmc333333:
The original IDD included three name fields per officer (first, middle, last),
but the SAM data only contained one ("name"). While we kicked off that
discussion, the code used a single column per officer. We've now settled on a
single column but with a different name.

Also fixes an error caused by a limit on the number of DUNS we can request at once.


dataactcore/utils/fileE.py
```diff
@@ -43,6 +43,10 @@ def getEntities(client, dunsList):
     result = client.service.getEntities(
         createAuth(client), createSearch(client, dunsList), params)
 
+    if result.transactionInformation.transactionMessage:
+        logger.warning("Message from SAM API: %s",
```
nmonga91: +1 for more logging!

dataactcore/utils/jobQueue.py
```diff
@@ -110,9 +110,13 @@ def generate_e_file(task, submission_id, job_id, interface_holder_class,
             filter(AwardFinancialAssistance.submission_id == submission_id).\
             distinct()
         dunsSet = {r.awardee_or_recipient_uniqu for r in d1.union(d2)}
+        dunsList = list(dunsSet)    # get an order
 
+        rows = []
+        for i in range(0, len(dunsList), 100):
+            rows.extend(fileE.retrieveRows(dunsList[i:i+100]))
```
nmonga91: Is this just to retrieve the rows 100 at a time, instead of getting all of them like below (on the deleted line 115)?

cmc333333: Exactly, we can only send 100 numbers to SAM (through `retrieveRows`) at a time

nmonga91:
Everything looks good, ready to merge

PR #345, opened 2016-10-11
remove db interfaces from validator tests - part 1

bsweger:
- [x] Sprint 23 is complete and `dev-stable` is created
- [ ] Code is reviewed

Remove the db interfaces from BaseTestValidator (excepting lines that save them for use by the sub-classes, which have not yet been re-factored).

The changes to `baseTestValidator.py` are not extensive, but they did involve updating a few related items:
* Remove testing of the no-longer-used threaded validation job
* Move the jobTracker function that returns the error and warning report path to a new `dataactcore/util` file. This short function seemed important enough to centralize, in case our report naming convention ever changes, but didn't quite fit into `function_bag.py`, since there's no db access involved.


tests/integration/baseTestValidator.py
```diff
@@ -116,57 +116,58 @@ def run_test(self, jobId, statusId, statusName, fileSize, stagingRows,
...
+            job = sess.query(Job).filter(Job.job_id == jobId).one()
+            if statusName is not False:
+                self.assertEqual(job.job_status_id, JOB_STATUS_DICT[statusName])
+
+            self.assertEqual(
+                response.headers.get("Content-Type"), "application/json")
+
+            # Check valid row count for this job
+            if stagingRows:
+                self.assertEqual(job.number_of_rows_valid, stagingRows)
```
cmc333333: Nice! This was the only use of getNumberOfValidRowsById, so you can remove the method, too

tests/integration/baseTestValidator.py
```diff
@@ -116,57 +116,58 @@ def run_test(self, jobId, statusId, statusName, fileSize, stagingRows,
...
+            # get the job from db
+            job = sess.query(Job).filter(Job.job_id == jobId).one()
+            if statusName is not False:
+                self.assertEqual(job.job_status_id, JOB_STATUS_DICT[statusName])
+
+            self.assertEqual(
+                response.headers.get("Content-Type"), "application/json")
+
+            # Check valid row count for this job
+            if stagingRows:
```
cmc333333: Note that this won't execute when `stagingRows` is zero. I don't know if that has an impact

tests/integration/baseTestValidator.py
```diff
@@ -116,57 +116,58 @@ def run_test(self, jobId, statusId, statusName, fileSize, stagingRows,
...
+                self.assertEqual(job.number_of_rows_valid, stagingRows)
+
+            if errorStatus is not False:
+                self.assertEqual(
+                    sess.query(File).filter(File.job_id == jobId).one().file_status_id,
```
cmc333333: It might be more legible to hop over from the `job` instead. Something like
```python
assert job.file.file_status_id == FILE_STATUS_DICT[errorStatus]
```
Same number of queries

dataactbroker/handlers/fileHandler.py
```diff
@@ -80,13 +82,17 @@ def getErrorReportURLsForSubmission(self, isWarning = False):
             safeDictionary = RequestDictionary(self.request)
             submissionId = safeDictionary.getValue("submission_id")
             responseDict ={}
+            sess = GlobalDB.db().session
             for jobId in self.jobManager.getJobsBySubmission(submissionId):
-                if(self.jobManager.getJobType(jobId) == "csv_record_validation"):
+                # get the job object here so we can call the refactored getReportPath
+                # todo: replace other db access functions with job object attributes
+                job = sess.query(Job).filter(Job.job_id == jobId).one()
+                if self.jobManager.getJobType(jobId) == "csv_record_validation":
```
bsweger: Great point about prefetch. Per our convo in Slack, I updated the `Job` model to do an eager load of job type, file type, and job status info, since we almost always want one of those pieces of info when querying a job.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -325,7 +333,7 @@ def runValidation(self, jobId, interfaces):
 
         # Get file size and write to jobs table
         if CONFIG_BROKER["use_aws"]:
-            fileSize = s3UrlHandler.getFileSize("errors/"+jobTracker.getReportPath(jobId))
+            fileSize = s3UrlHandler.getFileSize("errors/"+errorFileName)
```
bsweger: It's not super clear, but the error and warning reports on S3 are prefixed with `errors/`. So the code that goes to S3 to get the file size needs the prefix, and the non-S3 code will just get the file name as stored in the database. (`errorFileName` doesn't do any AWS checking).

Does that make sense, or am I missing something (entirely possible, as you can see). I'll put a comment here.

tests/integration/baseTestValidator.py
bsweger: Another good catch! If any test was trying to test a 0 row count, it would always pass. I'm 98.5% sure we're not trying to test a 0 here, but will fix this up in anticipation of that day.

tests/integration/baseTestValidator.py
bsweger: Sheesh, how did I miss that one? Deleting things from the interface files is a source of great joy.

tests/integration/baseTestValidator.py
bsweger: The point re: legibility is fair, but since the Job model doesn't have a relationship back to `file` (only `file_type`), I think I'm going to leave this for now and not pull the thread of re-jiggering the Job model.

bsweger:
@cmc333333 Thanks again for the thorough review and for catching some things that would have been bad news. I think I've addressed everything.

cmc333333:
Excellent additions @bsweger !

PR #346, opened 2016-10-12
Changed error message for bad links in D-file callback

jworcestBAH:
Updates language on callback URL error


bsweger:
@jworcestBAH Looks good-merging!

PR #348, opened 2016-10-12
remove db interfaces from validator tests - part 2

bsweger:
- [x] PR #345 is merged
- [x] Dev is merged into this branch and conflicts resolved
- [ ] Code review complete

This PR removes database interfaces from the submission and job setup done by the test classes that inherit from BaseTestValidator (FileTypeTests, JobTests, and MixedFileTests).


tests/integration/jobTests.py
```diff
@@ -1,110 +1,122 @@
...
+                prerequisite_id=jobDict["bad_upload"])
+            sess.add(dependency)
 
             colIdDict = {}
             for fileId in range(1, 5):
                 for columnId in range(1, 6):
-                    #TODO: get rid of hard-coded surrogate keys
                     if columnId < 3:
-                        fieldType = 1
+                        fieldType = FIELD_TYPE_DICT['INT']
```
cmc333333: 👏 :clap: :clap:

tests/integration/jobTests.py
```diff
@@ -159,7 +171,7 @@ def test_bad_type_job(self):
     # TODO uncomment this test if we start limiting the validator to only jobs that are "ready"
     #def test_finished_job(self):
     #    """ Test job that is already finished """
-    #    jobId = self.jobIdDict["finished"]
+    #    jobId = self.jobDict["finished"]
     #    self.run_test(jobId,400,"finished",False,False,"job_error",0)
```
cmc333333: What do you think about removing these commented tests altogether?

tests/integration/fileTypeTests.py
```diff
@@ -30,56 +30,79 @@ def setUpClass(cls):
...
-            jobIdDict = {}
-            for key in jobInfoList:
-                jobInfo = jobInfoList[key]  # Done this way to be compatible with python 2 and 3
-                jobInfo.append(jobDb.session)
-                job = cls.addJob(*jobInfo)
-                jobId = job.job_id
-                jobIdDict[key] = jobId
-                print("".join([str(key),": ",str(cls.jobTracker.getSubmissionId(jobId)), ", "]), end = "")
+            for job_type, job_id in jobDict.items():
+                print('{}: {}'.format(job_type, job_id))
```
cmc333333: I'm not sure if this is relevant, but py.test eats stdout and stderr.

tests/integration/jobTests.py
bsweger: When it comes to deleting something, you never have to ask me twice 😃

bsweger:
@cmc333333 Thanks! Just pushed a commit that deletes the duplicated code and unused tests. If that looks ok to you, I think we're good to go.

cmc333333:
Rock and roll

bsweger:
@cmc333333 sorry, forgot about the `flush()` question...I am doing that to get the id for the newly inserted object. It didn't feel quite right to me either, but I wasn't sure of a better way (didn't want to keep committing, since we're on auto-commit).

cmc333333:
@bsweger gotcha. Since we're auto-commiting, I think `flush` and `commit` are basically synonymous. If, instead, we had transactions, `flush` would hit the db within the transaction (and not commit). I may be misunderstanding the docs, though.

PR #350, opened 2016-10-13
Pointing C23 parts 3 and 4 at D2 target file instead of D1

jworcestBAH:
This fixes the issue seen with C23 not catching errors between C and D2


bsweger:
Nice catch @jworcestBAH on finding and fixing the incorrect target file types. Merging!

PR #352, opened 2016-10-14
Db 1274 install initialize update

alburde1:
- [ ] Contribution docs reviewed by @bsweger 

This pull request updates the initialize.py script in dataactbroker to not rely on dynamodb running. It also updates the installation instructions with several issues that I ran into while installing the project.


bsweger:
@alburde1 This is great! Thank you so much for taking the time to improve the install process--much appreciated! Just had one small question, and then we should be good to go.

doc/CONTRIBUTING.md
```diff
@@ -54,6 +54,12 @@ More complete install documentation is available on the PostgreSQL [wiki](https:
 **Note:** The database user will need to have permission to create databases.
 A "superuser" will do.
 
+### Install Xcode CLI (Mac users only)
+
+Xcode Command Line Interface (CLI) is needed to install several of the items listed in requirements.txt on a Mac.
```
alburde1: Yes, that makes sense, change made.

doc/CONTRIBUTING.md
```diff
@@ -73,6 +79,25 @@ If you already have a Python development environment on your machine and a prefe
 
 4. Tell virtualenvwrapper where on your machine to create virtual environments and add it to your profile. This is a one-time virtualenvwrapper setup step, and the process varies by operating system. [This tutorial](http://newcoder.io/begin/setup-your-machine/ "Python: setting up your computer") covers setting up virtualenvwrapper on OSX, Linux, and Windows.
    For Windows users, there may be extra steps needed.  If you run into an error on the import-module step, move the "VirtualEnvWrapper" folder from C:/Python27/Lib/site-packages/Users/*username*/Documents/WindowsPowerShell/Modules/ to C:/Users/*username*/Documents/WindowsPowerShell/Modules/.  Next, in powershell run the command "set-executionpolicy unrestricted".  Finally, in the VirtualEnvWrapper directory, open the file "VirtualEnvWrapperTabExpansion.psm1" and change "Function:TabExpansion" to "Function:TabExpansion2" in line 12.
+   
+   **Note:** After running:
```
bsweger: This is super helpful info. Is it specific to Mac users, or applicable to Windows folks too?

doc/CONTRIBUTING.md
```diff
@@ -211,6 +236,15 @@ Once the DATA Act broker's backend is up and running, you may also want to stand
 
 After following the website setup directions, you can log in with the admin e-mail and password you set in the [broker's backend config file](#create-broker-config-file "config file setup") (`admin_email` and `admin_password`).
 
+### Run Celery
```
bsweger: Thanks for adding this much-needed section on celery!

doc/CONTRIBUTING.md
alburde1: I honestly have no idea but it seems more like a Mac issue. I don't want to say only Mac users can have this problem because I wouldn't know and I don't want to lead them astray, but if it happens on Windows I think the process of fixing it would be similar, just not ~/.bash_profile (no idea where though, but wherever they store their paths)

bsweger: Yeah, it's a tough one. I can see this happening on Windows, but without access to a Windows machine can't verify that. Since your (terrific) directions are so specific to Mac, what do you think about updating **Note:** to **Note to Mac and Linux users:**?

alburde1: Good idea, updated the text and pulled the new dev stuff down to make sure the merge goes smoothly.

bsweger:
Thanks, @alburde1! Looks good...merging!

PR #353, opened 2016-10-17
Remove db interfaces from validator tests - part 3 and final

bsweger:
This PR removes the remaining database interfaces from the validator unit tests.

### DIFF
--- a/tests/integration/validatorTests.py
+++ b/tests/integration/validatorTests.py
@@ -122,7 +122,6 @@ class ValidatorTests(BaseTestValidator):
     def test_schema_optional_field(self):
         """Test optional fields."""
         schema = self.schema
-        interfaces = self.interfaces
         record = {
             "test1": "hello",
             "test2": "1.0",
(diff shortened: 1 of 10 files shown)


tests/integration/validatorTests.py
```diff
@@ -39,55 +40,48 @@ def setUpClass(cls):
         column1.required = True
         column1.field_type = stringType
         column1.file_id = 1
-        cls.interfaces.validationDb.populateFile(column1)
```
nmonga91: Looks like the tests are still passing. Pinging @jworcestBAH to ensure we're not overlooking an important use of that method.

jworcestBAH: What this method did was use FileColumn.file_id to populate the FileColumn.file relationship (which holds the file type), this would normally be populated by doing a joinedload, but here the FileColumn objects are being built manually.  The ValidatorTests setup populates the file type relationship for each of the columns, these column definitions get passed in to the validate function, I'm not sure how the file type is used within that code.

jworcestBAH: It's not happening elsewhere as these columns are built manually (not pulled from DB).  It's possible that validate no longer uses the file type relationship, or this may result in these rules being skipped because the file type doesn't match.

jworcestBAH: It looks like the validationManager does the filtering by file type, so by the time it gets passed to validate it only includes rules for the current file.  So the file type is probably not being used in these tests.

nmonga91:
Looks good, merging!

PR #355, opened 2016-10-17
Change TAS file format

cmc333333:
We're getting our TAS files from a different source now, which has a slightly
different set of columns. Modify the import script to accommodate and replace
the CSVs included in the app. The logic is simplified as our data source is
now more accurate (not including duplicate rows, for example).

This is a first step towards replacing the `tas` field in our models with the new `ACCT_NUM`-based approach.

### DIFF
--- a/dataactvalidator/scripts/loadTas.py
+++ b/dataactvalidator/scripts/loadTas.py
@@ -13,6 +13,36 @@ from dataactvalidator.scripts.loaderUtils import LoaderUtils
 logger = logging.getLogger(__name__)
 
 
+def cleanTas(csvPath):
+    """Read a CSV into a dataframe, then use a configured `cleanData` and
+    return the results"""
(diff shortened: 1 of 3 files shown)


jworcestBAH:
Have you tested loading the new all_tas_betc.csv file and making sure the tas_lookup table gets populated?  This file is not used during the testing, so we should test that the new file can be loaded correctly before making this change.

cmc333333:
@jworcestBAH thanks for pushing back! It does indeed load the data and the integration test data has been updated to use the new headers. I hadn't tried running through a validation, however, which is failing. Investigating now.

cmc333333:
Updated to account for strings of spaces and added some tests; ready for re-review

jworcestBAH:
Looks like my last question was about code that got deleted, so merging now

PR #357, opened 2016-10-18
Db 1261 update a16

jworcestBAH:
Updates rule A16 to apply to the first submission in a fiscal year regardless of what quarter it is, also removes ussgl fields for file C from A16


tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,21 +1,59 @@
...
+    # No previous submission and null budget_authority_unobligat_fyb
+    sub_new = SubmissionFactory()
+    ap_new = AppropriationFactory(submission_id = sub_new.submission_id, budget_authority_unobligat_fyb = None)
+    assert number_of_errors(_FILE, database, submission = sub_new, models=[ap_new]) == 1
+
+    # previous submission exists but is unpublished and has not been marked publishable
+    sub_prev_published = SubmissionFactory(publish_status_id = PUBLISH_STATUS_DICT['unpublished'], publishable = False)
+    insert_submission(database, sub_prev_published)
+    sub_new_published = SubmissionFactory(cgac_code = sub_prev_published.cgac_code, reporting_fiscal_year = sub_prev_published.reporting_fiscal_year)
+    ap_prev_published = AppropriationFactory(submission_id = sub_prev_published.submission_id)
```
bsweger: Same question re: do we need this record?

tests/unit/dataactvalidator/test_a16_award_financial.py
```diff
@@ -0,0 +1,59 @@
...
+    af_new_published = AwardFinancialFactory(submission_id = sub_new_published.submission_id, obligations_delivered_orde_fyb = None)
+    assert number_of_errors(_FILE, database, submission = sub_new_published,
+                      models=[af_prev_published, af_new_published]) == 0
+
+    # Previous submission marked as publishable also allows null
+    sub_prev_publishable = SubmissionFactory(publishable = True)
+    insert_submission(database, sub_prev_publishable)
+    sub_new_publishable = SubmissionFactory(cgac_code = sub_prev_publishable.cgac_code, reporting_fiscal_year = sub_prev_publishable.reporting_fiscal_year)
+    af_prev_publishable = AwardFinancialFactory(submission_id = sub_prev_publishable.submission_id)
+    af_new_publishable = AwardFinancialFactory(submission_id = sub_new_publishable.submission_id, obligations_delivered_orde_fyb = None)
```
bsweger: Do you think it makes sense to exercise the `OR` logic in the query by setting one or more of the other fields (e.g., _ussgl480100_undelivered_or_fyb_ and the like) to NULL as well?

tests/unit/dataactvalidator/test_a16_award_financial.py
```diff
@@ -0,0 +1,59 @@
...
+    # obligations_delivered_orde_fyb populated does not require a previous submission
+    sub_new = SubmissionFactory()
+    af_new = AwardFinancialFactory(submission_id = sub_new.submission_id)
+    assert number_of_errors(_FILE, database, submission = sub_new, models=[af_new]) == 0
+
+    # obligations_delivered_orde_fyb can be null if previous published submission shares cgac and fiscal year
+    sub_prev_published = SubmissionFactory(publish_status_id = PUBLISH_STATUS_DICT['published'])
+    insert_submission(database, sub_prev_published)
+    sub_new_published = SubmissionFactory(cgac_code = sub_prev_published.cgac_code, reporting_fiscal_year = sub_prev_published.reporting_fiscal_year)
+    af_prev_published = AwardFinancialFactory(submission_id = sub_prev_published.submission_id)
```
bsweger: I probably sound like a broker record by now, so I'll stop asking the same question repeatedly. Consider this a blanket question asking if we need to insert staging records for the prior submissions in these tests.

tests/unit/dataactvalidator/test_a16_award_financial.py
```diff
@@ -0,0 +1,59 @@
...
+    sub_new = SubmissionFactory()
+    af_new = AwardFinancialFactory(submission_id = sub_new.submission_id, obligations_delivered_orde_fyb = None)
+    assert number_of_errors(_FILE, database, submission = sub_new, models=[af_new], assert_num=1) == 1
+
+    # previous submission exists but is unpublished and has not been marked publishable
+    sub_prev_published = SubmissionFactory(publish_status_id = PUBLISH_STATUS_DICT['unpublished'], publishable = False)
+    insert_submission(database, sub_prev_published)
+    sub_new_published = SubmissionFactory(cgac_code = sub_prev_published.cgac_code, reporting_fiscal_year = sub_prev_published.reporting_fiscal_year)
+    af_prev_published = AwardFinancialFactory(submission_id = sub_prev_published.submission_id)
+    af_new_published = AwardFinancialFactory(submission_id = sub_new_published.submission_id, gross_outlays_undelivered_fyb = None)
```
bsweger: Same question as above--is there any value in having at least one of these test cases set more than one of the fields being checked to NULL?

tests/unit/dataactvalidator/test_a16_object_class_program_activity.py
```diff
@@ -0,0 +1,61 @@
...
+    ocpa_new = ObjectClassProgramActivityFactory(submission_id = sub_new.submission_id)
+    assert number_of_errors(_FILE, database, submission = sub_new, models=[ocpa_new]) == 0
+
+    # gross_outlays_delivered_or_fyb can be null if previous published submission shares cgac and fiscal year
+    sub_prev_published = SubmissionFactory(publish_status_id = PUBLISH_STATUS_DICT['published'])
+    insert_submission(database, sub_prev_published)
+    sub_new_published = SubmissionFactory(cgac_code = sub_prev_published.cgac_code, reporting_fiscal_year = sub_prev_published.reporting_fiscal_year)
+    ocpa_prev_published = ObjectClassProgramActivityFactory(submission_id = sub_prev_published.submission_id)
+    ocpa_new_published = ObjectClassProgramActivityFactory(submission_id = sub_new_published.submission_id, 
+       gross_outlays_delivered_or_fyb = None, ussgl480100_undelivered_or_fyb = None)
```
bsweger: Nice, checking a few of the fields on the list!

jworcestBAH:
Just spoke with Kristine, apparently the non-ussgl fields in file C are "ussgl-related", so file C is now being dropped entirely from this rule.  Pushing that change now.

jworcestBAH:
I don't remember everything about the publishable/publish_status conversation, this came out of the meeting we had with Drew, Arman, and Timothy about how the data store should identify which submissions to pull.  I believe publishable was intended to essentially be "should this submission be pulled to data store", while publish_status was intended to distinguish whether it was a new submission or an updated submission as they treat those cases differently.  We could certainly change this to five publish_status values (unpublished, publishable, published, updated-but-not-ready, updated-and-publishable), but we'd need to clear that with the data store team first as they may have started implementing this based on the last conversation.

tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,13 +1,31 @@
...
 
 
 _FILE = 'a16_appropriations'
 
 
 def test_success(database):
-    award = Appropriation(job_id=1, row_number=1)
-    assert error_rows(_FILE, database, models=[award]) == []
+    sub_new = SubmissionFactory()
+    sub_prev_published = SubmissionFactory()
```
jworcestBAH: The submission factory sets publishable to False and publish_status_id is left null.  I'm setting it in each test case that needs it.  This line currently sets the status to published, perhaps this comment was on an older commit?

bsweger:
@jworcestBAH Thanks for refreshing my memory on the publishable convo. For the sake of getting the A16 rule updated and into the code base, let's leave this for now. 

I'll put in a separate story for making this change because my vote would be to track status in a single place.

On the GTAS side, they have statuses like "validated" and "certified", where the former is similar to what we're calling "publishable." If/when we tackle this, might be good to use their terminology if it makes sense

tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,13 +1,31 @@
...
+    insert_submission(database, sub_new_publishable)
+    # budget_authority_unobligat_fyb can be null if previous submission shares cgac and fiscal year
+    ap_new = AppropriationFactory(submission_id = sub_new.submission_id)
+    ap_prev_published = AppropriationFactory(submission_id = sub_prev_published.submission_id)
+    ap_new_published = AppropriationFactory(submission_id = sub_new_published.submission_id, budget_authority_unobligat_fyb = None)
+    ap_prev_publishable = AppropriationFactory(submission_id = sub_prev_publishable.submission_id)
+    ap_new_publishable = AppropriationFactory(submission_id = sub_new_publishable.submission_id, budget_authority_unobligat_fyb = None)
+    assert error_rows(_FILE, database, models=[sub_new, sub_prev_published, sub_new_published, sub_prev_publishable,
+                                               sub_new_publishable, ap_new, ap_prev_published, ap_new_published,
+                                               ap_prev_publishable, ap_new_publishable]) == []
```
jworcestBAH: Split the test cases into separate methods, and changed the inline comments for each case into docstrings in the new methods

tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,21 +1,59 @@
...
+    # budget_authority_unobligat_fyb populated does not require a previous submission
+    sub_new = SubmissionFactory()
+    ap_new = AppropriationFactory(submission_id = sub_new.submission_id)
+    assert number_of_errors(_FILE, database, submission = sub_new, models=[ap_new]) == 0
+
+    # budget_authority_unobligat_fyb can be null if previous published submission shares cgac and fiscal year
+    sub_prev_published = SubmissionFactory(publish_status_id = PUBLISH_STATUS_DICT['published'])
+    insert_submission(database, sub_prev_published)
+    sub_new_published = SubmissionFactory(cgac_code = sub_prev_published.cgac_code, reporting_fiscal_year = sub_prev_published.reporting_fiscal_year)
+    ap_prev_published = AppropriationFactory(submission_id = sub_prev_published.submission_id)
```
jworcestBAH: Good catch, these are not actually doing anything and have now been removed

tests/unit/dataactvalidator/test_a16_appropriations.py
```diff
@@ -1,21 +1,59 @@
...
+    ap_new_published = AppropriationFactory(submission_id = sub_new_published.submission_id, budget_authority_unobligat_fyb = None)
+    assert number_of_errors(_FILE, database, submission = sub_new_published,
+                      models=[ap_prev_published, ap_new_published]) == 0
+
+    # Previous submission marked as publishable also allows null
+    sub_prev_publishable = SubmissionFactory(publishable = True)
+    insert_submission(database, sub_prev_publishable)
+    sub_new_publishable = SubmissionFactory(cgac_code = sub_prev_publishable.cgac_code, reporting_fiscal_year = sub_prev_publishable.reporting_fiscal_year)
+    ap_prev_publishable = AppropriationFactory(submission_id = sub_prev_publishable.submission_id)
```
jworcestBAH: Removed these

tests/unit/dataactvalidator/test_a16_appropriations.py
jworcestBAH: Also removed

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -1,100 +1,100 @@
...
+A11,SpendingAuthorityfromOffsettingCollectionsAmountTotal_CPE= CPE aggregate value for GTAS SF 133 line #1750 + #1850,Failed rule: SpendingAuthorityfromOffsettingCollectionsAmountTotal_CPE= CPE aggregate value for GTAS SF 133 line #1750 + #1850,n,appropriations,warning,,a11_appropriations
+A12,AdjustmentsToUnobligatedBalanceBroughtForward_CPE= CPE aggregate value for GTAS SF 133 line #1010 through 1042,Failed rule: AdjustmentsToUnobligatedBalanceBroughtForward_CPE= CPE aggregate value for GTAS SF 133 line #1010 through 1042,n,appropriations,warning,,a12_appropriations
+A14,GrossOutlayAmountByTAS_CPE= CPE value for GTAS SF 133 line #3020,Failed rule: GrossOutlayAmountByTAS_CPE= CPE value for GTAS SF 133 line #3020,n,appropriations,warning,,a14_appropriations
+A15,UnobligatedBalance_CPE= CPE value for GTAS SF 133 line #2490,Failed rule: UnobligatedBalance_CPE= CPE value for GTAS SF 133 line #2490,n,appropriations,warning,,a15_appropriations
+A16,All the elements that have FYB in file A are expected in quarter 1.,All the elements that have FYB in file A are expected in the first submission for a fiscal year,n,appropriations,warning,,a16_appropriations
```
jworcestBAH: Fixed to be consistent with the error message

tests/unit/dataactvalidator/test_a16_award_financial.py
jworcestBAH: All have been removed

tests/unit/dataactvalidator/test_a16_award_financial.py
jworcestBAH: The file B test has cases where two of the fields are set to null, and the file C test no longer exists.

jworcestBAH:
I think everything here has been addressed, let me know if there are any other changes to be made.

tests/unit/dataactvalidator/test_a16_appropriations.py
bsweger: @jworcestBAH Yes, apologies for the confusion--I was reading thru commit by commit and forgot to go back and remove that comment

bsweger:
@jworcestBAH Looks good--thank you!

PR #358, opened 2016-10-18
Updating C8 and C9 SQL Validations

mab6bf:


dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,14 +1,22 @@
...
+                    ON (afa.submission_id = af.submission_id
+                        AND (afa.fain IS NOT DISTINCT FROM af.fain
+                            AND afa.uri IS NOT DISTINCT FROM af.uri)
+                    )
+            WHERE COALESCE(CAST(afa.federal_action_obligation as numeric),0) <> 0
+                AND COALESCE(CAST(afa.original_loan_subsidy_cost as numeric),0) <> 0
+                AND afa.submission_id = {0}
+        ) OR (
+            afa.fain IS NOT NULL
+            AND afa.uri IS NOT NULL
```
jworcestBAH: Two questions on this.  First, should FAIN and URI both present get a warning for C8 and a warning for C9, or should these be consolidated in some way?  Second, should the C9 warning for FAIN and URI both present be skipped when the fields above are equal to zero?  From the text "Unique FAIN/URI from file D2 exists in file C, except D2 records where FederalActionObligation and OriginalLoanSubsidyCost = 0", I get the impression that exception only applies to the check between C and D2, not the FAIN and URI both provided warning.

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+    errors = number_of_errors(_FILE, database, models=[afa, af])
+    assert errors == 0
+
+
+def test_equal_uri(database):
+    """Tests that File D2 (award financial assistance) uri matches
+    File C (award financial) uri."""
+    tas = _TAS
+    afa = AwardFinancialAssistanceFactory(
+        tas=tas, fain=None, uri='xyz', federal_action_obligation='1', original_loan_subsidy_cost='1')
```
jworcestBAH: Might be good to have one of these tests set e.g. federal_action_obligation to 0 to make sure the logic is working correctly that only one needs to be nonzero

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+
+
+def test_both_fain_and_url_supplied(database):
+    """Tests File D2 (award financial assistance) having both uri and fain populated."""
+    tas = _TAS
+    afa = AwardFinancialAssistanceFactory(
+        tas=tas, fain='abc', uri='xyz', federal_action_obligation='1', original_loan_subsidy_cost='1')
+
+    errors = number_of_errors(_FILE, database, models=[afa])
+    assert errors == 1
```
jworcestBAH: If C9 above is changed to apply the check to AF this will need to change as well

dataactvalidator/config/sqlrules/c9_award_financial.sql
jworcestBAH: So after seeing it in the unit test below I realized this check is being done on D2, which we should definitely confirm as I thought the expectation was that D2 can have both and file C should be just one

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+        tas=tas, fain=None, uri=None, federal_action_obligation='1', original_loan_subsidy_cost='1')
+    af = AwardFinancialFactory(
+        tas=tas, submisson_id=afa.submission_id, fain=None, uri='abc')
+
+    errors = number_of_errors(_FILE, database, models=[afa, af])
+    assert errors == 1
+
+
+def test_zero_federal_action_obligation_and_original_loan_subsidy_cost(database):
+    """Tests that a single warning is thrown for both a federal action obligation of 0
```
jworcestBAH: This is testing for zero warnings, maybe the docstring should be updated?

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+    af = AwardFinancialFactory(
+        tas=tas, submisson_id=afa.submission_id, fain=None, uri=None)
+
+    errors = number_of_errors(_FILE, database, models=[afa, af])
+    assert errors == 0
+
+
+def test_zero_federal_action_obligation_and_positive_original_loan_subsidy_cost(database):
+    """Tests that a single warning is thrown for both a federal action obligation of 0
+     and an original loan subsidy cost of 0"""
```
jworcestBAH: This docstring probably also needs an update as olsc is not 0

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+    tas = _TAS
+    afa = AwardFinancialAssistanceFactory(
+        tas=tas, fain='abc', uri=None, federal_action_obligation='0', original_loan_subsidy_cost='1')
+    af = AwardFinancialFactory(
+        tas=tas, submisson_id=afa.submission_id, fain=None, uri=None)
+
+    errors = number_of_errors(_FILE, database, models=[afa, af])
+    assert errors == 0
```
jworcestBAH: I'm surprised that we don't expect this one to error, that fain is not present in C and the fields are not both 0, what am I missing?

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -1,37 +1,151 @@
...
+    """Tests that a single warning is thrown for both a federal action obligation of 0
+     and an original loan subsidy cost of 0"""
+    tas = _TAS
+    afa = AwardFinancialAssistanceFactory(
+        tas=tas, fain='abc', uri=None, federal_action_obligation='1', original_loan_subsidy_cost='0')
+    af = AwardFinancialFactory(
+        tas=tas, submisson_id=afa.submission_id, fain=None, uri=None)
+
+    errors = number_of_errors(_FILE, database, models=[afa, af])
+    assert errors == 0
```
jworcestBAH: Same as above, unless I'm misreading the rule it seems like this should get an error

dataactvalidator/config/sqlrules/c8_award_financial.sql
```diff
@@ -1,17 +1,23 @@
-SELECT af.row_number, af.fain, af.uri
+SELECT DISTINCT af.row_number, af.fain, af.uri
```
mab6bf: Tested again - there are not. Removing.

dataactvalidator/config/sqlrules/c8_award_financial.sql
```diff
@@ -1,17 +1,23 @@
...
+            FROM award_financial AS af
+                JOIN award_financial_assistance AS afa
+                    ON (af.submission_id = afa.submission_id
+                        AND (af.fain IS NOT DISTINCT FROM afa.fain
+                            AND af.uri IS NOT DISTINCT FROM afa.uri)
+                    )
+            WHERE af.submission_id = {0}
+        ) OR (
+            af.fain IS DISTINCT FROM NULL
+            AND af.uri IS DISTINCT FROM NULL
```
jworcestBAH: Also it seems like the FAIN + URI warning should go into the file C warning report, while the other part should go into the cross-file C-D2 warning report, so that may be another reason to split them.

tests/unit/dataactvalidator/test_c9_award_financial.py
mab6bf: Good catch

tests/unit/dataactvalidator/test_c9_award_financial.py
mab6bf: Good catch

dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,14 +1,22 @@
...
-			JOIN award_financial AS af
-				ON (afa.fain IS NOT DISTINCT FROM af.fain
-				AND afa.uri IS NOT DISTINCT FROM af.uri
-				AND afa.submission_id = af.submission_id)
-		WHERE CAST(afa.federal_action_obligation as numeric) > 0
-			AND afa.submission_id = {0}
-	);
\ No newline at end of file
+    AND COALESCE(CAST(afa.federal_action_obligation as numeric),0) <> 0
+    AND COALESCE(CAST(afa.original_loan_subsidy_cost as numeric),0) <> 0
```
mab6bf: Done

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -84,8 +84,8 @@ C6,GrossOutlaysUndeliveredOrdersPrepaidTotal (FYB or CPE) = USSGL(4802 + 4882).
...
-C9,"Unique FAIN/URI from file D2 exists in file C, except D2 records with zero FederalActionObligation","Unique FAIN/URI from file D2 exists in file C, except D2 records with zero FederalActionObligation",y,award,warning,award_financial,c9_award_financial
+C8,"Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted.","Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted".,y,award_financial,warning,award,c8_award_financial
+C9,"Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted.","Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted.",y,award,warning,award_financial,c9_award_financial
```
jworcestBAH: C and D2 should be flipped for C9

jworcestBAH:
Okay it looks like those changes have addressed everything except the fain and uri both present warnings:
1.  For C9, this warning should not be present as it does not apply to D2, we don't want to be warning agencies that ASP is providing more info than needed.  Also, the error message should flip 'C' and 'D2'.
2.  For C8, we had talked about possibly splitting this into a separate query so it can have a more specific error message that goes into the file-level report for file C.  Since the error message is broader now this is an optional change, C8 will work as-is if we don't want to do this split.

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -84,7 +84,8 @@ C6,GrossOutlaysUndeliveredOrdersPrepaidTotal (FYB or CPE) = USSGL(4802 + 4882).
...
 C7,GrossOutlaysDeliveredOrdersPaidTotal (FYB or CPE) = USSGL(4902 + 4908 + 4982). This applies to the award level.,GrossOutlaysDeliveredOrdersPaidTotalCPE = USSGL(4902 + 4908 + 4982). This applies to the award level.,n,award_financial,warning,,c7_award_financial_1
 C7,GrossOutlaysDeliveredOrdersPaidTotal (FYB or CPE) = USSGL(4902 + 4908 + 4982). This applies to the award level.,GrossOutlaysDeliveredOrdersPaidTotalFYB = USSGL4908. This applies to the award level.,n,award_financial,warning,,c7_award_financial_2
-C8,"Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted.","Unique FAIN/URI from file C exists in file D2. FAIN may be null for aggregated records.  URI may be null for non-aggregated records. For FAIN/URI, a warning will be provided if both are submitted".,y,award_financial,warning,award,c8_award_financial
+C8,"FAIN or URI should not be submitted together.","FAIN or URI should not be submitted together.".,N,award_financial,warning,,c8_award_financial_1
```
jworcestBAH: Should this be "FAIN and URI"?  Applies to both columns.

jworcestBAH:
Okay, looks good now, one last minor grammatical quibble, and then I think we're good to merge

PR #359, opened 2016-10-18
Added a uuid to the session ID to prevent collisions from users at th…

jworcestBAH:
…e same IP

### DIFF
--- a/dataactbroker/handlers/aws/session.py
+++ b/dataactbroker/handlers/aws/session.py
@@ -139,7 +139,7 @@ class LoginSession():
 
         resets the _uid in cases that the session becomes invalid
         """
-        session["_uid"] = _create_identifier()
+        session["_uid"] = "{}|{}".format(_create_identifier(),uuid4())
 
     @staticmethod
     def isSessionSecure(session):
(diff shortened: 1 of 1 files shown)


cmc333333:
Would it make sense to drop the `_create_identifier()` call and just use a UUID for the session id?

dataactbroker/handlers/aws/session.py
```diff
@@ -254,7 +254,7 @@ def save_session(self, app, session, response):
             else:
                 expiration = datetime.utcnow() + timedelta(seconds=SessionTable.TIME_OUT_LIMIT)
         if(not "_uid" in session):
-            session["_uid"] = _create_identifier()
+            session["_uid"] = "{}|{}".format(_create_identifier(),uuid4())
```
cmc333333: Small detail, but rather than duplicating the logic, could we call `self.resetId(session)`?

jworcestBAH:
I thought about that, but we have a check in 'isSessionSecure' that checks whether the _create_identifier() part matches their session id to prevent session ids from being intercepted and used to mimic a user (this includes user's IP).  To keep that functionality available I combined the two.

dataactbroker/handlers/aws/session.py
jworcestBAH: Good call, changed to call this

cmc333333:
Gotcha; agreed that that's outside the scope for this PR, but I created DB-1287 to track it for later. This is good as is, though.

PR #360, opened 2016-10-18
hotfix: add missing import to the dataactbroker initialize script

bsweger:
Didn't catch this in tests, because test setup invokes the
setup functions individually.


cmc333333:
Eep! Let's follow up soon about how to prevent this kind of thing in the future.

PR #361, opened 2016-10-18
Import TAS unique id from CSV

cmc333333:
This builds on #355 ([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/755-tas-source...db-755-tas-id)) to import the unique identifier assigned to each TAS by the CARS file. ~~It uses that knowledge to only drop TAS entries that are part of the new data set when importing.~~ It doesn't use those ids for anything special at the moment.

This also includes the updated CARS files.


cmc333333:
This needs to be updated per @kaitlin -- it should drop _all_ tas lookup entries on load

bsweger:
@cmc333333 since we have so many outstanding PRs at the moment, I'll defer reviewing this until the additional changes are in. 

Also, a question...I saw another version of the CARS file floating around with a few additional columns (as far as I can tell, those extra columns are info that the data store team needs, but not the broker). Am confirming with @kaitlin but in the meantime can ask this: will the load process ignore any extra columns that aren't in our tas lookup data model?

cmc333333:
Re the new CARS file -- the new columns _should_ be ignored; I'll update this PR with the new file, though and try it out.

cmc333333:
@bsweger updated! I think at this point, looking at the combined diff is easiest -- the history thrashes between a few different solutions.

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -24,11 +29,12 @@ def test_loadTas_multiple(database, tmpdir):
     """If we have two rows in the CSV, we should have two TASLookups"""
     results = import_tas(
         database.session, tmpdir,
-        {'ATA': 'aaa', 'AID': 'bbb', 'A': 'ccc', 'BPOA': 'ddd', 'EPOA': 'eee',
-         'MAIN': 'ffff', 'SUB': 'ggg'},
-        {'ATA': '111', 'AID': '222', 'A': '333', 'BPOA': '444', 'EPOA': '555',
-         'MAIN': '6666', 'SUB': '777'}
+        {'ACCT_NUM': '6', 'ATA': 'aaa', 'AID': 'bbb', 'A': 'ccc',
```
bsweger: No doubt I'm missing something obvious, but I'm confused about what we're testing here. Is this test to ensure that the helper `import_tas` function works as expected before it's used in other tests? Does `import_tas` need to read the .csv you're creating using pandas to make sure all the NULLs and spaces arrive in the expected way?

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -45,3 +51,38 @@ def test_loadTas_space_nulls(database, tmpdir):
...
+
+    sess.add(TASFactory(tas_id=222))
+    sess.add(TASFactory(tas_id=444, agency_identifier='other'))
+    sess.commit()
+
+    results = sess.query(TASLookup).order_by(TASLookup.tas_id).all()
+    assert len(results) == 2
+    sess.invalidate()
+
+    loadTas.updateTASLookups('ignored')
```
bsweger: Thank you for entertaining my various unit testing questions. What does the `'ignored'` do here? Is that a signal to code readers that `updateTASLookups` has been monkey patched?

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -1,6 +1,11 @@
...
+from unittest.mock import Mock
 
+import pandas as pd
+
+from dataactcore.models.domainModels import TASLookup
 from dataactvalidator.scripts import loadTas
+from tests.unit.dataactcore.factories.domain import TASFactory
 
 
 def import_tas(session, tmpdir, *rows):
```
cmc333333: Good call; we do not

bsweger:
Thanks for the tweaks, @cmc333333! Looks great, and thanks for our first-ever unit test on a data load. 🎉

PR #362, opened 2016-10-18
Start resolving linting errors

cmc333333:
In #360, @bsweger found an undefined variable on a common path of the application. To prevent this sort of issue from coming up in the future, I propose we include flake8 (a Python linter) in our CI tests. This PR resolves two classes of linting errors (unused imports and undefined variables), and ignores dozens of others. Ideally, we'll stop ignoring more and more until we're up to flake8 standards.


tests/unit/dataactvalidator/test_a1_appropriations.py
```diff
@@ -56,4 +56,4 @@ def test_failure(database):
 
     # Non-overlapping ranges of agency IDs should generate two errors
     errors = number_of_errors(_FILE, database, models=[tas, tas_null, approp, approp_null])
-    assert errors == 2
\ No newline at end of file
+    assert errors == 2
```
jworcestBAH: There seem to be a bunch of these last line "edits" even on files where nothing was changed, am I missing something here?

dataactvalidator/filestreaming/csvS3Reader.py
```diff
@@ -34,11 +34,6 @@ def openFile(self, region, bucket, filename, csvSchema, bucketName, errorFilenam
         super(CsvS3Reader, self).openFile(
             region, bucket, filename, csvSchema, bucketName, errorFilename)
 
-    def downloadFile(self, region, bucket, filename, targetLocation):
-        """ After opening a file, download entire file to filename """
-        self.s3File = initializeFile(region, bucket, filename)
-        self.s3File.get_contents_to_filename(targetLocation)
-
```
jworcestBAH: Haven't been able to figure out what this function was used for, doesn't seem to be called anywhere so removing it shouldn't break anything.

tests/unit/dataactvalidator/test_a1_appropriations.py
jworcestBAH: I guess these are files that don't end with a blank line, put a question in the summary about that.

cmc333333: Flake8 has a rule that all files must end in a new line, but we're actually ignoring that at the moment. The rule stems from the history of Python: part of the posix standard is for all lines in a text file (including the last) to end in a newline character. This also means that unixy text editors (including vim) add the character automatically. I mass edited the unit tests (as several had unnecessary imports), so vim added the extra newline; I figured that it'd be easier to commit the changes (and prep for the flake8 rule) than undo them.

dataactvalidator/filestreaming/csvS3Reader.py
cmc333333: :+1: the function name doesn't appear anywhere else, so I think we're safe unless we're doing something super dynamic.

jworcestBAH:
Looks good, merging

PR #364, opened 2016-10-19
Db 582 display total obligations after submission

alburde1:
This pull request creates a new API route that returns the total obligations, total financial assistance obligations, and total procurement obligations for a given submission ID.


dataactcore/models/jobTrackerInterface.py
```diff
@@ -382,3 +384,24 @@ def checkJobType(self, jobId):
             # Wrong type
             raise ResponseException("Wrong type of job for this service", StatusCode.CLIENT_ERROR, None,
                                     ValidationError.jobError)
+
+    def getTotalObligations(self,submissionId):
+        """ Return sum of all obligations incurred """
+        query = self.session.query(func.sum(AwardFinancial.transaction_obligated_amou)).filter(AwardFinancial.submission_id == submissionId).scalar()
+        if query == None:
```
bsweger: Small thing to point out here and in the other new methods...the preferred way to check for `None` in Python is `if query is None` 

(reference: https://www.python.org/dev/peps/pep-0008/#programming-recommendations...2nd bullet re: comparison to singletons)

tests/integration/fileTests.py
```diff
@@ -342,6 +350,31 @@ def test_check_status(self):
...
+        """Test get obligations route."""
+        postJson = {"submission_id": self.obligations_submission_id}
+
+        response = self.app.post_json("/v1/get_obligations/", postJson, headers={"x-session-id": self.session_id})
+        json = response.json
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(json["total_obligations"], "5000")
+        self.assertEqual(json["total_procurement_obligations"], "2000")
+        self.assertEqual(json["total_assistance_obligations"], "3000")
```
cmc333333: I agree with @bsweger -- this'd be much better as a unit test (e.g. of the `obligationStatsForSubmission` function described above). In the current setup, to understand this test, I need to look at `setUpClass`, follow through to `setupObligations` and potentially `insertObligation`. These two tests as unit tests would look like:

```python
def test_obligationStatsForSubmission_nonzero(database):
    submission = SubmissionFactory()
    financials = [
        AwardFinancialFactory(transaction_amount=1000, piid="1234", fain=None, uri=None, 
                              submission=submission),
        AwardFinancialFactory(transaction_amount=1000, piid="1235", fain=None, uri=None, 
                              submission=submission),
        AwardFinancialFactory(transaction_amount=1000, piid=None, fain="1235", uri=None, 
                              submission=submission),
        AwardFinancialFactory(transaction_amount=1000, piid=None, fain=None, uri="1235", 
                              submission=submission),
        AwardFinancialFactory(transaction_amount=1000, piid=None, fain="1234", uri="1235", 
                              submission=submission),
    ]
    database.session.add_all([submission] + financials)
    database.session.commit()
    assert obligationStatsForSubmission(submission.submission_id) == {
        "total_obligations": "5000",
        "total_procurement_obligations": "2000",
        "total_assistance_obligations": "3000"
    }


def test_obligationStatsForSubmission_zero(database):
    submission = SubmissionFactory()
    # no financials in db
    database.session.add(submission)
    database.session.commit()
    assert obligationStatsForSubmission == {
        "total_obligations": "0",
        "total_procurement_obligations": "0",
        "total_assistance_obligations": "0"
    }
```

The benefit to this approach is that there's very little context needed to read the test from start to finish (we really only need to know that the `database` fixture sets up a database and that `Factory` creates populated models).

We still want to exercise the endpoint, though, so maybe this test would be simplified to:

```python
def test_get_obligations(self):
    submission = SubmissionFactory()
    self.session.add(submission)
    self.session.commit()
    response = self.app.post_json(
        "/v1/get_obligations/", 
        {"submission_id": submission.submission_id},
        headers={"x-session-id": self.session_id}
    )
    assert response.status_code == 200
    assert "total_obligations" in response.json
```

alburde1:
I THINK I fixed everything @cmc333333 commented on (thank you!). Also added documentation, as requested (oops).

cmc333333:
I think this is great, but am way too biased. @bsweger would you mind reviewing again?

tests/unit/dataactbroker/test_obligations.py
```diff
@@ -0,0 +1,39 @@
+from tests.unit.dataactcore.factories.job import SubmissionFactory
+from tests.unit.dataactcore.factories.staging import AwardFinancialFactory
+from dataactcore.models.jobTrackerInterface import obligationStatsForSubmission
+
+def test_obligationStatsForSubmission_nonzero(database):
+    submission = SubmissionFactory()
+    database.session.add(submission)
+    database.session.commit()
+    print(submission.submission_id)
```
bsweger: Is this a leftover debug?

tests/integration/fileTests.py
```diff
@@ -341,6 +343,14 @@ def test_check_status(self):
             self.assertEqual(json["created_on"], datetime.utcnow().strftime("%m/%d/%Y"))
             self.assertEqual(json["last_updated"], submission.updated_at.strftime("%Y-%m-%dT%H:%M:%S"))
 
+    def test_get_obligations(self):
```
bsweger: Thanks for being a great test case re: the unit vs integration test conversation!

dataactcore/models/jobTrackerInterface.py
```diff
@@ -1,14 +1,29 @@
...
 from dataactvalidator.validation_handlers.validationError import ValidationError
 
+def obligationStatsForSubmission(submission_id):
+    sess = GlobalDB.db().session
+    base_query = sess.query(func.sum(AwardFinancial.transaction_obligated_amou)).\
+        filter(AwardFinancial.submission_id == submission_id)
+    procurement = base_query.filter(AwardFinancial.piid != None)
+    fin_assist = base_query.filter(or_(AwardFinancial.fain != None, AwardFinancial.uri != None))
+    return {
+        "total_obligations": float(base_query.scalar() or 0),
```
bsweger: You and @cmc333333 spoke about this, so apologies for being behind, do we need to cast the return value at all here?

tests/integration/fileTests.py
```diff
@@ -33,6 +34,7 @@ def setUpClass(cls):
         with createApp().app_context():
             # get the submission test user
             sess = GlobalDB.db().session
+            cls.session = sess
```
alburde1: yes, we decided to move the creation of the submission into the test itself for this particular test so we could have it be at least slightly more autonomous and there would be less searching involved to find everything it was associated with. We couldn't pass sess to a random test so we had to make it a class variable.

dataactcore/models/jobTrackerInterface.py
alburde1: Yes, because it gets mad at trying to JSON serialize Decimal('NUMBER')

tests/unit/dataactbroker/test_obligations.py
alburde1: Oops, removed

tests/integration/fileTests.py
bsweger: Aaah, just looked at this again. I see what you did there! Thanks for the context.

bsweger:
@alburde1 Everything looks good!

PR #365, opened 2016-10-19
Hotfix: address a few things caught by the linter

bsweger:
This PR fixes a few things flagged by the linter that @cmc333333 added. As far as I can tell, both involved code that is not run.

While in the broker's app.py, I also removed the try/catch block from the broker's app.py.` createApp()` function with the idea that these kinds of non-specific, unanticipated exceptions should flow on their own to our logs.


dataactvalidator/filestreaming/csvS3Reader.py
```diff
@@ -34,11 +34,6 @@ def openFile(self, region, bucket, filename, csvSchema, bucketName, errorFilenam
         super(CsvS3Reader, self).openFile(
             region, bucket, filename, csvSchema, bucketName, errorFilename)
 
-    def downloadFile(self, region, bucket, filename, targetLocation):
-        """ After opening a file, download entire file to filename """
-        self.s3File = initializeFile(region, bucket, filename)
-        self.s3File.get_contents_to_filename(targetLocation)
-
```
jworcestBAH: This is removed in the other PR from the linter errors

bsweger: Nice! I'll back this out so it doesn't conflict with the other branch.

dataactbroker/app.py
```diff
@@ -15,97 +12,79 @@
...
-
-        return app
-
-    except Exception as e:
-        exc_type, exc_obj, exc_tb = sys.exc_info()
-        trace = traceback.extract_tb(exc_tb, 10)
-        CloudLogger.logError('Broker App Level Error: ', e, trace)
-
-        del exc_tb
-        raise
```
bsweger: Yes, that's my understanding too.

I'm trying to avoid this need to explicitly wrap any code that could possibly error out in a try/catch to make sure it goes through the CloudLogger. Paging @cmc333333 here, since I know he has some thoughts about a glorious future with streamlined logging. Is there a better way to do what I'm trying to do here?

jworcestBAH: Does anyone know offhand how much information winds up in the uwsgi log for uncaught errors?  Does that include the stack trace?  We should perhaps try to find an example in past logs to see how much uwsgi captures.

cmc333333: If I understand correctly, @jworcestBAH is correct -- application initialization failures would be captured in uwsgi log rather than CloudLogger. That seems like an okay thing, however, given that the exception would be raised when the application was _initializing_. uwsgi logs should include the traceback, yes (it'd be on separate lines, however). In general, I'm a big proponent of ripping out CloudLogger wherever possible and preferring the more standard (and configurable) `logging` system, or (as in this case), deferring to the underlying runner (uwsgi, celery, etc.) for log capture.

cmc333333: @jworcestBAH bump

nmonga91:
Looks good to me!

PR #367, opened 2016-10-19
A33 Implementation Fix

mab6bf:


tests/unit/dataactvalidator/test_a33_appropriations_1.py
```diff
@@ -18,38 +18,70 @@ def test_column_headers(database):
...
+
     sf1 = SF133Factory(tas=tas, period=period, fiscal_year=year,
-                       agency_identifier=code)
+                       allocation_transfer_agency=code, agency_identifier=code)
     sf2 = SF133Factory(tas=tas, period=period, fiscal_year=year,
-                       agency_identifier=code)
-    submission = SubmissionFactory(
-        submission_id=submission_id, reporting_fiscal_period=period,
-        reporting_fiscal_year=year, cgac_code=code)
+                       allocation_transfer_agency=code, agency_identifier=code)
```
jworcestBAH: Are sf1 and sf2 the same, I'm not seeing the difference here?  Also, we may want to test with allocation_transfer_agency and agency_identifier having different values to make sure the logic in the rule works the way we expect.  So basically a case where ATA matches but AID does not, and a case where ATA is null and AID matches.

jworcestBAH: The null ATA case is tested below, so maybe just change the AID here to not match so we're sure it's using the ATA

jworcestBAH:
Looks good, merging

PR #370, opened 2016-10-21
Db 1237 max login

nmonga91:
Integrates MAX SSO login functionality.

- [x] Reviewed by back-end
- [x] Reviewed by front-end to ensure API contract is upheld
- [x] Coordination for deployment complete (config updates, etc)
- [x] MAX dev/staging environment enabled `usaspending.gov` wildcard to allow access
- [x] All individuals requiring website access have MAX accounts and appropriate group membership


dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+                                                       "name": user.name, "title": user.title,
+                                                       "agency_name": agency_name,
+                                                       "cgac_code": user.cgac_code, "permissions": permissionList})
+
+        except (TypeError, KeyError, NotImplementedError) as e:
+            # Return a 400 with appropriate message
+            return JsonResponse.error(e,StatusCode.CLIENT_ERROR)
+        except ValueError as e:
+            # Return a 401 for login denied
+            return JsonResponse.error(e,StatusCode.LOGIN_REQUIRED)
```
kevinli-work: Do these `JsonResponse.error` functions return
```
{"message": "User displayable message"}
```
?

nmonga91: Good question, I will test to ensure it does. The old `login` had it so I was assuming it does, but I will check regardless.

nmonga91: @kevinli-work yes it does

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+            # Call MAX's serviceValidate endpoint and retrieve the response
+            url = CONFIG_BROKER['cas_service_url'].format(ticket, service)
+            max_xml = requests.get(url).content
+            max_dict = xmltodict.parse(max_xml)
+
+            if not 'cas:authenticationSuccess' in max_dict['cas:serviceResponse']:
+                raise ValueError("You have failed to login successfully with MAX")
+
+            # Grab the email and list of groups from MAX's response
+            email = max_dict['cas:serviceResponse']['cas:authenticationSuccess']['cas:attributes']['maxAttribute:Email-Address']
```
bsweger: Are all of these keys (`cas:serviceReasponse`, `cas:authenticationSuccess`, `cas:attributes`, `maxAttribute:Email-Address`) guaranteed to be present if the check for `cas:AuthenticationSuccess` above passes? Do we need additional checking for the possibility of missing information, or will these always be present as part of the contract with this service?

bsweger: (I'm all for not wrapping this up in additional error handling if the service says the info should be there, am just asking because I haven't read the relevant MAX docs)

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+            group_list_all = max_dict['cas:serviceResponse']['cas:authenticationSuccess']['cas:attributes']['maxAttribute:GroupList'].split(',')
+            group_list = list(filter(lambda g:  g.startswith(parent_group), group_list_all))
```
bsweger: Perhaps just a style preference and not something that should hold up a merge, but I find simple list comprehensions to be easier to read than using a combination of filter and lambda and then casting the result to a list. For example, something like:
`group_list = [g for g in group_list_all if g.startswith(parent_group)]`

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+            # Deny access if not in the parent group aka they're not allowed to access the website all together
+            if not parent_group in group_list:
+                raise ValueError("You have logged in with MAX but do not have permission to access the broker. Please "
+                                 "contact DATABroker@fiscal.treasury.gov to obtain access.")
+
+            cgac_group = list(filter(lambda g: len(g) == len(parent_group + "-CGAC_")+3, group_list))
```
bsweger: Same question here about whether or not a list comprehension would be easier to read here.

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+            # Deny access if they are not aligned with an agency
+            if not cgac_group:
+                raise ValueError("You have logged in with MAX but do not have permission to access the broker. Please "
+                                 "contact DATABroker@fiscal.treasury.gov to obtain access.")
+
+            try:
+                sess = GlobalDB.db().session
```
bsweger: Thanks for using the new db access style!

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+                    user.email = email
+
+                    # Check for None first so the condition can short-circuit without
+                    # having to worry about calling strip() on a None object
+                    if middle_name is None or middle_name.strip() == '':
+                        user.name = first_name + " " + last_name
+                    else:
+                        user.name = first_name + " " + middle_name[0] + ". " + last_name
+                    user.user_status_id = UserHandler().getUserStatusId('approved')
```
bsweger: This is a great reminder that I should communicate this more broadly--apologies! Part of the db refactor is using a new series of lookup dictionaries for our "helper tables". So for this, you can:
```python
from dataactcore.models.lookups import USER_STATUS_DICT
...
user.user_status_id = USER_STATUS_DICT['approved']
```

Essentially, we're now using a pre-defined set of key/value pairs to both load these database tables and to access the info, bypassing the db entirely (though we still load it for reference and to use for ad-hoc queries)

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+                        'maxAttribute:Middle-Name']
+                    last_name = max_dict["cas:serviceResponse"]['cas:authenticationSuccess']['cas:attributes'][
+                        'maxAttribute:Last-Name']
+
+                    user.email = email
+
+                    # Check for None first so the condition can short-circuit without
+                    # having to worry about calling strip() on a None object
+                    if middle_name is None or middle_name.strip() == '':
+                        user.name = first_name + " " + last_name
```
bsweger: Will there be any unwanted side effects from having a space in our username?

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+                    else:
+                        user.name = first_name + " " + middle_name[0] + ". " + last_name
+                    user.user_status_id = UserHandler().getUserStatusId('approved')
+
+                    # If part of the SYS agency, use that as the cgac otherwise use the first agency provided
+                    if list(filter(lambda g: g.endswith("SYS"), cgac_group)):
```
bsweger: Same question about list comprehension vs. filter/lambda. Again, not something huge, and would well be a personal preference. Just want to make sure both techniques are out there for consideration.

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+                        user.cgac_code = "SYS"
+                        # website admin permissions
+                        user.permissions = 2
```
bsweger: We should try to do this without using a magic number. Is that possible?

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+                    else:
+                        user.cgac_code = cgac_group[0][-3:]
+                        # regular user permissions
+                        user.permissions = 1
```
bsweger: Same comment re: magic number.

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+                    sess.add(user)
+                    sess.commit()
+
+            except MultipleResultsFound:
+                raise ValueError("An error occurred during login.")
```
bsweger: Will raising our own error here, without included the original error/traceback, suppress any info we would need to debug this scenario?

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+            # Create session
+            LoginSession.login(session, user.user_id)
```
bsweger: This looks similar to the [session creation code](https://github.com/fedspendingtransparency/data-act-broker-backend/blob/00bed561a97a22e1c858e2e7a7e91f80f31e22fe/dataactbroker/handlers/accountHandler.py#L110) in the old login route. Do you think it's worth abstracting session creation into its own function?

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+            permissionList = []
+            for permission in self.interfaces.userDb.getPermissionList():
```
bsweger: Again, my fault for not sharing this info more widely. Instead of using the `getPermissionList()` to get a list of permission, you can do `from dataactcore.models.lookups import PERMISSION_TYPE_DICT` to access a dictionary of permission type ids and values.

dataactbroker/handlers/accountHandler.py
```diff
@@ -137,6 +144,117 @@ def login(self,session):
...
+
+        arguments:
+
+        session  -- (Session) object from flask
+
+        return the reponse object
+
+        """
+        try:
+            safeDictionary = RequestDictionary(self.request)
```
nmonga91: We need this to process the POST contents from the front-end. It's not the response we're getting directly from MAX. The front-end does the initial request, gets the payload, then forwards it to the broker.

dataactbroker/handlers/accountHandler.py
nmonga91: They are supposed to be there, if they're not, there'll be an exception and we can handle it then with MAX. So I don't think we need to do too much error handling with those keys. If you think we should incorporate some, we can but I think the exception handling we have should be sufficient.

dataactbroker/handlers/accountHandler.py
nmonga91: I like the simplicity of that. I think i was just getting too fancy here :)

dataactbroker/handlers/accountHandler.py
nmonga91: I'll opt for simplicity

dataactbroker/handlers/accountHandler.py
nmonga91: Yes we can do a permission lookup, I'll add that in.

dataactbroker/handlers/accountHandler.py
nmonga91: Raising the ValueError here will result in a stack trace which should lead us here, so we should be able to trace where the error came from.

dataactbroker/handlers/accountHandler.py
nmonga91: Totally agree with the abstraction. This is exactly what I wanted to get out of the review. Thank you!

dataactbroker/handlers/accountHandler.py
bsweger: Thanks for clarifying--all good!

dataactbroker/handlers/accountHandler.py
bsweger: Sorry for not being clear. What I'm questioning is...why does the broker need a special class like this to process an incoming request, rather than using whatever request-process means are included in the Flask framework?

bsweger: FYI, since this question is out of scope for the PR (I was just pulling a thread 😄), we can revisit later.

dataactbroker/handlers/accountHandler.py
bsweger: If we had spaces before, this should be fine. Just wanted to make sure!

dataactbroker/README.md
```diff
@@ -56,6 +56,37 @@ Example output:
 
 ### User Routes
 
+#### POST "/v1/max_login/"
```
bsweger: Thanks for updating the docs!

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -0,0 +1,103 @@
...
+
+    mock_dict = Mock()
+    mock_dict.return_value.safeDictionary.side_effect = {'ticket': '', 'service': ''}
+    monkeypatch.setattr(dataactbroker.handlers.accountHandler, 'RequestDictionary', mock_dict)
+
+    max_dict= {'cas:serviceResponse': {}}
+    monkeypatch.setattr(ah, 'get_max_dict', Mock(return_value=max_dict))
+    config = {'parent_group': 'parent-group'}
+    monkeypatch.setattr(dataactbroker.handlers.accountHandler, 'CONFIG_BROKER', config)
+    max_dict = {
```
bsweger: Do you think `max_dict` is a candidate for a fixture that can be updated as needed for the individual scenarios? Or is that overkill since it's a resource that will be used in a very limited number of tests?

nmonga91: That may be overkill as this is the only test that needs it. I would opt for a fixture if other tests could also use it.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -0,0 +1,107 @@
...
+            }
+    }
+    monkeypatch.setattr(ah, 'get_max_dict', Mock(return_value=max_dict))
+
+    # If it gets to this point, that means the user was in all the right groups aka successful login
+    monkeypatch.setattr(ah, 'create_session_and_response',
+                        Mock(return_value=JsonResponse.create(StatusCode.OK, {"message": "Login successful"})))
+    json_response = ah.max_login(Mock())
+
+    print(json.loads(json_response.get_data().decode("utf-8")))
```
nmonga91: Good catch, it is indeed leftover from debugging.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -0,0 +1,107 @@
...
+    assert "Login successful" == json.loads(json_response.get_data().decode("utf-8"))['message']
+
+
+def test_max_login_failure(monkeypatch):
+    ah = dataactbroker.handlers.accountHandler.AccountHandler(Mock())
+
+    mock_dict = Mock()
+    mock_dict.return_value.safeDictionary.side_effect = {'ticket': '', 'service': ''}
+    monkeypatch.setattr(dataactbroker.handlers.accountHandler, 'RequestDictionary', mock_dict)
```
nmonga91: Mock is necessary to mock the RequestDictionary class. Since we'll be revisiting that in the future, I didn't worry too much about it during testing.

bsweger:
Everything on the list is complete--let's do it!

PR #372, opened 2016-10-25
Unset any submissions with invalid users before fk

cmc333333:
This modifies an existing migration which would fail if a submission existed
with a bad user id. Now that migration updates the submissions prior to adding
the fk.

For environments where this migration passed successfully before, this
execution doesn't change anything.


jworcestBAH:
Looks good, will merge as soon as Jenkins finishes

PR #373, opened 2016-10-25
Remove several unused functions

cmc333333:
While tracking down TASLookups, I found a function that was never called. There's a neat tool called [vulture](https://pypi.python.org/pypi/vulture) which highlights potentially unused functions, variables, classes, etc. I went through its results for functions and classes, inspected each (grepping for the string in our code base) to weed out false positives, and deleted the remaining. It's possible these are still accessed indirectly (e.g. via `globals()` or `getattr`), so a second set of eyes would be appreciated.


dataactcore/models/jobTrackerInterface.py
```diff
@@ -1,10 +1,12 @@
 import traceback
-from uuid import uuid4
 from sqlalchemy import func, or_
 from sqlalchemy.orm import joinedload
+
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.models.baseInterface import BaseInterface
-from dataactcore.models.jobModels import Job, JobDependency, JobStatus, JobType, Submission, FileType, PublishStatus, FileGenerationTask
+from dataactcore.models.jobModels import (
```
bsweger: Huh, didn't know you could use `(` in an import statement. Thanks!

cmc333333:
@bsweger I removed several of those ignores, but not all. I played it safe and left some that looked like they might still be referenced:
* `responseErrorLog` in dataactcore/utils/jsonResponse.py
* `test-reports` in tests/intergation/runTests.py
* `validatorResponse` in dataactbroker/handlers/fileHandler.py
* `responseErrorLog` in dataactcore/utils/jsonResponse.py
* `errorLog` in dataactvalidator/app.py
* `dbErrors.log` in dataactcore/models/baseInterface.py

bsweger:
@cmc333333 Thank you, thank you!

PR #374, opened 2016-10-26
Updating A33

mab6bf:
- Updated A33_2 to exclude results in which all monetary amounts are 0 for a specific TAS
- Modified both A33 unit test files to use FactoryBoy
- Fixed an extra "." in the C8 error description I noticed during my demo


bsweger:
For transparency, @mab6bf and I chatted about the test data in Slack...he did a good job explaining the philosophy of unit test data (it was a philosophical question to begin with...never a blocker for merging). Let's do this...

PR #375, opened 2016-10-26
Dev Stable -> Staging

cmc333333:


bsweger:
Looks good. Fly away to staging, sprint 24!

PR #376, opened 2016-10-27
Db 1199 part 1

alburde1:
- Changed a lot of variable names from camelCase to underscored (only in functions I was already updating for the actual changes)
- Removed many functions from errorInterface.py (some completely factored out, some moved to function_bag.py) and updated related function calls to use the function_bag functions instead of relying on the interface.
- Updated several calls to directly look up from the dicts instead of calling functions to do so
- Replaced several functions to just query using sqlalchemy directly, eliminating extra function calls and levels of obscurity
- Connected all updated functions to GlobalDB.db().session instead of specific error interfaces


dataactcore/interfaces/function_bag.py
```diff
@@ -243,3 +246,78 @@ def addJobsForFileType(fileType, filePath, filename, submissionId, existingSubmi
...
+    fileRec.file_status_id = FILE_STATUS_DICT[ValidationError.getErrorTypeString(error_type)]
+    if extra_info is not None:
+        if "missing_headers" in extra_info:
+            fileRec.headers_missing = extra_info["missing_headers"]
+        if "duplicated_headers" in extra_info:
+            fileRec.headers_duplicated = extra_info["duplicated_headers"]
+
+    sess.add(fileRec)
+    sess.commit()
+    return True
```
cmc333333: I know this return value existed before you tackled this code, but if we're reworking this function now -- does it make sense to remove the `return`? It doesn't seem to be providing any value (the only other way out of this function is an exception).

dataactcore/interfaces/function_bag.py
```diff
@@ -243,3 +246,93 @@ def addJobsForFileType(fileType, filePath, filename, submissionId, existingSubmi
...
+        filename: name of error report in S3
+
+    Returns:
+        True if successful
+    """
+    sess = GlobalDB.db().session
+    fileComplete = createFileIfNeeded(job_id, filename)
+    fileComplete.file_status_id = FILE_STATUS_DICT['complete']
+    sess.commit()
+    return True
```
cmc333333: Similar note here about return value. In both cases, we'd want to update the docstring, too.

dataactcore/interfaces/function_bag.py
alburde1: Updated this and the below comment

dataactbroker/handlers/fileHandler.py
```diff
@@ -339,7 +339,7 @@ def getStatus(self):
             submission = self.jobManager.getSubmissionById(submission_id)
 
             # Check that user has access to submission
-            user = self.checkSubmissionPermission(submission)
```
cmc333333: Good call

cmc333333:
Excellent work

PR #377, opened 2016-10-27
Remove validate_threaded route

bsweger:
This is something we'd discussed a while back. The validate_threaded
route is not currently in use. If we ever need this in the future,
we'd likely want to revist the approach because so much has changed
since the route was originally implemented (e.g., single code
base, migration to Python 3). In the meantime, though, this commit
removes the route entirely so we don't need to update it as part
of the ongoing database refactor.


cmc333333:
Makes sense to me and I don't see any references in data-act-broker-backend or data-act-broker-web-app

PR #378, opened 2016-10-28
Db 301 remove dynamo

jworcestBAH:
This PR replaces dynamo session handling with a session table in the postgres RDS


dataactbroker/handlers/aws/session.py
```diff
@@ -387,14 +307,15 @@ def newSession(uid,data,expiration) :
...
-            item[SessionTable.DATE_FIELD] = toUnixTime(expiration)
-            item.save(overwrite=True)
-        else :
-            SessionTable.getTable().put_item(data={
-                SessionTable.KEY_NAME: uid,
-                SessionTable.DATA_FIELD: data,
-                SessionTable.DATE_FIELD: toUnixTime(expiration)
-            })
+        sess = GlobalDB.db().session
+        user_session = sess.query(SessionMap).filter_by(uid=uid).first()
```
bsweger: I'm not sure of the expected data in SessionMap, but if you want to make sure there's only one row per uid (maybe you don't!), you could use the SqlAlchemy `one_or_none()` function.

dataactbroker/app.py
```diff
@@ -50,7 +46,7 @@ def createApp():
         cors = CORS(app, supports_credentials=False, origins=CONFIG_SERVICES['cross_origin_url'],
                     allow_headers = "*", expose_headers = "X-Session-Id")
     # Enable AWS Sessions
```
bsweger: Very small nit--should this comment still say `AWS Sessions` or something else now?

dataactbroker/app.py
```diff
@@ -103,7 +83,7 @@ def runApp():
 
 if __name__ == '__main__':
     runApp()
-    proc = multiprocessing.Process(target=checkDynamo)
+    proc = multiprocessing.Process()
```
bsweger: Can we get rid of this line (and the next 2) entirely? It looks like @nmonga91 added them as part of the "check if dynamo is running on startup" story: https://github.com/fedspendingtransparency/data-act-broker-backend/commit/5db468d38b1da008e5af6cbdf14a5c0afc1aa14c

dataactbroker/handlers/aws/session.py
```diff
@@ -152,30 +146,20 @@ def toUnixTime(datetimeValue) :
...
+    if(type(datetimeValue) == datetime):
+        # If argument is a datetime object, convert to timestamp
+        return (datetimeValue-datetime(1970,1,1)).total_seconds()
+    return datetimeValue
 
 
 
-class DynamoSession(CallbackDict, SessionMixin):
+
+class UserSession(dict, SessionMixin):
```
bsweger: What do you think about putting a `pass` statement here to make it clear that this is intentionally left blank [per the docs](http://flask.pocoo.org/docs/0.11/api/#session-interface)?

tests/integration/runTests.py
```diff
@@ -39,7 +39,7 @@ def runTests(argv=''):
     suite.addTests(unittest.makeSuite(MixedFileTests))
 
     # to run a single test:
-    #suite.addTest(FileTests('test_file_generation'))
+    #suite.addTest(LoginTests('test_login'))
```
bsweger: Did you mean to check this in as part of the PR?

dataactbroker/handlers/aws/session.py
```diff
@@ -1,14 +1,10 @@
...
-from boto.dynamodb2.fields import HashKey, GlobalAllIndex
-from boto.dynamodb2.table import Table, exceptions
-from boto.dynamodb2.types import NUMBER
 from datetime import datetime, timedelta
 from flask.sessions import SessionInterface, SessionMixin
-from werkzeug.datastructures import CallbackDict
-from boto.dynamodb2 import connect_to_region
-from boto.dynamodb2.layer1 import DynamoDBConnection
 from flask_login import _create_identifier
+from dataactcore.models.baseInterface import GlobalDB
```
jworcestBAH: Fixed

dataactbroker/handlers/aws/session.py
jworcestBAH: Added a pass

dataactbroker/handlers/aws/session.py
jworcestBAH: Good idea, switched to one_or_none

dataactcore/read_config.py
```diff
@@ -134,7 +124,10 @@
         category = CONFIG_CATEGORIES[category_yaml_name]
         category_message = "### {}".format(category_yaml_name)
         for key in CONFIG_LOGGING["values_to_log"][category_yaml_name]:
-            value = category[key]
+            if key not in category:
```
jworcestBAH: Thanks, fixed

dataactbroker/app.py
jworcestBAH: Switched to "Enable DB session table handling" to cover both RDS and local cases

dataactbroker/app.py
jworcestBAH: Paging @nmonga91 on this one as I'm not actually sure what these are doing.  This code is not used in the AWS version of the app or the unit tests, but should still be used by local installs.

tests/integration/runTests.py
jworcestBAH: Ah, no I did not, will revert that.

dataactbroker/app.py
cmc333333: Agreed that this can be removed; no need for a separate process anymore (though it might not have been hit while the app was running anyway?)

dataactbroker/handlers/aws/session.py
```diff
@@ -152,30 +146,20 @@ def toUnixTime(datetimeValue) :
 
     returns int
     """
-    if(type(datetimeValue) == Decimal):
-        # If argument is already Decimal, assume it's been converted previously
-        return datetimeValue
-    return (datetimeValue-datetime(1970,1,1)).total_seconds()
+    if(type(datetimeValue) == datetime):
```
cmc333333: Small point, generally `isinstance` is preferred as it'll handle subclasses:
```python
if isinstance(datetimeValue, datetime):
```

dataactbroker/app.py
bsweger: @jworcestBAH This was code that Nipun added specifically for local broker instances. It uses a separate process to do a Dynamo call...if nothing comes back in 5 seconds (or whatever the time is), the app returns reminder to start dynamo. We added this code b/c the local app would just hang with no messages if Dynamo wasn't started.

dataactbroker/handlers/aws/session.py
```diff
@@ -198,18 +183,27 @@ def open_session(self, app, request):
...
 
         """
         sid = request.headers.get("x-session-id")
         if(sid and SessionTable.doesSessionExist(sid)):
             if SessionTable.getTimeout(sid)> toUnixTime(datetime.utcnow()):
-                return DynamoSession(initial=SessionTable.getData(sid),sid=sid)
+                session_dict =  UserSession()
+                # Replace single quotes with double and read as dict, also need to correct True, False, and None for json
+                data_string = SessionTable.getData(sid).replace("'",'"').replace("True","true")\
+                    .replace("False","false").replace("None","null")
```
cmc333333: This is the beginning of a custom serialization/deserialization protocol. Why not just store the data in JSON? Also, I'm pretty sure you can kick all that work over to Sqlalchemy if using [JSON](http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.JSON) or [PickleType](http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.PickleType)

dataactbroker/handlers/aws/session.py
```diff
@@ -198,18 +183,27 @@ def open_session(self, app, request):
...
+                data = loads(data_string)
+                for key in data.keys():
+                    session_dict[key] = data[key]
+                session_dict["sid"] = sid
```
cmc333333: This line shouldn't be doing anything, right? It should have be in the serialized data.

dataactbroker/handlers/aws/session.py
```diff
@@ -364,21 +287,19 @@ def getTimeout(uid) :
...
 
     @staticmethod
     def getData(uid) :
         """
         uid -- (String) the uid
         return (Session) the session data
         """
-        response = SessionTable.getTable().get_item(uid=uid)
-        return response[SessionTable.DATA_FIELD]
+        return GlobalDB.db().session.query(SessionMap).filter_by(uid=uid).one().data
```
cmc333333: What prevents a user from having more than one session?

dataactcore/models/userModel.py
```diff
@@ -62,3 +62,11 @@ class EmailToken(Base):
     email_token_id = Column(Integer, primary_key=True)
     token = Column(Text)
     salt = Column(Text)
+
+class SessionMap(Base):
+    """ This table maps session IDs to user data """
+    __tablename__ = "session_map"
+    session_id = Column(Integer, primary_key=True)
+    uid = Column(Text)
```
cmc333333: Looks like we query by uid often, too. Should it have an index?

cmc333333:
Good stuff! I think we should tighten up the marshalling/unmarshalling, but either way, this is much cleaner than requiring Dynamo.

dataactbroker/app.py
jworcestBAH: Okay, have removed the proc stuff so running locally now just hits runApp()

dataactbroker/handlers/aws/session.py
jworcestBAH: Yep, removed this

dataactcore/models/userModel.py
jworcestBAH: Good point, added an index on this

dataactcore/read_config.py
```diff
@@ -134,7 +124,7 @@
         category = CONFIG_CATEGORIES[category_yaml_name]
         category_message = "### {}".format(category_yaml_name)
         for key in CONFIG_LOGGING["values_to_log"][category_yaml_name]:
-            value = category[key]
+            value = category.get(key,"Value not provided in config")
```
jworcestBAH: Can you clarify what behavior you'd like here?  Currently this is written to the log as "... , key: Value not provided in config, ..." where the ... is messages for other keys.  Would you like that written in a different way?

cmc333333: Oof, my mistake, I thought we were using these values for business logic

cmc333333:
I think we're good to go; nice swap @jworcestBAH !

PR #380, opened 2016-10-29
Remove nonpytest runner

cmc333333:
This also makes it possible to just run `py.test` from the root of the project without errors.


jworcestBAH:
Looks good, merging

PR #382, opened 2016-10-31
Update ErrorInterface and finish moving functions to function_bag

alburde1:
- Move functions from ErrorHandler to function_bag
- Separate ErrorInterface from BaseInterface and create instances to pass around where recording rowErrors is needed
- Fixing variables for functions I was already editing to underscored instead of camelCase
- Fixing some linting errors


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -144,123 +147,125 @@ def readRecord(self,reader,writer,fileType,interfaces,rowNumber,jobId,fields):
...
+            job_id: ID of current job
+            submission_id: ID of current submission
+            passed_validations: True if record has not failed first validations
             interfaces: InterfaceHolder object
             writer: CsvWriter object
-            rowNumber: Current row number
-            fileType: Type of file for current job
+            row_number: Current row number
+            file_type: Type of file for current job
+            error_list: list of errors found during this validation
```
bsweger: Same note as above re: describing `error_list` as a list.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -144,123 +147,125 @@ def readRecord(self,reader,writer,fileType,interfaces,rowNumber,jobId,fields):
...
+            record["job_id"] = job_id
+            record["submission_id"] = submission_id
+            record["valid_record"] = passed_validations
+            stagingInterface.insertSubmissionRecordByFileType(record, file_type)
+        except ResponseException:
             # Write failed, move to next record
-            writer.write(["Formatting Error", ValidationError.writeErrorMsg, str(rowNumber),""])
-            errorInterface.recordRowError(jobId, self.filename,
-                "Formatting Error",ValidationError.writeError, rowNumber,severity_id=interfaces.validationDb.getRuleSeverityId("fatal"))
+            writer.write(["Formatting Error", ValidationError.writeErrorMsg, str(row_number),""])
```
bsweger: Old code again, but I don't think we need to cast `row_number` to str because it looks like the `writer.write` method is doing that.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -144,123 +147,125 @@ def readRecord(self,reader,writer,fileType,interfaces,rowNumber,jobId,fields):
...
             interfaces: InterfaceHolder object
-            jobId: ID of current job
-            shortColnames: Dict mapping short names to long names
+            job_id: ID of current job
+            short_colnames: Dict mapping short names to long names
             writer: CsvWriter object
-            rowNumber: Current row number
+            warning_writer: CsvWriter object
+            row_number: Current row number
+            error_list: list of errors found during this validation
```
bsweger: Same comment as above--sorry to sound like a broker record!

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -144,123 +147,125 @@ def readRecord(self,reader,writer,fileType,interfaces,rowNumber,jobId,fields):
...
+        return fatal_error_found
 
-    def runValidation(self, jobId, interfaces):
+    def runValidation(self, job_id, interfaces):
         """ Run validations for specified job
         Args:
-            jobId: Job to be validated
-            jobTracker: Interface for job tracker
+            job_id: Job to be validated
+            interfaces: All interfaces
```
bsweger: Thanks for making this more accurate!

dataactcore/interfaces/function_bag.py
```diff
@@ -327,4 +327,27 @@ def markFileComplete(job_id, filename=None):
...
-    sess.commit()
\ No newline at end of file
+    sess.commit()
+
+def getErrorMetricsByJobId(job_id, include_file_types=False, severity_id=None):
+    """ Get error metrics for specified job, including number of errors for each field name and error type """
+    sess = GlobalDB.db().session
+    result_list = []
+
+    query_result = sess.query(File).options(joinedload("file_status")).filter(File.job_id == job_id).one()
```
alburde1: I think that, for now, we can leave it as-is. It's about a 50/50 split of needing file_status and not needing it so I wouldn't waste the extra processing power lazily adding it every time. Maybe in the future it would be useful but for now I'd value speed over the few times we have to add that extra option in text

dataactcore/interfaces/function_bag.py
```diff
@@ -327,4 +327,27 @@ def markFileComplete(job_id, filename=None):
...
+
+    if not query_result.file_status.file_status_id == FILE_STATUS_DICT['complete']:
+        return [{"field_name": "File Level Error", "error_name": query_result.file_status.name,
+                 "error_description": str(query_result.file_status.description), "occurrences": 1, "rule_failed": ""}]
```
alburde1: Taking it out doesn't break anything so it'll be out in the next commit

bsweger:
@alburde1 Thanks for the changes. Merging!

PR #383, opened 2016-11-01
TAS History

cmc333333:
I apologize ahead of time for the size of the PR -- it needed to change several things at once.

We're looking to keep a running history of our TAS imports. The TAS file will include the _current_ valid TASes, but we'll have submissions from the past, so instead of deleting the TAS entries, we'll mark their dates of activity. `test_updateTASLookups` is an integration test that hits several of the cases (a TAS which fell off the import file, one that remained unchained, one that has been modified, and one that is completely new) if you want to check my understanding of the business logic. This PR also fixes a few odds and ends.

Note that this does _not_ modify the TAS validations; that'll be a separate PR.


tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -56,12 +59,14 @@ def test_cleanTas_space_nulls(tmpdir):
...
+    """Verify that TAS with the same account_num can be modified, that we
+    "close" any non-present TASes, and that we add new entries"""
     sess = database.session
     monkeypatch.setattr(
         loadTas, 'cleanTas', Mock(return_value=pd.DataFrame({
             'availability_type_code': ['0', '1', '2'],
-            'tas_id': [111, 222, 333],
-            'agency_identifier': ['0', '1', '2'],
+            'account_num': [111, 222, 333],
+            'agency_identifier': ['new', '1', 'new'],
```
alburde1: Why are some identifiers "new" and some numbers? Below you also set some to "old"

dataactcore/models/domainModels.py
```diff
@@ -24,8 +24,7 @@ class TASLookup(Base) :
       TASLookup.ending_period_of_availability,
       TASLookup.availability_type_code,
       TASLookup.main_account_code,
-      TASLookup.sub_account_code,
-      unique=True)
+      TASLookup.sub_account_code)
```
bsweger: In many of our other domain values tables, we use unique indexes as a safety net for keeping the data clean. 

It makes sense to keep this non-unique index on the tas components; do you think it's necessary to have an second, unique index on the tas + the internal date end field? Or does a combination of carefully written load code + tests suffice to keep things clean?

This is a broader question for data pipelines in general--does everyone enforce data integrity these days?

tests/unit/conftest.py
```diff
@@ -26,13 +26,13 @@ def full_database_setup():
     dataactcore.config.CONFIG_DB = config
 
     createDatabase(config['db_name'])
-    db = dbConnection()
+    db = GlobalDB.db()
```
bsweger: Good catch--thanks!

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -88,3 +91,33 @@ def test_updateTASLookups(database, monkeypatch):
...
+    """We backdate all of the TASLookups when there aren't any in the
+    database"""
+    assert database.session.query(TASLookup).count() == 0
+    data = pd.DataFrame({'dummy': ['data']})
+    loadTas.add_start_date(data)
+
+    assert data['internal_start_date'][0] == date(2015, 1, 1)
+
+
+@pytest.mark.parametrize("today_date,fiscal_date", [
```
bsweger: Had to google `pytest.mark.parametrize` - nifty!

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -88,3 +91,33 @@ def test_updateTASLookups(database, monkeypatch):
...
+    ('2016-08-08', date(2016, 7, 1)),
+    ('2016-12-31', date(2016, 10, 1)),
+])
+def test_add_start_date_fiscal_year(database, today_date, fiscal_date):
+    """We should be calculating the correct beginning of a fiscal year"""
+    database.session.add(TASFactory())
+    database.session.commit()
+
+    data = pd.DataFrame({'dummy': ['data']})
+    with freeze_time(today_date):
```
bsweger: Traveling through time is also nifty. Thanks for continuing to introduce these helpful tools to the code base!

tests/unit/dataactvalidator/test_scripts_loadTas.py
cmc333333: The no-doubt-unclear idea was to emphasize that these elements weren't in the original dataset. Let me rework the order of the calls here a little to try to make the logic clearer.

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -73,24 +75,43 @@ def test_updateTASLookups(database, monkeypatch):
...
-    assert results[2].tas_id == 333
-    assert results[2].agency_identifier == '2'
+    sess.invalidate()
+    results = sess.query(TASLookup).\
+        order_by(TASLookup.account_num, TASLookup.agency_identifier).all()
+    assert len(results) == 5
+    t111, t222, t333_new, t333_old, t444 = results
+
+    assert t111.account_num == 111
+    assert t111.internal_end_date == None               # still active
```
bsweger: Making sure I'm following along here...is `t111` `# still active` or is it an entirely new, active TAS? If it's the latter, does it make sense to tweak the comment to differentiate this from the the `t222` case? (something like `# new, active TAS`).

tests/unit/dataactvalidator/test_scripts_loadTas.py
```diff
@@ -59,12 +59,14 @@ def test_cleanTas_space_nulls(tmpdir):
 
 
 def test_updateTASLookups(database, monkeypatch):
```
bsweger: This test is great. My only overall suggestion might be to make the setup a bit more explicit for the benefit of people who might be familiar with how the TAS load is supposed to work but less familiar with the constructs of unit testing (people like me _cough cough_).

For example, does it make sense to do something like:

```python
incoming_tas_data = pd.DataFrame({
'availability_type_code': ['0', '1', '2'],
'account_num': [111, 222, 333],
'agency_identifier': ['new', '1', 'new'],
etc.
})

existing_tas_data = the list of TAS factories from line 80

# add existing tas to db
sess.add_all(the list of TAS factories)

# monkeypatch the TAS load to return the mocked incoming .csv values
monkeypath.setattr(loadTas, 'cleanTas, Mock(etc etc)
```

Feel free to ignore this suggestion if it's too rudimentary, but I think it easier to think of these tests in very basic terms like "here is the new stuff" and "here is the stuff that's already in the tas table"

dataactcore/migrations/versions/4dc597500734_add_tas_history_fields.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # Reset tas_id values; we bump up all the existing tas_ids first to
+    # prevent duplicates during the renumber
+    max_tas_id = conn.execute(
+        sa.sql.select([sa.func.max(tas_table.c.tas_id)])
+    ).scalar() or 0
+    conn.execute(
+        tas_table.update().values(tas_id=tas_table.c.tas_id + max_tas_id))
+    op.execute("ALTER SEQUENCE tas_lookup_tas_id_seq RESTART")
+    op.execute("UPDATE tas_lookup SET tas_id = DEFAULT")
```
cmc333333: Yep, `field = DEFAULT` tells postgres to use whatever default exists for that field (in our case, incrementing the sequence). I found it in [this thread](http://stackoverflow.com/a/4678452), but would be happy to do something different if there's a more standard solution. Would `nextval('tas_lookup_tas_id_seq')` fit the bill?

dataactcore/migrations/versions/4dc597500734_add_tas_history_fields.py
```diff
@@ -0,0 +1,86 @@
...
+    conn.execute(
+        tas_table.update().values(internal_start_date=date(2015, 1, 1)))
+    op.alter_column('tas_lookup', 'internal_start_date', nullable=False)
+
+
+def downgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('tas_lookup', 'internal_start_date')
+    op.drop_column('tas_lookup', 'internal_end_date')
+    op.execute("UPDATE tas_lookup SET tas_id = account_num")
```
cmc333333: Good question; _immediately_ prior to this, the sequence was ignored, but prior to _that_, it was being used. What do you think? Maybe we reset it to the max `tas_id` + 1?

dataactcore/models/domainModels.py
cmc333333: That sort of safety net is a good idea; I'll add a constraint.

tests/unit/dataactvalidator/test_scripts_loadTas.py
cmc333333: You're right (and this is something @alburde1 pointed to as well). I'll see if I can make the test flow a bit more easily

tests/unit/dataactvalidator/test_scripts_loadTas.py
cmc333333: Good eye! Will update

cmc333333:
@bsweger @alburde1 updated! I _think_ that big `test_updateTASLookups` is a little easier to read; what do you think?

dataactcore/migrations/versions/4dc597500734_add_tas_history_fields.py
bsweger: Heh, I'm probably the wrong person to ask. Coming from an MSSQL background, I find the Postgres construct of separate sequences a bit odd. From that perspective, `nextval('tas_lookup_tas_id_seq')` would be clearer--I like that idea. Thanks!

dataactcore/migrations/versions/4dc597500734_add_tas_history_fields.py
bsweger: Thanks for thinking this through with me. I like your idea, especially since you've taken such care to craft a bespoke migration for the `upgrade` piece of this. That said, I suspect it's not worth holding up the PR for.

dataactcore/migrations/versions/9dded6e6bf79_add_sanity_check_to_tas.py
```diff
@@ -0,0 +1,40 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.create_unique_constraint('tas_lookup_sanity_check', 'tas_lookup', ['account_num', 'allocation_transfer_agency', 'agency_identifier', 'beginning_period_of_availability', 'ending_period_of_availability', 'availability_type_code', 'main_account_code', 'sub_account_code'])
```
bsweger: nice index name, one of the best I've seen

bsweger:
@cmc333333 Wow, yes--much easier to read imo. Thanks so much for taking the time to tweak that! Let's see what @alburde1 thinks in the AM.

dataactcore/migrations/versions/4dc597500734_add_tas_history_fields.py
cmc333333: Thanks for pointing this out -- turns out downgrading needs to remove the "duplicate" TAS entries and I wouldn't have caught that if not testing for your suggestion.

bsweger:
@cmc333333 migration changes look good-thank you!

PR #384, opened 2016-11-01
Db 1309 redirect to signed s3 url

alburde1:
This PR creates a new route `/sign_submission_file` that takes a submission ID and a file name and creates a signed S3 url (if the user has permissions) and redirects to that url.

Locally, it downloads the file from the appropriate folder.


dataactbroker/handlers/fileHandler.py
```diff
@@ -133,6 +133,28 @@ def getErrorReportURLsForSubmission(self, isWarning = False):
             # Unexpected exception, this is a 500 server error
             return JsonResponse.error(e,StatusCode.INTERNAL_ERROR)
 
+    def getSignedURLForSubmissionFile(self):
```
alburde1: I wasn't sure about function names. I thought maybe leaving function names camelCase and make variable names underscored. We only discussed variable names specifically in the channel discussion so I figured it'd be a good way to differentiate them. Let me know if you disagree

dataactbroker/handlers/fileHandler.py
```diff
@@ -133,6 +133,28 @@ def getErrorReportURLsForSubmission(self, isWarning = False):
...
+        """ Gets the signed URL for the specified file """
+        try:
+            self.s3manager = s3UrlHandler()
```
alburde1: cm and I had this conversation as well and I found that in one of the other functions there's a line that says `self.s3manager = s3UrlHandler(CONFIG_BROKER["aws_bucket"])` which may or may not override the original vanilla declaration so I just set it again just in case to clear it out. I'm not sure if that's ever something that we'd run into but I wanted to make sure in case we did.

nmonga91:
Looks good, merging!

PR #385, opened 2016-11-02
Combine file type tables

bsweger:
We've had a long-standing technical debt issue to combine the `FileType` and `FileTypeValidation` models, which hold pretty much the same information (the two separate models reflect the pre-db-merge days).

This PR folds `FileTypeValidation` into `FileType`. The changes will make the most sense if you step through commit by commit.


bsweger:
Oops, forgot an important note for the reviewer: the two file type tables had the same primary keys assigned for the same file types, which made switching everything out relatively painless.

dataactcore/models/validationModels.py
```diff
@@ -28,8 +28,8 @@ class FileColumn(Base):
     __tablename__ = "file_columns"
 
     file_column_id = Column(Integer, primary_key=True)
-    file_id = Column(Integer, ForeignKey("file_type_validation.file_id"), nullable=True)
-    file = relationship("FileTypeValidation", uselist=False)
+    file_id = Column(Integer, ForeignKey("file_type.file_type_id", name="fk_file_column_file_type"), nullable=True)
```
cmc333333: Curious: it looks like we don't consistently name our FKs (often using the default sqlalchemy generates) -- is this a step towards standardizing on explicit FK names?

dataactcore/interfaces/function_bag.py
```diff
@@ -350,4 +350,23 @@ def getErrorMetricsByJobId(job_id, include_file_types=False, severity_id=None):
...
+    # validations don't apply to all file types (file types
+    # with a NULL file_order--like E and F--are excluded
+    # from the results of the query below)
+    target_files = sess.query(FileType).subquery()
+    cross_file_combos = sess.query(
+        FileType.name.label('first_file_name'),
+        FileType.file_type_id.label('first_file_type_id'),
+        target_files.c.name.label('second_file_name'),
+        target_files.c.file_type_id.label('second_file_type_id')
+    ).filter(FileType.file_order < target_files.c.file_order)
```
cmc333333: Outside the scope of this PR, but I wanted to see what you'd think:
1. Hitting the DB to rely on some implicit NULL handling feels a little icky. Would it make sense to do the logic (including excluding E and F) in Python?
2. I _did_ like that the previous version dealt with pairs of objects (though `source` and `target` weren't great names) as opposed to a sort of joined object with field names prefixed with `first` and `second`. Which do you think is easier to work with?

dataactcore/interfaces/function_bag.py
```diff
@@ -350,4 +350,23 @@ def getErrorMetricsByJobId(job_id, include_file_types=False, severity_id=None):
             record_dict['source_file'] = FILE_TYPE_DICT_ID.get(result.file_type_id, '')
             record_dict['target_file'] = FILE_TYPE_DICT_ID.get(result.target_file_type_id, '')
         result_list.append(record_dict)
-    return result_list
\ No newline at end of file
+    return result_list
+
+def get_cross_file_combos():
+    """Return a list of source and target file types used in cross-file validations."""
```
cmc333333: I don't fully understand this docstring; specifically, the terms `source` and `target` seem kinda arbitrary? Also, can we document the return value type?

dataactcore/models/validationModels.py
bsweger: Good question...we discovered a while back that not putting in an explicit FK name was causing problems in alembic downgrades (I don't remember the specifics, and [this](http://alembic.zzzcomputing.com/en/latest/naming.html) implies that we should be okay without explicit constraint names, given our SQLAlchemy and Alembic versions, so ???)

It could well be something related to the way we're defining indices in the model, too--it looks like there are several ways to do that.

We haven't been consistent in constraint naming. Would be happy to do so going forward, but have also learned from experience that it's hard to make a constraint naming convention stick and not always worth the effort.

dataactcore/interfaces/function_bag.py
bsweger: Good point--will work on this to make the terminology more consistent and understandable.

dataactcore/interfaces/function_bag.py
bsweger: This PR seems like a good enough place to tackle your points, especially since combining the file type tables is the cause of needing to deal with the E and F files here. Can you clarify point 2...by previous version did you mean code that looped through dictionaries? Do you think the entire process would be clearer in straight-up Python w/o involving the database at all?

bsweger: Messing around with this a bit...

```python
from itertools import combinations

from dataactcore.models.lookups import FILE_TYPE_DICT

cross_file_exclude_list = ['sub_award', 'awardee_attributes']
cross_file_dict = {file_name: file_id for file_name, file_id in FILE_TYPE_DICT.items() if file_name not in cross_file_exclude_list}

test = list(map(dict, combinations(cross_file_dict.items(), 2)))
```

creates something to work with:
```
[{'award': 4, 'award_financial': 3},
 {'award': 4, 'program_activity': 2},
 {'appropriations': 1, 'award': 4},
 {'award': 4, 'award_procurement': 5},
 {'award_financial': 3, 'program_activity': 2},
 {'appropriations': 1, 'award_financial': 3},
 {'award_financial': 3, 'award_procurement': 5},
 {'appropriations': 1, 'program_activity': 2},
 {'award_procurement': 5, 'program_activity': 2},
 {'appropriations': 1, 'award_procurement': 5}]
```

Overkill?

bsweger:
@cmc333333 Just pushed some change that I think are an improvement for generating cross-file pairs (it's now more explicit and database-free). It still smells a bit off, though hopefully less so than it used to.

bsweger:
@cmc333333 Just realized--I forgot about your suggestion to drop the physical file_type_validation. If it works for you, I can add that to the next PR, in progress.

cmc333333:
Hey there, sorry to leave this hanging!

dataactcore/utils/report.py
```diff
@@ -1,17 +1,39 @@
...
+    Returns:
+        a list of tuples, where the first tuple represents file #1 in a pair
+        and the second tuple represent file #2 in a pair
+    """
+    # make sure the list is sorted by files' order attributes to ensure that files
+    # in pairs are always listed in the same order
+    crossfile_sorted = sorted([f for f in FILE_TYPE if f.crossfile], key=attrgetter('order'))
+    # create unique combinations of all files eligible for cross-file validation
+    crossfile_combos = itertools.combinations(crossfile_sorted, 2)
+    return list(map(list, crossfile_combos))
```
cmc333333: This seems pretty concise. I wondered if doing two loops like
```python
file_types = [f for f in FILE_TYPE if f.crossfile]
for idx, left in enumerate(file_types):
    for right in file_types[idx+1:]:
        yield (left, right)
```
would be simpler, but it's less explicit. I think "combinations" gives a better indicator.

bsweger: Thanks for this suggestion @cmc333333...good to add these tricks to the toolbox!

PR #386, opened 2016-11-02
Remove unused imports

cmc333333:
As the linter's not part of the CI just yet, the code base has drifted a bit.
Account for unused imports now to make the later linting a bit easier.


nmonga91:
Looks good, merging

PR #387, opened 2016-11-02
Account for historical TAS entries in rule A1 tests

cmc333333:
This builds off #383 ([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1300-tas-history...db-1300-use-tas-history)).

With the TAS history in place, we need to account for validating against inactive TAS entries in the A1 validation rules. This also tries to remove duplicated logic between the three A1 SQL tests (which varied only in the type of object and file name they were testing). This will be easiest to review changeset by changeset.


tests/unit/dataactvalidator/test_a1_appropriations.py
```diff
@@ -54,6 +80,88 @@ def test_failure(database):
...
+    submission = SubmissionFactory(
+        reporting_start_date=date(2014, 1, 1),
+        reporting_end_date=date(2014, 3, 31)
+    )
+
+    assert 1 == number_of_errors(
+        _FILE, database, submission, models=[tas, approp])
+
+
+@pytest.mark.parametrize('begin,end', [
```
bsweger: This is great--thanks for covering all the edge cases here!

PR #389, opened 2016-11-04
Db 1023 submission dashboard

nmonga91:
Merged extra `list_submission*` routes into a single `list_submission` route used for both Recent Submissions and Submission Dashboard. Also reworked the way a submission status is derived and added an additional status if the submission has validated with warnings.

- [x] Reviewed by backend
- [x] @kevinli-work approved


dataactbroker/handlers/accountHandler.py
```diff
@@ -660,52 +660,54 @@ def listUsersWithStatus(self):
...
+        try:
+            limit = int(limit) if limit is not None else 5
+        except:
+            raise ValueError("Incorrect type specified for 'limit'. Please enter a positive number.")
+
+        if certified is not None:
+            certified = certified.lower()
+        # If certified is none, get all submissions without filtering
+        if certified is not None and certified not in ['true', 'false']:
+            raise ValueError("Incorrect value specified for the 'certified' parameter")
```
cmc333333: We talked about this a bit before, but looking at this now, do you think it'd make sense to move this type checking into the route function? It'd keep the HTTP handling/munging in the route and leave this function more pure.

Also, should you convert `certified` to a boolean rather than leaving it as a string?

dataactbroker/handlers/accountHandler.py
```diff
@@ -660,52 +660,54 @@ def listUsersWithStatus(self):
...
+
         user_id = LoginSession.getName(flaskSession)
-        user = self.interfaces.userDb.getUserByUID(user_id)
-        submissions = self.interfaces.jobDb.getSubmissionsByUserAgency(user)
-        submissionDetails = []
+        submissions = self.interfaces.jobDb.getSubmissionsByUserId(user_id, limit=limit, offset=(limit*(page-1)), certified=certified)
```
cmc333333: Do you think this query should be inlined? You're more or less already passing in the filters by parameter anyway.

dataactbroker/handlers/accountHandler.py
```diff
@@ -660,52 +660,54 @@ def listUsersWithStatus(self):
...
+        submission_details = []
+
         for submission in submissions:
             job_ids = self.interfaces.jobDb.getJobsBySubmission(submission.submission_id)
             total_size = 0
             for job_id in job_ids:
                 file_size = self.interfaces.jobDb.getFileSize(job_id)
-                total_size += file_size if file_size is not None else 0
+                total_size += (file_size if file_size is not None else 0)
```
cmc333333: Quick one liner, if it looks simpler to you
```python
total_size += file_size or 0
```

dataactbroker/handlers/accountHandler.py
```diff
@@ -660,52 +660,54 @@ def listUsersWithStatus(self):
...
-                {"submission_id": submission.submission_id, "last_modified": submission.updated_at.strftime('%m/%d/%Y'),
-                 "size": total_size, "status": status, "errors": error_count, "reporting_start_date": str(submission.reporting_start_date),
-                                      "reporting_end_date": str(submission.reporting_end_date), "user": {"user_id": str(user_id),
-                                                                                                    "name": user.name}})
-        return JsonResponse.create(StatusCode.OK, {"submissions": submissionDetails})
+            submission_details.append({"submission_id": submission.submission_id,
+                                       "last_modified": submission.updated_at.strftime('%m/%d/%Y'),
+                                       "size": total_size, "status": status, "errors": submission.number_of_errors,
+                                       "reporting_start_date": str(submission.reporting_start_date),
+                                       "reporting_end_date": str(submission.reporting_end_date),
```
cmc333333: Outside the scope for this PR, but do you know we're using two different date formats here?

dataactbroker/README.md
```diff
@@ -857,6 +857,51 @@ File download or redirect to signed URL
 
 ## File Generation Routes
 
+#### GET "/v1/list_submissions/"
+List submissions for all agencies for which the current user is a member of. Optional query parameters are `?page=[page #]&limit=[limit #]&certified=[true|false]` which correspond to the current page number and how many submissions to return per page (limit). If the query parameters are not present, the default is `page=1`, `limit=5` and if `certified` is not provided, all submissions will be returned containing a mix of the two.
```
cmc333333: This doesn't seem to describe the same logic I'm seeing below. In `getSubmissionsByUserId`, aren't we only getting submissions that the current user has submitted (in contrast to "all agencies for which the current user is a member of")?

dataactbroker/handlers/accountHandler.py
```diff
@@ -660,52 +660,54 @@ def listUsersWithStatus(self):
...
+                                       "user": {"user_id": submission.user_id,
+                                                "name": submission_user_name}})
+
+        total_submissions = GlobalDB.db().session.query(Submission).filter(Submission.user_id == user_id).count()
```
cmc333333: Not at all important, but in case it makes querying easier in the future, you can use `filter_by`:
```python
total_submissions = GlobalDB.db().session.query(Submission).filter_by(user_id=user_id).count()
```

Love seeing the simplicity here, though.

dataactbroker/handlers/fileHandler.py
```diff
@@ -374,7 +374,7 @@ def getStatus(self):
             submissionInfo["reporting_period_end_date"] = self.interfaces.jobDb.getEndDate(submission)
             submissionInfo["created_on"] = self.interfaces.jobDb.getFormattedDatetimeBySubmissionId(submission_id)
             # Include number of errors in submission
-            submissionInfo["number_of_errors"] = sumNumberOfErrorsForJobList(submission_id)
+            submissionInfo["number_of_errors"] = submission.number_of_errors
```
cmc333333: <3

dataactbroker/handlers/jobHandler.py
```diff
@@ -32,9 +32,14 @@ def getSubmissionsByUserAgency(self,user,limit=5):
...
-    def getSubmissionsByUserId(self,userId):
+    def getSubmissionsByUserId(self,userId, limit, offset, certified):
         """ Returns all submissions associated with the specified user ID """
-        return self.session.query(Submission).filter(Submission.user_id == userId).all()
+        if certified is None:
+            return self.session.query(Submission).filter(Submission.user_id == userId).order_by(
+                Submission.updated_at.desc()).limit(limit).offset(offset).all()
+        else:
+            return self.session.query(Submission).filter(and_(Submission.user_id == userId, Submission.publishable == certified)).order_by(
+                Submission.updated_at.desc()).limit(limit).offset(offset).all()
```
cmc333333: Given that these branches are so similar, I think we can build up the query instead:
```python
query = self.session.query(Submission).filter(Submission.user_id == userId).\
    order_by(Submission.updated_at.desc()).limit(limit).offset(offset)
if certified is not None:
    query = query.filter_by(publishable=certified)
return query.all()
```
The benefit here is that we'd only need to adjust logic in one place.

dataactbroker/handlers/jobHandler.py
```diff
@@ -32,9 +32,14 @@ def getSubmissionsByUserAgency(self,user,limit=5):
...
 
-    def getSubmissionsByUserId(self,userId):
+    def getSubmissionsByUserId(self,userId, limit, offset, certified):
         """ Returns all submissions associated with the specified user ID """
-        return self.session.query(Submission).filter(Submission.user_id == userId).all()
+        if certified is None:
+            return self.session.query(Submission).filter(Submission.user_id == userId).order_by(
+                Submission.updated_at.desc()).limit(limit).offset(offset).all()
+        else:
+            return self.session.query(Submission).filter(and_(Submission.user_id == userId, Submission.publishable == certified)).order_by(
```
cmc333333: If I followed the logic correctly,  `certified` is a string here. That might be a problem if sqlalchemy is casting the string (as `bool('false') == True`).

dataactcore/models/jobTrackerInterface.py
```diff
@@ -263,37 +263,45 @@ def setJobRowcounts(self, jobId, numRows, numValidRows):
...
+    def getSubmissionStatus(self,submission):
+        job_ids = self.getJobsBySubmission(submission.submission_id)
         status_names = self.getJobStatusNames()
         statuses = dict(zip(status_names,[0]*len(status_names)))
         skip_count = 0
 
         for job_id in job_ids:
             job = self.getJobById(job_id)
-            if job.job_type.name != "external_validation":
+            if job.job_type.name not in ["external_validation", None]:
```
cmc333333: Curious: why do we need to account for `None` but we didn't before?

dataactcore/models/jobTrackerInterface.py
```diff
@@ -263,37 +263,45 @@ def setJobRowcounts(self, jobId, numRows, numValidRows):
...
                 job_status = job.job_status.name
                 statuses[job_status] += 1
             else:
                 skip_count += 1
```
cmc333333: Significantly outside the scope for this PR, but we've got an n+1 query situation here (we're hitting the db for each job). It'd probably be worth refactoring to do most of this logic in the db.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -263,37 +263,45 @@ def setJobRowcounts(self, jobId, numRows, numValidRows):
...
+            status = "file_errors"
+        elif statuses["running"] != 0:
+            status = "running"
+        elif statuses["waiting"] != 0:
+            status = "waiting"
+        elif statuses["ready"] != 0:
+            status = "ready"
+        elif statuses["finished"] == len(job_ids)-skip_count: # need to account for the jobs that were skipped above
+            status = "validation_successful"
+            if submission.number_of_warnings is not None and submission.number_of_warnings > 0:
```
cmc333333: Outside the scope of this PR, but should we make this column non-nullable (defaulting to zero?)

dataactcore/models/jobTrackerInterface.py
```diff
@@ -263,37 +263,45 @@ def setJobRowcounts(self, jobId, numRows, numValidRows):
...
+                status = "validation_successful_warnings"
+            if submission.publishable:
```
cmc333333: Curious, should this be an `elif`, or do we want to override the previous values?

tests/integration/userTests.py
```diff
@@ -209,9 +209,9 @@ def test_list_submissions(self):
         response = self.app.get("/v1/list_submissions/", headers={"x-session-id": self.session_id})
         self.check_response(response, StatusCode.OK)
         self.assertIn("submissions", response.json)
-        self.assertEqual(len(response.json["submissions"]), 6)
+        self.assertEqual(len(response.json["submissions"]), 5)
```
cmc333333: If I understand correctly, this is `5` due to the new page size; mind adding a comment to explain either way where `5` comes from?

dataactbroker/handlers/accountHandler.py
nmonga91: I originally had it in the route function and I think we should move it back there. That way any subsequent method will assume the correct type and function normally.

I didn't convert `certified` because sqlalchemy doesn't care as long as the string matches True or False. But I can convert if that's best practice.

dataactbroker/handlers/accountHandler.py
nmonga91: Yeah we can. The only thing we'd add is the check if certified is `None` and do a different query based on that.

dataactbroker/handlers/accountHandler.py
nmonga91: I like this so much better!

dataactbroker/handlers/accountHandler.py
nmonga91: you're correct. That should be consistent. nice catch

dataactbroker/README.md
nmonga91: Good catch. That's something that should've been refactored when merging the different routes. I will fix that.

dataactcore/models/jobTrackerInterface.py
nmonga91: There are some jobs that don't have a job_type and their status shouldn't be accounted for since they're not serving a purpose (I can't remember why they're there though)

dataactcore/models/jobTrackerInterface.py
nmonga91: here it should over-ride the previous values

dataactbroker/handlers/jobHandler.py
nmonga91: I don't think it's casting. Looks to be working as intended (based on the results i'm getting back) and I believe we're using string values elsewhere also.

nmonga91:
[run tests]

nmonga91:
[run tests]

tests/unit/dataactbroker/utils.py
```diff
@@ -0,0 +1,17 @@
+def add_models(db, models=None):
+    if models is None:
+        models = []
+
+    # Needs to happen in this order due to foreign key constraints between User and Submission
+    for model in models:
+        db.session.add(model)
+        db.session.commit()
```
cmc333333: This feels a little kludgey; can we do something like
```python
submission.user = user
session.add(submission)
session.commit()
```
and expect the user to be added transitively?

If not, maybe we should rename this function to something like `add_models_in_order` to indicate how it's different.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -99,4 +103,115 @@ def test_max_login_failure(monkeypatch):
                     "Please contact DATABroker@fiscal.treasury.gov to obtain access."
 
     # Not in cgac group
-    assert error_message == json.loads(json_response.get_data().decode("utf-8"))['message']
\ No newline at end of file
+    assert error_message == json.loads(json_response.get_data().decode("utf-8"))['message']
+
+
+def test_list_submissions_success(database, job_constants, monkeypatch):
```
cmc333333: Would it make sense to split this into multiple tests? Then each test could be named to describe the situation it's testing a bit more accurately and you wouldn't need to clean up after every set of assertions.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -99,4 +103,115 @@ def test_max_login_failure(monkeypatch):
...
+    assert json.loads(json_response.get_data().decode("utf-8"))['total'] == 1
+    assert json.loads(json_response.get_data().decode("utf-8"))['submissions'][0]['status'] == "validation_successful_warnings"
+    delete_models(database, [user, sub])
+
+    sess = database.session
+    user = UserFactory(user_id=1, cgac_code='cgac')
+    sub = SubmissionFactory(user_id=1, submission_id=1, cgac_code='cgac')
+    job = JobFactory(submission_id=1, job_status=sess.query(JobStatus).filter_by(name='finished').one(),
+                     job_type=sess.query(JobType).filter_by(name='csv_record_validation').one(),
+                     file_type=sess.query(FileType).filter_by(name='award').one())
```
cmc333333: Can we use the dictionaries in `dataactcore.models.lookups` rather than querying the db? Something like:
```python
job = JobFactory(submission_id=1, job_status_id=JOB_STATUS_DICT['finished'], ...
```

Also, are the `job_type` and `file_type` relevant to this test? If not, we can just use the randomly-generated values to make this test easier to read.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -99,4 +103,115 @@ def test_max_login_failure(monkeypatch):
...
+    sub = SubmissionFactory(user_id=1, submission_id=1, cgac_code='cgac')
+    job = JobFactory(submission_id=1, job_status=sess.query(JobStatus).filter_by(name='ready').one(),
+                     job_type=sess.query(JobType).filter_by(name='csv_record_validation').one(),
+                     file_type=sess.query(FileType).filter_by(name='award').one())
+    add_models(database, [user, sub, job])
+
+    json_response = ah.list_submissions(None, None, None)
+    assert json.loads(json_response.get_data().decode("utf-8"))['total'] == 1
+    assert json.loads(json_response.get_data().decode("utf-8"))['submissions'][0]['status'] == "ready"
+    delete_models(database, [user, sub, job])
```
cmc333333: I'm not sure these last three sets of asserts are needed -- they're all just verifying that the submission's status is being used. I think we should probably replace them with tests of the other logic in the `list_submission` function, like paging, `certified`, and user name.

tests/unit/dataactbroker/test_account_handler.py
```diff
@@ -99,4 +103,115 @@ def test_max_login_failure(monkeypatch):
...
+    sub = SubmissionFactory(user_id=1, submission_id=1, cgac_code='cgac')
+    job = JobFactory(submission_id=1, job_status=sess.query(JobStatus).filter_by(name='invalid').one(),
+                     job_type=sess.query(JobType).filter_by(name='csv_record_validation').one(),
+                     file_type=sess.query(FileType).filter_by(name='award').one())
+    add_models(database, [user, sub, job])
+
+    json_response = ah.list_submissions(None, None, None)
+    assert json.loads(json_response.get_data().decode("utf-8"))['total'] == 1
+    assert json.loads(json_response.get_data().decode("utf-8"))['submissions'][0]['status'] == "file_errors"
+    delete_models(database, [user, sub, job])
```
cmc333333: Similarly, it doesn't seem like these tests are testing something new.

dataactbroker/fileRoutes.py
```diff
@@ -65,9 +65,28 @@ def list_submissions():
...
+        except:
+            raise ValueError("Incorrect type specified for 'limit'. Please enter a positive number.")
+
+        if certified is not None:
+            certified = certified.lower()
+        else:
+            raise ValueError("Missing required parameter 'certified'")
+        # If certified is none, get all submissions without filtering
+        if certified is not None and certified not in ['mixed', 'true', 'false']:
+            raise ValueError("Incorrect value specified for the 'certified' parameter")
```
cmc333333: I'm all for moving this user-error handling here, but I think that means we need to handle the `ValueError`s differently. `RouteUtils.run_instance_function` is where that crazy try-catch logic is currently happening. Maybe this is a good time to add a [flask-style exception handler](http://flask.pocoo.org/docs/0.11/patterns/apierrors/)?

klrBAH:
[run tests]

tests/unit/dataactbroker/utils.py
nmonga91: I like the transitive approach. I will refactor to use that.

klrBAH:
[run tests]

klrBAH:
[run tests]

klrBAH:
[run tests]

tests/unit/dataactbroker/test_account_handler.py
nmonga91: I can break it down into multiple tests and mimic `error_rows` and have something similar that adds and cleans up in one go. But the clean up would still need to happen explicitly.

tests/unit/dataactbroker/test_account_handler.py
nmonga91: I added those to specifically test that the correct status comes back based on what the related jobs' statuses were. I wanted to check every possible status to make sure we get back what's expected when a job has a specific status.

dataactbroker/fileRoutes.py
nmonga91: this has been updated!

cmc333333:
[run tests]

dataactbroker/exception_handler.py
```diff
@@ -0,0 +1,11 @@
+from dataactbroker.exceptions.invalid_usage import InvalidUsage
+from flask import jsonify
+
+
+def add_exception_handlers(app):
+
+    @app.errorhandler(InvalidUsage)
+    def handle_invalid_usage(error):
+        response = jsonify(error.to_dict())
```
cmc333333: Very nice!

cmc333333:
Looks good to me and certainly enough progress to be worth merging. Quick note that there are branch conflicts.

cmc333333:
A lot of good work here. I'm particularly excited about the error handler!

PR #393, opened 2016-11-04
More validation interface refactor

bsweger:
This is the last bit of the validationInterface removal. I would recommend going through commit by commit, but it probably doesn't matter too much, since the goal wasn't to add anything but rather to take things away.


dataactcore/models/lookups.py
```diff
@@ -58,13 +61,13 @@
...
-    LookupFileType(5, 'award_procurement', '', 'D1', 5),
-    LookupFileType(6, 'awardee_attributes', '', 'E', None),
-    LookupFileType(7, 'sub_award', '', 'F', None)
+    LookupFileType(1, 'appropriations', '', 'A', 1, Appropriation),
+    LookupFileType(2, 'program_activity', '', 'B', 2, ObjectClassProgramActivity),
+    LookupFileType(3, 'award_financial', '', 'C', 3, AwardFinancial),
+    LookupFileType(4, 'award', '', 'D2', 4, AwardFinancialAssistance),
+    LookupFileType(5, 'award_procurement', '', 'D1', 5, AwardProcurement),
+    LookupFileType(6, 'awardee_attributes', '', 'E', None, None),
+    LookupFileType(7, 'sub_award', '', 'F', None, None)
```
cmc333333: I like this. The only thing to watch out for is circular dependencies -- so long as we always point lookups to models (and never the reverse) we're fine.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -169,29 +169,30 @@ def readRecord(self,reader,writer,file_type,interfaces,row_number,job_id,fields,
...
         Args:
             record: Record to be written
             job_id: ID of current job
             submission_id: ID of current submission
             passed_validations: True if record has not failed first validations
-            interfaces: InterfaceHolder object
             writer: CsvWriter object
             row_number: Current row number
-            file_type: Type of file for current job
+            model: orm model for the current file
```
cmc333333: Good call wrt passing in the most specific thing the function needs. We can always re-generalize later.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -267,8 +268,12 @@ def runValidation(self, job_id, interfaces):
 
         rowNumber = 1
         fileType = jobTracker.getFileType(job_id)
+        # Get orm model for this file
+        model = [ft.model for ft in FILE_TYPE if ft.name == fileType][0]
```
cmc333333: I'm not sure what I think of this syntax, but I've seen some folks use it before:

```python
model = next(ft.model for ft in FILE_TYPE if ft.name == fileType)
```

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -46,6 +46,12 @@ def __init__(self,isLocal =True,directory=""):
         self.isLocal = isLocal
         self.directory = directory
 
+        # create long-to-short (and vice-versa) column name mappings
+        sess = GlobalDB.db().session
+        colnames = sess.query(FileColumn.name, FileColumn.name_short).all()
+        self.long_to_short_dict = {row.name: row.name_short for row in colnames}
+        self.short_to_long_dict = {row.name_short: row.name for row in colnames}
```
cmc333333: Solid! Nice and contained.

dataactcore/migrations/versions/a97dabbd44f4_drop_filetype_validation_table.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for ft in FILE_TYPE:
+        fileType = FileTypeValidation(
+            file_id=ft.id,
+            name=ft.name,
+            description=ft.desc,
+            file_order=ft.order
+        )
+        session.merge(fileType)
+    session.commit()
```
cmc333333: <3 for remembering the backwards data migration.

dataactcore/migrations/versions/a97dabbd44f4_drop_filetype_validation_table.py
```diff
@@ -21,14 +21,6 @@
...
 Session = sessionmaker()
 
-class FileTypeValidation(Base):
-    __tablename__ = 'file_type_validation'
-
-    file_id = sa.Column(sa.Integer, primary_key=True)
-    name = sa.Column(sa.Text)
-    description = sa.Column(sa.Text)
-    file_order = sa.Column(sa.Integer, nullable=False, server_default="0")
-
```
cmc333333: What happens if we _do_ declare this class outside of the downgrade? If there's an explosion, we may need to update some other migrations 😅

PR #394, opened 2016-11-07
Prep for logging refactor

cmc333333:
A step towards replacing CloudLogger

* Remove no-longer-used logstash configurations
* Consolidate logging configuration to a single function
* Add default logging config values (we'll tweak these in later commits) 

Next up:
1. replace CloudLogger with roughly equivalent Python `logging` configurations
2. leverage Python `logging` to provide more structured and context-sensitive log messages


nmonga91:
Do we still want to call `CloudLogger` "CloudLogger"? Since the logging methodology is changing, does it make sense to also change the name of the class?

cmc333333:
The plan is to remove CloudLogger entirely in favor of Python's `logging` facilities

cmc333333:
Closing as all of these changes were part of #397

PR #396, opened 2016-11-08
Use new, consolidated debug flag when creating Flask apps

bsweger:
This commit uses the new debug flag for all debug-related settings.
Going forward, code that needs to know the debug status of the app
can either get it from app.debug, if the Flask app is handy, or
from CONFIG_SERVICES['debug'].

The new flag has already been merged into the config repo: https://github.com/fedspendingtransparency/data-act-broker-config/pull/25

Folks will have to add debug to their local configs once this is merged.

### DIFF
--- a/dataactbroker/app.py
+++ b/dataactbroker/app.py
@@ -20,8 +20,7 @@ def createApp():
     local = CONFIG_BROKER['local']
     app.config.from_object(__name__)
     app.config['LOCAL'] = local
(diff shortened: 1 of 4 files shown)


klrBAH:
[run tests]

cmc333333:
This makes sense to me and should allow a single config to support both current prod and dev versions of the codebase.

PR #397, opened 2016-11-08
Down with CloudLogger (logging refactor part 2)

cmc333333:
Builds on #394([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1110-logging-cleanup-1...db-1110-logging-cleanup-2)) to replace `CloudLogger` with `logging`-based analogs. It's a big one, particularly the `JsonResponse` patch, but it's easiest to review changeset by changeset. Luckily, the logic isn't dramatically different.

* Creates several "deprecated" loggers and configures them to write to the same files we used `CloudLogger` for
* Switches over all log calls to use one of these loggers
* Switches to raising exceptions rather than creating them and passing them around. This'll give us access to the exception-handing context (which, in turn, gives us backtraces, etc. for free). I think that, ideally, we wouldn't need to try-catch these exceptions all over the place, that they'd percolate to the application handler.


nmonga91:
Looks good, merging!

dataactcore/utils/jsonResponse.py
```diff
@@ -41,10 +47,10 @@ def error(exception, errorCode, **kwargs):
             responseDict[key] = kwargs[key]
 
 
-        exc_type, exc_obj, exc_tb = sys.exc_info()
+        _, _, exc_tb = sys.exc_info()
```
bsweger: Not sure if this is better or not, but I think an alternate way to extract the traceback from an exception in Python 3 is `exception.__traceback__`

dataactvalidator/filestreaming/schemaLoader.py
```diff
@@ -141,4 +141,5 @@ def loadAllFromPath(cls,path):
             cls.loadFields(key,filepath)
 
 if __name__ == '__main__':
+    configure_logging()
```
cmc333333: The general idea is to call `configure_logging` at the entry points to the code base rather than having them set automatically on import. The benefit here is largely around testing (we had been writing all of the logs to disk, effectively silently ignoring them), which often has different logging requirements than normal application runs. More broadly, though, doing configuration at entry points allows the code to be imported/reused without setting global state for everyone else.

dataactvalidator/scripts/loadFile.py
```diff
@@ -129,6 +129,7 @@ def loadDomainValues(basePath, localProgramActivity = None):
 
 
 if __name__ == '__main__':
+    configure_logging()
```
cmc333333: This might help explain: If you were in a `python` session and called `loadFile` directly, you'd get the default logging settings of your environment (usually, writing INFO/WARNING/ERROR to the console), regardless of what files we'd "normally" write to.

dataactcore/models/jobTrackerInterface.py
```diff
@@ -193,9 +192,9 @@ def checkJobDependencies(self,jobId):
         for depJobId in self.getDependentJobs(jobId):
             isReady = True
             if not (self.getJobStatus(depJobId) == self.getJobStatusId('waiting')):
-                CloudLogger.logError("Job dependency is not in a 'waiting' state",
-                                     ResponseException("Job dependency is not in a 'waiting' state",StatusCode.CLIENT_ERROR, ValueError),
-                                     traceback.extract_stack())
+                _exception_logger.error(
```
cmc333333: Somewhat related -- I noticed that a lot of these log messages don't contain too much context. They're just "The state is wrong, line 55". I'm _hoping_ that we can set user and submission ids on `g` and then retrieve them in the logger to tack that information on, if it's available.

dataactvalidator/filestreaming/schemaLoader.py
bsweger: Aaaaaah--I see now. Thanks for explaining!

dataactvalidator/validation_handlers/validator.py
```diff
@@ -188,7 +194,9 @@ def validateFileBySql(cls, submissionId, fileType, short_to_long_dict):
 
         # For each rule, execute sql for rule
         for rule in rules:
-            CloudLogger.logError("VALIDATOR_INFO: ", "Running query: "+str(RuleSql.query_name)+" on submissionID: " + str(submissionId) + " fileType: "+ fileType, "")
+            _exception_logger.info(
+                'VALIDATOR_INFO: Running query: %s on submissionId %s, '
+                'fileType: %s', rule.query_name, submissionId, fileType)
```
cmc333333: Hah. Good on you for a thorough review!

dataactcore/models/jobTrackerInterface.py
bsweger: Oh, that sounds like a good plan

dataactbroker/handlers/accountHandler.py
```diff
@@ -494,42 +529,46 @@ def update_user(self, system_email):
...
+                    "following: status, permissions, is_active",
+                    StatusCode.CLIENT_ERROR
+                )
             # Find user that matches specified uid
-            user = sess.query(User).filter(User.user_id == int(request_dict.getValue("uid"))).one()
-        except NoResultFound:
-            exc = ResponseException("No users with that uid", StatusCode.CLIENT_ERROR)
+            user = sess.query(User).filter_by(
+                user_id=int(request_dict['uid'])).one_or_none()
+            if user is None:
```
cmc333333: I don't think it's always the right answer; in these cases, I more wanted to avoid nested `try`s :)

dataactbroker/handlers/accountHandler.py
```diff
@@ -830,45 +886,61 @@ def set_skip_guide(self, session):
...
             return JsonResponse.error(exc, exc.status)
 
         current_user = sess.query(User).filter(User.user_id == session["name"]).one()
 
-        user_ids = request_dict.getValue("users")
-        submission_id = request_dict.getValue("submission_id")
+        user_ids = request_dict['users']
+        submission_id = request_dict['submission_id']
         # Check if submission id is valid
         self.jobManager.getSubmissionById(submission_id)
```
cmc333333: I don't think the logging changes affect that decision. This PR does all of the `try-catch`s to mimic what was happening before (where we were explicitly returning an error `JsonResponse`). Given that `getSubmissionById` is currently raising `ResponseException`, you'll need to decide whether it makes sense to continue to raise that, handle the Sqlalchemy exceptions here (or elsewhere), or use some other strategy. The bit that ties into logging is that `JsonResponse.error` needs to be called within an exception-handing context, if you're planning to use that.

dataactcore/utils/jsonResponse.py
cmc333333: Good to know. I'll try it out; the `exc_info` stuff is the definition of an incantation.

dataactbroker/handlers/accountHandler.py
bsweger: Ok...will give this some more thought when the time comes. Thanks for weighing in!

PR #399, opened 2016-11-08
Swap out 0000-00-00 for 0001-01-01 in SOAP calls

cmc333333:
Both are nonsensical, but the latter can be parsed.


alburde1:
Looks good, merging!

PR #402, opened 2016-11-10
Updating the submission and job models

alburde1:
This pull request upgrades the number_of_errors and number_of_warnings fields in the submission and job models to default to 0 to prevent errors.


dataactcore/migrations/versions/bb33cc8f0a3e_setting_submission_error_numbers_to_.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+
+
+
+def upgrade_data_broker():
+    op.execute('UPDATE job SET number_of_errors = 0 WHERE number_of_errors IS NULL')
+    op.alter_column('job', 'number_of_errors',
+               existing_type=sa.INTEGER(),
+               nullable=False)
```
cmc333333: Hm. This doesn't appear to be including the `server_default`; could be that alembic missed it. Would you mind verifying that your local DB has this default set? `\d+t job`

alburde1: Weird that alembic missed it, fixed though

cmc333333:
Nice work! Migrations aren't so bad :)

PR #403, opened 2016-11-10
Add docs around logging config

cmc333333:
Initial docs describing how to modify the log settings. Anything else
interesting/questions that should be answered here?


bsweger:
@cmc333333 Thanks for adding this! Maybe I'm just being dense, but do you think it would be helpful to have (or link to) a snippet of code that devs can use "out of the box" if they want to explicitly log something when coding. I get that we don't want to recreate the Python docs, but I don't find them especially straightforward on logging.

PR #404, opened 2016-11-10
Global validation error handling

bsweger:
I went through a lot of iterations on this and finally came back to @cmc333333's advice of keeping
things simple for the first go-round.

This PR tries to reduce the layers of try/except that wrap the validation application and process. Ultimately, I needed to untangle these layers to best proceed with the JobTracker database refactor.

Best to step through the commits--I tried to explain everything in the messages.

Once this is reviewed and merged, the next step is to remove the second try/except block in `validation_job`

### DIFF
--- a/dataactvalidator/validation_handlers/validationManager.py
+++ b/dataactvalidator/validation_handlers/validationManager.py
@@ -524,49 +524,49 @@ class ValidationManager:
         Returns:
         Http response object
         """
-        sess = GlobalDB.db().session
         # Create connection to job tracker database
         self.filename = None
-        job_id = None
-        jobTracker = None
+        sess = GlobalDB.db().session
 
-        try:
-            jobTracker = interfaces.jobDb
-            requestDict = RequestDictionary(request)
-            if requestDict.exists("job_id"):
-                job_id = requestDict.getValue("job_id")
-            else:
-                # Request does not have a job ID, can't validate
-                raise ResponseException("No job ID specified in request",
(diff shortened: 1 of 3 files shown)


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -524,49 +524,50 @@ def validate_job(self, request, interfaces):
...
+        jobTracker = interfaces.jobDb
+        requestDict = RequestDictionary(request)
+        if requestDict.exists('job_id'):
+            job_id = requestDict.getValue('job_id')
+        else:
+            # Request does not have a job ID, can't validate
+            validation_error_type = ValidationError.jobError
+            raise ResponseException('No job ID specified in request',
+                                    StatusCode.CLIENT_ERROR, None,
+                                    validation_error_type)
```
alburde1: Why are we adding more variables to all the `raise ResponseException` lines? Is this required everywhere now and will it break what we already have if we don't?

bsweger: Good question! The last value here (`validation_error_type`) was being added in the old code, down on line 562 (`e.errorType = ValidationError.jobError`). So this new code definitely has some repetition, but overall I'm hoping it's more explicit and easier to read than relying on a `try/except` to wrap the raising of `ResponseException`.

The `None` that precedes it is also just being more explicit.

To the question of "is this required everyone now"...`ResponseException` itself hasn't been changed, so it should continue to work the same way.

PR #405, opened 2016-11-14
Add tas_id FK to staging tables; use it in validations

cmc333333:
We have been using a concatenated `tas` field to compare objects across staging tables. We have a history of valid TAS values, however, so it would be better to replace that field with a foreign key to the TAS table. This takes a step in that direction by adding (and deriving) the appropriate value and using it in validations. We continue to use the `tas` field when joining against SF_133, though this will go away in a later PR.

This looks like a lot of changes, but in terms of new functionality, it's actually pretty small. Best to review changeset by changeset. Please particularly review the A32 change to check me.


dataactvalidator/config/sqlrules/a32_appropriations.sql
```diff
@@ -9,10 +9,9 @@ SELECT approp.row_number,
 FROM appropriation AS approp
 WHERE approp.submission_id = {0}
 	AND EXISTS (
-		SELECT other.row_number,
-			other.tas
+		SELECT 1
 		FROM appropriation AS other
 		WHERE other.row_number <> approp.row_number
-			AND other.tas IS NOT DISTINCT FROM approp.tas
+			AND other.tas_id IS NOT DISTINCT FROM approp.tas_id
```
bsweger: What about changing this line to `AND other.tas_id = approp.tas_id`? I think this is safe to do since previous rules are already generating errors for rows with a null `tas_id`. If I'm reading this correctly, keeping the `IS DISTINCT FROM` has potential to generate a lot of noise if there are multiple rows in a file with NULL `tas_id`.

Historically, there's been a lot of back and forth about the question of: do we show all error messages for all rules, even if multiple rule failures stem from the same data issue. Regardless of the differing opinions on that question, I don't see anything to be gained by potentially allowing false positives.

tests/unit/dataactvalidator/test_b20_object_class_program_activity.py
```diff
@@ -17,28 +17,42 @@ def test_column_headers(database):
 
 def test_success(database):
     """ Tests that all combinations of TAS, program activity code, and object class in File C exist in File B """
-    tas = "".join([_TAS, "_success"])
-
-    op = ObjectClassProgramActivity(job_id=1, row_number=1, tas=tas, program_activity_code='1', object_class='1')
```
bsweger: Thanks for removing the extraneous bits here and leaving what's necessary--easier to grok this way.

dataactcore/models/stagingModels.py
```diff
@@ -53,6 +36,7 @@ class Appropriation(Base):
     tas = Column(Text, index=True, nullable=False, default=concatTas, onupdate=concatTas)
     tas_id = Column(Integer, ForeignKey("tas_lookup.tas_id", name='fk_tas'),
                     nullable=True)
+    tas_obj = relationship(TASLookup)
```
bsweger: Good call.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -600,3 +606,48 @@ def validate_job(self, request, interfaces):
...
+    sess = GlobalDB.db().session
+    submission = sess.query(Submission).\
+        filter_by(submission_id=submission_id).one()
+
+    # Due to the OVERLAPS, tuples, and IS NOT DISTINCT FROM, this query is
+    # more gnarly in sqlalchemy than straight SQL, so using SQL instead
+    sql = """
+        UPDATE {table_name}
+        SET tas_id = (
+            SELECT min(tas.tas_id)
```
cmc333333: From what I understand, it's technically possible, but shouldn't come up in practice. I'll document in a note to see what you think.

dataactvalidator/config/sqlrules/a32_appropriations.sql
cmc333333: Can do and understood about the noise. There's the potential that we wouldn't show an A32 error if A1 was also in error, but that seems better than showing an A32 error _whenever_ A1 errors.

dataactcore/models/stagingModels.py
cmc333333: Turns out, this also resolves a runtime dependency issue where sqlalchemy didn't know about the `tas_lookup` table when it processed the staging models. Flipping the imports and adding this relation ensured that they'd be processed in the right order at runtime, too.

bsweger:
Thanks for the updates, @cmc333333!

PR #406, opened 2016-11-14
Db 1289 max permissions

nmonga91:
Converting legacy permissions handling (using bits for multiple roles) to MAX permission handling (using an ID with cascading permissions).

- [x] Reviewed by backend
- [x] MAX groups created and contain the necessary individuals
- [x] Merge coordinated with https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/349


dataactbroker/handlers/accountHandler.py
```diff
@@ -582,11 +583,11 @@ def update_user(self, system_email):
         if 'is_active' in request_dict:
             is_active = bool(request_dict['is_active'])
             if not user.is_active and is_active:
-                # Reset password count to 0
-                self.reset_password_count(user)
+                if is_local:
+                    # Reset password count to 0
+                    self.reset_password_count(user)
```
cmc333333: Is this a good time to remove `reset_password_count` altogether?

dataactbroker/handlers/accountHandler.py
```diff
@@ -166,12 +167,7 @@ def max_login(self,session):
             group_list_all = max_dict['cas:serviceResponse']['cas:authenticationSuccess']['cas:attributes']['maxAttribute:GroupList'].split(',')
             group_list = [g for g in group_list_all if g.startswith(parent_group)]
 
-            # Deny access if not in the parent group aka they're not allowed to access the website all together
-            if not parent_group in group_list:
-                raise ValueError("You have logged in with MAX but do not have permission to access the broker. Please "
-                                 "contact DATABroker@fiscal.treasury.gov to obtain access.")
-
-            cgac_group = [g for g in group_list if len(g) == len(parent_group + "-CGAC_")+3]
+            cgac_group = [g for g in group_list if g.startswith(parent_group+"-CGAC_")]
```
cmc333333: Much easier to read, nice.

dataactbroker/handlers/accountHandler.py
```diff
@@ -302,17 +309,11 @@ def ThreadedFunction (username="", title="", cgac_code="", user_email="" , link=
                 filter(CGAC.cgac_code == cgac_code).\
                 one_or_none()
             agency_name = "Unknown" if agency_name is None else agency_name
-            for user in getUsersByType("website_admin"):
+            for user in sess.query(User).filter_by(permission_type_id=PERMISSION_TYPE_DICT['website_admin']).all():
```
cmc333333: NBD, but you don't need to use all() here -- if you iterate over a query, you'll let sqlalchemy handle the paging and won't need to instantiate the whole list in memory.

dataactbroker/permissions.py
```diff
@@ -34,8 +34,9 @@ def decorated_function(*args, **kwargs):
                 elif LoginSession.isLogin(session):
                     user = sess.query(User).filter(User.user_id == session["name"]).one()
                     valid_user = True
-                    for permission in permission_list :
-                        if not has_permission(user, permission):
+                    for permission in permission_list:
+                        if permission in PERMISSION_TYPE_DICT and \
+                                not user.permission_type_id == PERMISSION_TYPE_DICT[permission]:
```
cmc333333: I think we probably want to error pretty loudly if the permission specified isn't in the DICT. Maybe we do something like:
```python
def permissions_check(f=None,permission_list=[]):
    for permission in permission_list:
        if permission not in PERMISSION_TYPE_DICT:
            raise ValueError("{} not a valid permission".format(permission))
```
That'd error out at application startup (and during tests) rather than silently failing at run time.

Also, this block confuses me a bit -- if we require that a user have all of the permissions in the `permission_list`, but a user can only have one permission, why have a list?

dataactbroker/handlers/accountHandler.py
nmonga91: Yeah let's get rid of it. I already removed the reset_password_email call. If you're running locally you'll be using the admin account from the config and we don't wanna accidentally log that out. Should we also get rid of the 120 days check to see if a user should be marked inactive?

dataactbroker/handlers/accountHandler.py
```diff
@@ -575,12 +557,19 @@ def list_users(self):
...
-                        "permissions": ",".join(self.interfaces.userDb.getUserPermissions(user)), "status": user.user_status.name}
+                        "permission": self.get_permission_name(user.permission_type_id), "status": user.user_status.name}
             user_info.append(thisInfo)
         return JsonResponse.create(StatusCode.OK,{"users":user_info})
 
+    def get_permission_name(self, perm_id):
+        for permission_name, permission_id in PERMISSION_TYPE_DICT.items():
+            if permission_id == perm_id:
+                return permission_name
+        return None
```
nmonga91: Yes! I'm all for that and will incorporate

dataactbroker/handlers/accountHandler.py
cmc333333: Makes sense to me

dataactbroker/permissions.py
nmonga91: I've added the louder exception handling and have removed the use of a list. You're right, we're not longer using a list for checking permissions anywhere (double checked since we're using this for checking email token and password token, which will be removed soon)

dataactbroker/permissions.py
```diff
@@ -11,12 +11,16 @@
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.models.userModel import User
 from dataactbroker.exceptions.invalid_usage import InvalidUsage
-from dataactcore.models.lookups import PERMISSION_TYPE_DICT
+from dataactcore.models.lookups import PERMISSION_TYPE_DICT, PERMISSION_MAP, PERMISSION_TYPE_DICT_ID
+
+temp_perm_list = ["check_email_token", "check_password_token"]
```
cmc333333: Mind adding a comment explaining what these are used for?

dataactbroker/permissions.py
```diff
@@ -38,12 +42,20 @@ def decorated_function(*args, **kwargs):
                 elif LoginSession.isLogin(session):
                     user = sess.query(User).filter(User.user_id == session["name"]).one()
                     valid_user = True
-                    if permission in PERMISSION_TYPE_DICT and \
-                            not user.permission_type_id == PERMISSION_TYPE_DICT[permission]:
-                        valid_user = False
+
+                    if permission is not None and not user.permission_type_id == PERMISSION_TYPE_DICT['website_admin']:
```
cmc333333: Isn't `website_admin` the top permission? If so, why does it need to be excluded?

dataactbroker/permissions.py
```diff
@@ -38,12 +42,20 @@ def decorated_function(*args, **kwargs):
...
+                        perm_hierarchy = {d['name']: d['order'] for d in PERMISSION_MAP.values()}
+                        # if the users permission is not higher than the one specified, check their permission
+                        # if user's perm order is < than what's passed in, it means they have higher permissions
+                        if perm_hierarchy[PERMISSION_TYPE_DICT_ID[user.permission_type_id]] > perm_hierarchy[permission]:
```
cmc333333: Should this be
```python
if perm_hierarchy[PERM...] < perm_hierarchy[perm...]:
    valid_user = False
```

dataactcore/models/lookups.py
```diff
@@ -90,6 +90,8 @@
 ]
 PERMISSION_TYPE_DICT = {item.name: item.id for item in PERMISSION_TYPE}
 PERMISSION_TYPE_DICT_ID = {item.id: item.name for item in PERMISSION_TYPE}
+PERMISSION_MAP = {'r': {'name': 'reader', 'order': 3}, 'w': {'name': 'writer', 'order': 2},
+                  's': {'name': 'submitter', 'order': 1}}
```
cmc333333: Good call

dataactcore/models/userModel.py
```diff
@@ -13,7 +13,8 @@ class User(Base):
     name = Column(Text)
     cgac_code = Column(Text)
     title = Column(Text)
-    permissions = Column(Integer)
+    permission_type_id = Column(Integer, ForeignKey(column="permission_type.permission_type_id",
+                                                    name="user_permission_type_fk"))
```
cmc333333: I think you'll need to modify the name in the migration, too.

dataactcore/migrations/versions/812387580a0b_rename_user_permissions_column.py
```diff
@@ -0,0 +1,44 @@
...
+    op.add_column('users', sa.Column('permission_type_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(None, 'users', 'permission_type', ['permission_type_id'], ['permission_type_id'])
+    op.drop_column('users', 'permissions')
+    ### end Alembic commands ###
+
+
+def downgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('permissions', sa.INTEGER(), autoincrement=False, nullable=True))
+    op.drop_constraint(None, 'users', type_='foreignkey')
```
cmc333333: At line 41, the user table has both columns, so that's where we'd migrate the data. As you say, we wouldn't want to use the `User` model directly (though we could replicate it) as it wouldn't have the right fields. Here's one example: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/dataactcore/migrations/versions/a97dabbd44f4_drop_filetype_validation_table.py#L51 

In any event, I think there are more important issues at hand. As long as we're okay with the remote possibility of permission-less users, let's not worry about it.

dataactcore/models/userModel.py
nmonga91: added

dataactbroker/permissions.py
```diff
@@ -38,12 +42,20 @@ def decorated_function(*args, **kwargs):
...
+                            print(perm_hierarchy[PERMISSION_TYPE_DICT_ID[user.permission_type_id]])
+                            print(perm_hierarchy[permission])
```
nmonga91: removed

dataactbroker/permissions.py
nmonga91: Higher permissions have a lower `order` value so that when it's ordered it's a higher priority. that's why i had the comparison as it is above.

dataactbroker/permissions.py
nmonga91: I've fixed this handling

nmonga91:
[run tests]

cmc333333:
Looks good to me. Ready to merge whenever.

cmc333333:
Nice work, @nmonga91

PR #408, opened 2016-11-14
Exclude migrations from flake8

cmc333333:
These are auto-generated files, so we don't really need to lint them. Also fixes a linting bug. We should be able to turn on linting on Jenkins after this.


bsweger:
Thanks for staying on top of this, @cmc333333!

PR #409, opened 2016-11-15
Checking for empty permissions when listing users on the admin page

nmonga91:


dataactbroker/handlers/accountHandler.py
```diff
@@ -606,9 +606,10 @@ def list_users(self):
                 filter(CGAC.cgac_code == user.cgac_code).\
                 one_or_none()
 
+            permission = None if user.permission_type_id is None else PERMISSION_TYPE_DICT_ID[user.permission_type_id]
```
cmc333333: Shouldn't hold this back, but this is a little more concise (it'll set permission to `None` if it can't be found in the dict)
```python
permission = PERMISSION_TYPE_DICT_ID.get(user.permission_type_id)
```

cmc333333:
:+1:

PR #411, opened 2016-11-16
Finish global validation error handling and related job interface moves

bsweger:
This PR does two major things:

* Removes the 2nd big `try/except` block around the validation process. Instead, any errors raised during validation are caught by exception handlers at the app level.
* Move/removes related functions from the jobTrackerInterface (_e.g._ functions and their dependencies that are used by the aforementioned exception handlers).

The PR also addresses a few minor things that cropped up along the way:
* POST to the validate route wasn't being sent a JSON (problematic b/c our integration tests were sending as JSON, and the difference resulted in some errors not being caught in tests)
* FIx intermittent Flask `AssertionErrors` in integration tests

Left to do on the job "db" refactor:
* remove remaining functions from `jobHandler.py` and `jobTrackerInterface.py` + cleanup
* tidy up the exception handlers in the validator's `app.py`, they contain a lot of duplicate code
* fix the enqueue import hack


dataactbroker/handlers/fileHandler.py
```diff
@@ -397,14 +398,15 @@ def getStatus(self):
             submissionInfo["last_updated"] = submission.updated_at.strftime("%Y-%m-%dT%H:%M:%S")
 
             for job_id in jobs:
+                job = sess.query(Job).filter_by(job_id = job_id).one()
```
cmc333333: Small detail, but would it make more sense to grab all the jobs in one go around line 385? Otherwise we're hitting the db for each job unnecessarily.

dataactcore/interfaces/function_bag.py
```diff
@@ -503,3 +507,66 @@ def run_job_checks(job_id):
...
+            if unfinished_prerequisites == 0:
+                # this job has no unfinished prerequisite jobs,
+                # so it is eligible to be set to a 'ready'
+                # status and added to the queue
+                mark_job_status(dep_job_id, 'ready')
+                # add to the job queue
+                logging.getLogger('deprecated.info').info(
+                    'Sending job %s to job manager', dep_job_id)
+                # will move this later
+                from dataactcore.utils.jobQueue import enqueue
```
cmc333333: Nice workaround

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -237,22 +200,20 @@ def writeErrors(self, failures, interfaces, job_id, short_colnames, writer, warn
             elif failure[4] == "warning":
                 # write to warnings file
                 warning_writer.write([field_name,error_msg,str(row_number),failed_value,original_rule_label])
-            error_list.recordRowError(job_id,self.filename,field_name,error,row_number,original_rule_label,severity_id=severityId)
+            error_list.recordRowError(job_id,job.filename,field_name,error,row_number,original_rule_label,severity_id=severityId)
         return fatal_error_found
 
-    def runValidation(self, job_id, interfaces):
+    def runValidation(self, job, interfaces):
```
cmc333333: So much love for passing around models instead of querying based on ids for everything.

dataactvalidator/app.py
```diff
@@ -39,22 +42,39 @@ def before_request():
     def handle_response_exception(error):
         """Handle exceptions explicitly raised during validation."""
         logger.error(str(error))
+        job_id = g.get('job_id', None)
```
cmc333333: I like this. One risk we should watch out for, though, is documentation. We're setting a dynamic attribute in pseudo-global space; this creates an implicit protocol about the presence of that state and how to access it. Given that we're only using it in a few places, though, maybe we can just comment those lines? "job_id is set at the beginning of the validate function" or similar?

dataactvalidator/app.py
```diff
@@ -39,22 +42,39 @@ def before_request():
...
-            job = sess.query(Job).filter(Job.job_id == job_id).one()
-            job.status_id = JOB_STATUS_DICT['failed']
-            sess.commit()
+            job = sess.query(Job).filter(Job.job_id == job_id).one_or_none()
+            if job:
+                if job.filename is not None:
+                    writeFileError(job_id, job.filename, ValidationError.unknownError)
+                # next 2 lines are very temporary, until the job interface refactor is done
+                jobDb = JobHandler()
+                mark_job_status(job_id, 'failed')
```
cmc333333: We might need to be more careful about corner cases in this function as it's basically a top-level exception handler. Could any of these functions raise an exception that we need to handle immediately?

Also, looks like you don't need `JobHandler` anymore.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -567,47 +529,23 @@ def validate_job(self, request, interfaces):
...
-            _exception_logger.exception(str(e))
-            self.markJob(job_id, jobTracker, "invalid", self.filename,e.errorType, e.extraInfo)
-            return JsonResponse.error(e, e.status)
-        except ValueError as e:
-            _exception_logger.exception(str(e))
-            # Problem with CSV headers
-            exc = ResponseException(str(e),StatusCode.CLIENT_ERROR,type(e), ValidationError.unknownError) #"Internal value error"
-            self.markJob(job_id,jobTracker, "invalid", self.filename, ValidationError.unknownError)
-            return JsonResponse.error(exc, exc.status)
-        except Error as e:
```
cmc333333: It doesn't look like this CSV error is being explicitly caught anymore, which I think means that the job would be marked "failed" instead of "invalid" now. Maybe not a problem?

dataactvalidator/app.py
```diff
@@ -39,22 +42,39 @@ def before_request():
     def handle_response_exception(error):
```
cmc333333: Agreed that we probably want to combine this with the generic Exception handler somehow. Maybe just pull out the shared logic into a helper function?

dataactcore/utils/jobQueue.py
```diff
@@ -54,15 +56,14 @@ def enqueue(jobID):
 
 
 @contextmanager
-def job_context(task, interface_holder_class, job_id):
```
cmc333333: yayayaya

dataactcore/utils/jobQueue.py
```diff
@@ -72,21 +73,18 @@ def job_context(task, interface_holder_class, job_id):
             except MaxRetriesExceededError:
                 logger.warning('Job %s completely failed', job_id)
                 # Log the error
-                job_manager.getJobById(job_id).error_message = str(e)
-                mark_job_status(job_id, "failed")
-
-        job_manager.close()
+                job = sess.query(Job).filter_by(job_id=job_id).one_or_none()
+                if job:
+                    job.error_message = str(e)
```
cmc333333: Is this change ever committed? Might have had the same problem previously, too.

dataactvalidator/validation_handlers/validationManager.py
bsweger: Thanks for the sharp eyes here..exactly what's needed. I definitely agonized a bit over all of this when stripping out this `try/except`. What I think is happening:

* The csv-related validation errors (missing/invalid headers, etc.) are being raised as a `ValueError` "wrappedException" within a `ResponseException`.  [Example](https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/dataactvalidator/filestreaming/csvAbstractReader.py#L48)
* Admittedly, I'm not sure what the purpose of the 'wrappedException' is, except to provide more information for debugging. It doesn't seem to affect the error type checking itself.
* Thus, these errors are caught as a `ResponseException` and not a `ValueError` and therefore handled by `handle_response_exception` in app.py, which sets the job status to `invalid`

To confirm, I ran a validation on a headerless csv, and the job status was invalid.

If you want to dig more, here's some before and after output when trying to validate a file with no header row:

*before*
```
2016-11-17 15:37:35,030 INFO:deprecated.info:Validator response: {'wrappedType': "<class 'ValueError'>", 'errorType': "<class 'dataactcore.utils.responseException.ResponseException'>", 'message': 'Errors in header row: Missing: adjustmentstounobligatedbalancebroughtforward_cpe, agencyidentifier, allocationtransferagencyidentifier, availabilitytypecode, beginningperiodofavailability, borrowingauthorityamounttotal_cpe, budgetauthorityappropriatedamount_cpe, budgetauthorityavailableamounttotal_cpe, budgetauthorityunobligatedbalancebroughtforward_fyb, contractauthorityamounttotal_cpe, deobligationsrecoveriesrefundsbytas_cpe, endingperiodofavailability, grossoutlayamountbytas_cpe, mainaccountcode, obligationsincurredtotalbytas_cpe, otherbudgetaryresourcesamount_cpe, spendingauthorityfromoffsettingcollectionsamounttotal_cpe, statusofbudgetaryresourcestotal_cpe, subaccountcode, unobligatedbalance_cpe', 'trace': ["('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/validation_handlers/validationManager.py', 580, 'validate_job', 'self.runValidation(job_id, interfaces)')", "('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/validation_handlers/validationManager.py', 308, 'runValidation', 'bucketName, errorFileName, self.long_to_short_dict)')", "('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/filestreaming/csvLocalReader.py', 27, 'open_file', 'region, bucket, filename, csv_schema, bucket_name, error_filename, long_to_short_dict)')", '(\'/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/filestreaming/csvAbstractReader.py\', 133, \'open_file\', \'raise ResponseException("Errors in header row: " + str(error_string), StatusCode.CLIENT_ERROR, ValueError,ValidationError.headerError,**extra_info)\')'], 'wrappedMessage': 'Errors in header row: Missing: adjustmentstounobligatedbalancebroughtforward_cpe, agencyidentifier, allocationtransferagencyidentifier, availabilitytypecode, beginningperiodofavailability, borrowingauthorityamounttotal_cpe, budgetauthorityappropriatedamount_cpe, budgetauthorityavailableamounttotal_cpe, budgetauthorityunobligatedbalancebroughtforward_fyb, contractauthorityamounttotal_cpe, deobligationsrecoveriesrefundsbytas_cpe, endingperiodofavailability, grossoutlayamountbytas_cpe, mainaccountcode, obligationsincurredtotalbytas_cpe, otherbudgetaryresourcesamount_cpe, spendingauthorityfromoffsettingcollectionsamounttotal_cpe, statusofbudgetaryresourcestotal_cpe, subaccountcode, unobligatedbalance_cpe'}
```
*after*

```
2016-11-17 15:26:02,343 INFO:deprecated.info:Validator response: {'errorType': "<class 'dataactcore.utils.responseException.ResponseException'>", 'trace': ["('/Users/rebeccasweger/Dev/.virtualenvs/data-act-broker/lib/python3.4/site-packages/flask/app.py', 1639, 'full_dispatch_request', 'rv = self.dispatch_request()')", "('/Users/rebeccasweger/Dev/.virtualenvs/data-act-broker/lib/python3.4/site-packages/flask/app.py', 1625, 'dispatch_request', 'return self.view_functions[rule.endpoint](**req.view_args)')", "('dataactvalidator/app.py', 92, 'validate', 'return validation_manager.validate_job(request,interfaces)')", "('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/validation_handlers/validationManager.py', 541, 'validate_job', 'self.runValidation(job, interfaces)')", "('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/validation_handlers/validationManager.py', 269, 'runValidation', 'bucketName, errorFileName, self.long_to_short_dict)')", "('/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/filestreaming/csvLocalReader.py', 27, 'open_file', 'region, bucket, filename, csv_schema, bucket_name, error_filename, long_to_short_dict)')", '(\'/Users/rebeccasweger/Dev/data-act-broker/dataactvalidator/filestreaming/csvAbstractReader.py\', 133, \'open_file\', \'raise ResponseException("Errors in header row: " + str(error_string), StatusCode.CLIENT_ERROR, ValueError,ValidationError.headerError,**extra_info)\')'], 'message': 'Errors in header row: Missing: adjustmentstounobligatedbalancebroughtforward_cpe, agencyidentifier, allocationtransferagencyidentifier, availabilitytypecode, beginningperiodofavailability, borrowingauthorityamounttotal_cpe, budgetauthorityappropriatedamount_cpe, budgetauthorityavailableamounttotal_cpe, budgetauthorityunobligatedbalancebroughtforward_fyb, contractauthorityamounttotal_cpe, deobligationsrecoveriesrefundsbytas_cpe, endingperiodofavailability, grossoutlayamountbytas_cpe, mainaccountcode, obligationsincurredtotalbytas_cpe, otherbudgetaryresourcesamount_cpe, spendingauthorityfromoffsettingcollectionsamounttotal_cpe, statusofbudgetaryresourcestotal_cpe, subaccountcode, unobligatedbalance_cpe', 'wrappedType': "<class 'ValueError'>", 'wrappedMessage': 'Errors in header row: Missing: adjustmentstounobligatedbalancebroughtforward_cpe, agencyidentifier, allocationtransferagencyidentifier, availabilitytypecode, beginningperiodofavailability, borrowingauthorityamounttotal_cpe, budgetauthorityappropriatedamount_cpe, budgetauthorityavailableamounttotal_cpe, budgetauthorityunobligatedbalancebroughtforward_fyb, contractauthorityamounttotal_cpe, deobligationsrecoveriesrefundsbytas_cpe, endingperiodofavailability, grossoutlayamountbytas_cpe, mainaccountcode, obligationsincurredtotalbytas_cpe, otherbudgetaryresourcesamount_cpe, spendingauthorityfromoffsettingcollectionsamounttotal_cpe, statusofbudgetaryresourcestotal_cpe, subaccountcode, unobligatedbalance_cpe'}
```

dataactbroker/handlers/fileHandler.py
bsweger: Heh, yeah. This is happening in several places. I went ahead and refactored out the `getJobsBySubmission` function, which was returning a list of job_ids so that calling code could loop through said list and get the jobs. Now everyone just pulls the jobs.

dataactvalidator/app.py
bsweger: Fair point. It seems safe enough, since this app only has one functional route, and the beginning of that route's code is putting the `job_id` in flask.g. I'll add some comments to this effect--good idea.

dataactbroker/handlers/fileHandler.py
```diff
@@ -932,11 +929,9 @@ def list_submissions(self, page, limit, certified):
...
 
         for submission in submissions:
-            job_ids = self.interfaces.jobDb.getJobsBySubmission(submission.submission_id)
-            total_size = 0
-            for job_id in job_ids:
-                file_size = self.interfaces.jobDb.getFileSize(job_id)
-                total_size += file_size or 0
+            total_size = sess.query(func.sum(Job.file_size)).\
+                filter_by(submission_id=submission.submission_id).\
+                scalar()
```
cmc333333: Weird corner case -- does this need to account for null values? Will a sql sum correctly handle rows with null `file_size` (i.e. ignore or add zero) and do we need to worry about a null when no such jobs exist (e.g. `.scalar() or 0`)?

cmc333333:
At this point, there are lots of little nice to haves, but I'd be content to merge once we're comfortable that:
1. the `csv.Error` is accounted for or doesn't need to be
2. nulls are accounted for, or don't need to be

dataactvalidator/app.py
bsweger: Oops--thanks for the `JobHandler` heads up. In theory, as long as we're making sure the job exists, the `writeFileError` and `mark_job_status` functions here shouldn't error (barring something like a db connection failure). They either error out on a bad job id or catch their own exceptions.

dataactcore/utils/jobQueue.py
bsweger: ooooh, good catch.

dataactbroker/handlers/fileHandler.py
bsweger: Yeah, I was surprised when I tested this on a set of jobs with some `NULL` file sizes and got an actual number . Will update to handle the case where the query doesn't return any rows at all.

dataactvalidator/app.py
```diff
@@ -108,6 +97,16 @@ def runApp():
         port=CONFIG_SERVICES['validator_port']
     )
 
+def get_current_job():
```
cmc333333: Good idea

dataactvalidator/app.py
```diff
@@ -41,45 +42,33 @@ def handle_response_exception(error):
...
-            sess = GlobalDB.db().session
-            job = sess.query(Job).filter(Job.job_id == job_id).one_or_none()
-            if job:
-                if job.filename is not None:
-                    writeFileError(job_id, job.filename, ValidationError.unknownError)
-                mark_job_status(job_id, 'failed')
-
-        return JsonResponse.error(error, 500)
+        # csv-specific errors get a different job status and response code
+        if isinstance(error, ValueError) or isinstance(error, csv.Error):
```
cmc333333: No biggie, but we could do this with a separate
```python
@app.errorhandler(ValueError)
@app.errorhandler(csv.Error)
def handle_client_exception(error):
    ...
```

cmc333333:
Great stuff @bsweger; I particularly like the error handling cleanups!

PR #412, opened 2016-11-17
Db 1309 signed s3 url updates

alburde1:
Updating `/sign_submission_file` route.

- Updating to POST instead of GET
- No longer expecting route parameters, instead expects request body
- Returning JSON instead of auto redirecting or downloading
- Updated related tests


dataactbroker/handlers/fileHandler.py
```diff
@@ -141,18 +141,20 @@ def get_signed_url_for_submission_file(self):
...
+            submission_id = safe_dictionary.getValue("submission")
             submission = self.jobManager.getSubmissionById(submission_id)
             # Check that user has access to submission
             # If they don't, throw an exception
             self.check_submission_permission(submission)
 
+            response_dict = {}
             if self.isLocal:
-                return send_from_directory(self.serverPath, file_name)
+                response_dict["url"] = os.path.join(self.serverPath, file_name)
```
cmc333333: Should this _always_ join with `/` instead? If the server's running windows, we'd still want forward slashes in the generated url, right?

tests/unit/dataactbroker/test_handlers_fileHandler_sign_url.py
```diff
@@ -2,31 +2,42 @@
...
+    interfaces = InterfaceHolder()
+    file_handler = fileHandler.FileHandler(Mock(), interfaces=interfaces, isLocal=False)
+    file_handler.check_submission_permission = Mock()
+
+    mock_dict = Mock()
+    mock_dict.return_value.getValue.side_effect = ['file_name', str(submission.submission_id)]
+    monkeypatch.setattr(fileHandler, 'RequestDictionary', mock_dict)
+
+    mock_dict = Mock()
+    mock_dict.return_value.getSignedUrl.side_effect = ['/signed/url/path/file_name.csv']
```
cmc333333: Little detail, but if you want this to always return just one value, you should use the more standard
```python
mock_dict.return_value.getSignedUrl.return_value = '/signed/url/path/file_name.csv'
```

dataactbroker/handlers/fileHandler.py
alburde1: This is how we've been doing it the whole time (see above function) so I have to assume it's been working :)

tests/unit/dataactbroker/test_handlers_fileHandler_sign_url.py
alburde1: Updated this, thanks for catching it.

cmc333333:
:+1: short and sweet

PR #413, opened 2016-11-17
Db 703 flex fields

michaeldhess:
Allowing users to submit files with "_flex" columns that are added to the error reports.

- Read "flex_" columns into flex dictionary
- Store columns in flex_fields table
- Add flex data to error reports


klrBAH:

klrBAH:
[run tests]

dataactcore/models/stagingModels.py
```diff
@@ -21,6 +21,22 @@ def concatTas(context):
...
+    job_id = Column(Integer, nullable=False, index=True)
+    row_number = Column(Integer, nullable=False)
+    header = Column(Text)
+    cell = Column(Text)
+    
+    def __init__(self, **kwargs):
+        # broker is set up to ignore extra columns in submitted data
+        # so get rid of any extraneous kwargs before instantiating
+        cleanKwargs = {k: v for k, v in kwargs.items() if hasattr(self, k)}
+        super(FlexField, self).__init__(**cleanKwargs)
```
cmc333333: Is this required? We have it on another model as a convenience, but hopefully we don't need it here as well.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -167,8 +174,9 @@ def readRecord(self,reader,writer,file_type,interfaces,row_number,job_id,fields,
                 writer.write(["Formatting Error", ValidationError.readErrorMsg, str(row_number), ""])
                 error_list.recordRowError(job_id,self.filename,"Formatting Error",ValidationError.readError,row_number,severity_id=interfaces.validationDb.getRuleSeverityId("fatal"))
                 row_error_found = True
-            return {}, reduce_row, True, False, row_error_found
-        return record, reduce_row, False, False, row_error_found
+
+            return {}, reduce_row, True, False, row_error_found, None
+        return record, reduce_row, False, False, row_error_found, flex_cols
```
cmc333333: Given how many args we're returning, would it make sense to define a `namedtuple` to make this a bit easier to read? Relatedly, would you mind modifying the docstring to explain what the return values are? Finally, is there any reason to return `None` for the flex_cols but `{}` for the record? Seems like we should do one or the other but not both.

dataactvalidator/filestreaming/csvAbstractReader.py
```diff
@@ -32,6 +32,7 @@ def open_file(self, region, bucket, filename, csv_schema, bucket_name, error_fil
         self.unprocessed = ''
         self.extra_line = False
         self.lines = []
+        self.flex_dictionary = {}
         self.header_dictionary = {}
```
cmc333333: While we're here, does it make sense to replace these with lists? It looks like we're only indexing them by incrementing integer values.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -145,9 +146,14 @@ def readRecord(self,reader,writer,file_type,interfaces,row_number,job_id,fields,
...
+            record = FieldCleaner.cleanRow(next_record, self.long_to_short_dict, fields)
             record["row_number"] = row_number
+            if not flex_cols:
+                flex_cols = None
+            else:
+                flex_cols["row_number"] = row_number
+
             if reader.is_finished and len(record) < 2:
                 # This is the last line and is empty, don't record an error
                 return {}, True, True, True, False  # Don't count this row
```
cmc333333: Does this need a sixth element?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -240,6 +247,30 @@ def writeErrors(self, failures, interfaces, job_id, short_colnames, writer, warn
...
+            Boolean indicating whether to skip current row
+        """
+        sess = GlobalDB.db().session
+        try:
+            flex_cols["job_id"] = job_id
+            flex_cols["submission_id"] = submission_id
+
+            sess.add(FlexField(**flex_cols))
+            sess.commit()
+        except ResponseException as e:
```
cmc333333: I don't think the above would raise a ResponseException -- is there some sqlalchemy exception you want to catch instead?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -200,15 +197,27 @@ def validateFileBySql(cls, submissionId, fileType, short_to_long_dict):
             if failures.rowcount:
                 # Create column list (exclude row_number)
                 cols = failures.keys()
+
+                # Create flex column list
+                flex_dict = {}
+                flex_results = sess.execute("SELECT * FROM flex_field WHERE submission_id="+str(submissionId))
```
cmc333333: Would you mind using the ORM here? It'd make it much clearer to understand what fields are being accessed.

dataactvalidator/validation_handlers/validationManager.py
cmc333333: Sure thing! The other two `return`s in this function are returning a six-element tuple, but this one is only returning a five-element tuple. I think you just missed one `return` when you added that sixth element to the other two.

cmc333333:
Looks good! @michaeldhess has agreed to a follow up PR to add tests.

PR #414, opened 2016-11-17
Db 1105 make data submission work locally

alburde1:
Allowing D files to be generated locally

- CONFIGS HAVE CHANGED! <- will need to remind everyone about this on the channel when this is merged
- Added sample D files in tests/integration/data folder so the system can copy them instead of trying to generate new ones locally
- Updated contribution docs to mention that virtualenvwrapper with Python 3 and Windows on occasion don't play nice.


dataactbroker/handlers/fileHandler.py
```diff
@@ -666,15 +671,22 @@ def call_d_file_api(self, api_url):
...
+                if response.status_code != 200:
+                    # Could not download the file, return False
+                    return False
+                # write to file
+                response.encoding = "utf-8"
+                file.write(response.text)
+                return True
+        else:
+            try:
+                copyfile(file_url, local_file_path)
```
cmc333333: Solid

dataactbroker/handlers/fileHandler.py
```diff
@@ -859,31 +871,39 @@ def getProtectedFiles(self):
         response["urls"] = self.s3manager.getFileUrls(bucket_name=CONFIG_BROKER["static_files_bucket"], path=CONFIG_BROKER["help_files_path"])
         return JsonResponse.create(StatusCode.OK, response)
 
-    def completeGeneration(self, generationId):
+    def completeGeneration(self, generationId, file_type=None):
```
cmc333333: It looks like a good chunk of this method is specific to processing the remote file. What do you think about splitting this apart and only calling the bits relevant to local generation from `addJobInfoForDFile`?

dataactbroker/handlers/fileHandler.py
```diff
@@ -859,31 +871,39 @@ def getProtectedFiles(self):
...
+                        StatusCode.CLIENT_ERROR
+                    )
+
+                url = request_dict['href']
+                _smx_logger.debug('Download URL => %s', url)
+            else:
+                if file_type == "D1":
+                    url = CONFIG_SERVICES["d1_file_path"]
+                else:
+                    url = CONFIG_SERVICES["d2_file_path"]
```
cmc333333: Quick question -- do we need this flexibility, or would it be simpler to just load the files from the same place in the repository tree each time?

tests/integration/fileTests.py
```diff
@@ -470,7 +470,7 @@ def test_file_generation(self):
         json = response.json
         self.assertIn(json["status"], ["waiting","finished"])
         self.assertEqual(json["file_type"], "D1")
-        self.assertEqual("#", json["url"])
+        self.assertIs("url" in json, True)
```
cmc333333: Super minor, but we'd get better messages from:
```python
self.assertIn("url", json)
# or
assert "url" in json
```

dataactbroker/handlers/fileHandler.py
alburde1: They're local files so the file path will be unique to the person every time. Right now, for me, they're in the same place as the test files for integration tests but other people may have other preferences, plus they can name them what they want. I think we would run into more problems than it's worth to load the files from the same place every time.

tests/integration/fileTests.py
alburde1: Thanks! Updated

cmc333333:
Looks good; I'm excited to use.

PR #415, opened 2016-11-17
DB 1346 ignore empty rows in submissions

alburde1:
+ Updated variable names within the validate method
+ Updated tests to match the new numbers of errors
+ Ignores completely blank rows in the submission files

### DIFF
--- a/dataactvalidator/validation_handlers/validator.py
+++ b/dataactvalidator/validation_handlers/validator.py
@@ -54,7 +54,7 @@ class Validator(object):
         return failures
 
     @classmethod
-    def validate(cls, record, csvSchema):
+    def validate(cls, record, csv_schema):
         """
         Run initial set of single file validation:
         - check if required fields are present
@@ -63,7 +63,7 @@ class Validator(object):
 
         Args:
         record -- dict representation of a single record of data
-        csvSchema -- dict of schema for the current file.
+        csv_schema -- dict of schema for the current file.
 
         Returns:
         Tuple of three values:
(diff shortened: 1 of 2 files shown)


dataactvalidator/validation_handlers/validator.py
```diff
@@ -63,58 +63,66 @@ def validate(cls, record, csvSchema):
...
-                recordFailed = True
-                failedRules.append([fieldName, ValidationError.lengthError, currentData,"", "warning"])
-
-        return (not recordFailed), failedRules, (not recordTypeFailure)
+                record_failed = True
+                failed_rules.append([field_name, ValidationError.lengthError, current_data,"", "warning"])
+
+        # if all columns are blank (empty row), set it so it doesn't add to the error messages or write the line, just ignore it
+        if total_fields == blank_fields:
+            record_failed = False
```
cmc333333: Is there a risk of overriding a `record_failed` that was set in that loop?

alburde1: Not a bad risk. If the `blank_fields` total is equal to the `total_fields` then everything in there is blank and it shouldn't matter. It will actually override it because, technically, it's going to set it to `True` earlier because we're missing required column headers, but we want it to be `False` so it can be ignored. It's only checking one row at a time so we won't be messing up `record_failed` for other rows.

alburde1:
Do we need a unit test? The integration tests have several files with blank rows that should fail if this isn't working right.

dataactvalidator/validation_handlers/validator.py
cmc333333: Good call!

cmc333333:
Fair enough; they're all magic numbers to me, but the code's being exercised.

PR #417, opened 2016-11-17
Remove RouteUtils

cmc333333:
As part of an unrelated task, I removed the `RouteUtils` indirection layer. I figured it was worth a separate PR.


dataactbroker/app.py
```diff
@@ -66,6 +68,16 @@ def before_request():
...
 
+    @app.errorhandler(ResponseException)
+    def handle_response_exception(exception):
+        return JsonResponse.error(exception, exception.status)
+
+    @app.errorhandler(Exception)
+    def handle_exception(exception):
+        wrapped = ResponseException(str(exception), StatusCode.INTERNAL_ERROR,
+                                    type(exception))
+        return JsonResponse.error(wrapped, wrapped.status)
```
alburde1: I was wondering how we were going to handle the exceptions when we removed RouteUtils, great solution.

tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -4,7 +4,7 @@
 from tests.unit.dataactbroker.utils import add_models, delete_models
 from tests.unit.dataactcore.factories.job import SubmissionFactory, JobFactory
 from tests.unit.dataactcore.factories.user import UserFactory
-from dataactbroker.routeUtils import InterfaceHolder
+from dataactcore.interfaces.interfaceHolder import InterfaceHolder
```
alburde1: Why were we even going the long way around for this to begin with? Seems silly.

PR #418, opened 2016-11-18
Changing login error text to include link to MAX to request group access

nmonga91:
Linking login errors to MAX's webpage to request group access


cmc333333:
:+1:

PR #419, opened 2016-11-18
Add submission narrative models/endpoints

cmc333333:
Users would like to be able to add notes ("narratives") about their files. We'll allow appropriate users to do this. This uses a more RESTful set of endpoints, but I'm happy to remove if we think it's better to keep the existing, imperative style.

Easiest to review changeset by changeset


dataactbroker/README.md
```diff
@@ -843,6 +843,58 @@ Get total obligations and specific obligations. Calls to this route should inclu
...
+{}
+```
+
+
+#### GET "/v1/submission/<int:submission_id>/narrative"
+Retrieve existing submission narratives (explanations/notes for particular
+files). Submission id should be the integer id associated with the submission
+in question. Users must have appropriate permissions to access these
+narratives (write access for the agency of the submission, or SYS).
+
```
alburde1: Is this a duplicate of the section starting at line 847 without the response?

alburde1:
I see no problem with changing the style of API calls but I also haven't gotten super used to the old style so it's not a hard change for me. Might be good to get a second opinion.

dataactbroker/fileRoutes.py
```diff
@@ -138,3 +139,17 @@ def sign_submission_file():
         fileManager = FileHandler(request, isLocal=IS_LOCAL, serverPath=SERVER_PATH)
         fileManager.addInterfaces(InterfaceHolder())    # soon to be removed
         return fileManager.get_signed_url_for_submission_file()
+
+    @app.route("/v1/submission/<int:submission_id>/narrative", methods=['GET'])
```
bsweger: Thanks for nudging us in this direction!

dataactbroker/handlers/fileHandler.py
```diff
@@ -41,6 +43,30 @@
 _smx_logger = logging.getLogger('deprecated.smx')
 
 
+def user_agency_matches(submission):
```
bsweger: much better as a separate function--thanks!

tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -120,3 +124,87 @@ def test_list_submissions_failure(database, job_constants, monkeypatch):
...
+    fn(sub.submission_id)
+
+    monkeypatch.setattr(fileHandler, 'user_agency_matches',
+                        Mock(return_value=False))
+    with pytest.raises(ResponseException):
+        fn(sub.submission_id)
+
+
+def test_narratives(database, job_constants, monkeypatch):
+    """Integration test that verifies we can add, retrieve, and update
```
bsweger: Is it important to retain our current distinction, location-wise, between unit and integration tests?

cmc333333: Eh, I don't have clear definitions, so I'm probably using the term incorrectly. I meant "this isn't testing a single unit of work, it's testing a few at once". I can swap out to be more explicit if that'd be better?

bsweger:
@cmc333333 This looks good to me--I'm ready to merge if you are. Not for this PR, but should we consider telling CodeClimate to ignore the Alembic migrations?

cmc333333:
Thanks @bsweger! Updated a docstring. I agree about code climate; I started work on a config but saw all the potential options and punted :)

bsweger:
Thanks for the doctring clarification! Merging

PR #424, opened 2016-11-22
The next installment of the job database refactor

bsweger:
There's more work to do, but this is a reasonable stopping point for a review.

@alburde1 and I tag-teamed on this one to remove/move a few more of the old job database functions from handlers and interfaces.

This is probably best reviewed commit by commit. There are few large commits at the end that could use extra scrutiny, since they impact how the jobs and job dependencies are created. I tried to simplify a bit when moving the code to `function_bag.py`, but the main focus is still on getting code out of the interfaces.


dataactbroker/handlers/fileHandler.py
```diff
@@ -58,7 +59,9 @@ class FileHandler:
     STATUS_MAP = {"waiting":"invalid", "ready":"invalid", "running":"waiting", "finished":"finished", "invalid":"failed", "failed":"failed"}
     VALIDATION_STATUS_MAP = {"waiting":"waiting", "ready":"waiting", "running":"waiting", "finished":"finished", "failed":"failed", "invalid":"failed"}
 
-    def __init__(self,request,interfaces = None,isLocal= False,serverPath =""):
+    UploadFile = namedtuple('UploadFile', ['file_type', 'upload_name', 'file_name', 'upload_job_id'])
```
cmc333333: It doesn't look like `upload_job_id` is ever used. Do we need it for something? 10 points for using a named tuple here.

dataactcore/interfaces/function_bag.py
```diff
@@ -529,7 +406,197 @@ def create_submission(user_id, submission_values, existing_submission_id):
...
+            val_dependency = JobDependency(job_id=validation_job.job_id, prerequisite_id=job_id)
+            sess.add(val_dependency)
+            ext_dependency = JobDependency(job_id=external_job.job_id, prerequisite_id=job_id)
+            sess.add(ext_dependency)
+
+    sess.commit()
+    upload_dict["submission_id"] = submission_id
+    return upload_dict
+
+def add_jobs_for_file_type(upload_file, submission_id, existing_submission):
```
cmc333333: No biggie, but the `for_file_type` here doesn't seem to make sense anymore. `for_uploaded_file`? Also, this is outside the scope for this PR, but this function is waaaaay too big. We should break it up.

dataactbroker/handlers/jobHandler.py
```diff
@@ -77,83 +77,7 @@ def createDate(dateString):
...
-
-        # First do award_financial and award_procurement jobs so they will be available for later dependencies
-        for fileType, filePath, filename in filenames:
-            if fileType in ["award_financial", "award_procurement"]:
-                jobsRequired, uploadDict = addJobsForFileType(fileType, filePath, filename, submissionId, existingSubmission, jobsRequired, uploadDict)
-
-        # Then do all other file types
-        for fileType, filePath, filename in filenames:
-            if fileType not in ["award_financial", "award_procurement"]:
-                jobsRequired, uploadDict = addJobsForFileType(fileType, filePath, filename, submissionId, existingSubmission, jobsRequired, uploadDict)
```
cmc333333: It looks like we're losing some of this ordering (`award_financial` and `award_procurement`before the other files). Is that important to keep? We could just sort the `uploaded_files`

dataactbroker/handlers/fileHandler.py
```diff
@@ -387,82 +428,82 @@ def getStatus(self):
...
                     continue
 
-                jobInfo["job_id"] = job.job_id
-                jobInfo["job_status"] = job.job_status.name
-                jobInfo["job_type"] = job_type
-                jobInfo["filename"] = self.jobManager.getOriginalFilenameById(job.job_id)
+                job_info["job_id"] = job.job_id
+                job_info["job_status"] = job.job_status.name
+                job_info["job_type"] = job_type
+                job_info["filename"] = sess.query(Job).filter_by(job_id = job.job_id).one().original_filename
```
cmc333333: Can we just use `job.original_filename` here?

dataactbroker/handlers/fileHandler.py
```diff
@@ -387,82 +428,82 @@ def getStatus(self):
...
+                    job_info["error_type"] = getErrorType(job.job_id)
+                    job_info["error_data"] = getErrorMetricsByJobId(
                         job.job_id, job_type=='validation', severity_id=RULE_SEVERITY_DICT['fatal'])
-                    jobInfo["warning_data"] = getErrorMetricsByJobId(
+                    job_info["warning_data"] = getErrorMetricsByJobId(
                         job.job_id, job_type=='validation', severity_id=RULE_SEVERITY_DICT['warning'])
                 # File size and number of rows not dependent on error DB
                 # Get file size
-                jobInfo["file_size"] = self.jobManager.getFileSizeById(job.job_id)
+                job_info["file_size"] = sess.query(Job).filter_by(job_id = job.job_id).one().file_size
```
cmc333333: Similar here

dataactbroker/handlers/fileHandler.py
```diff
@@ -387,82 +428,82 @@ def getStatus(self):
...
                 # Get number of rows in file
-                jobInfo["number_of_rows"] = self.jobManager.getNumberOfRowsById(job.job_id)
+                job_info["number_of_rows"] = sess.query(Job).filter_by(job_id = job.job_id).one().number_of_rows
```
cmc333333: Similar here

dataactbroker/handlers/fileHandler.py
```diff
@@ -387,82 +428,82 @@ def getStatus(self):
...
 
                 try :
-                    jobInfo["file_type"] = self.jobManager.getFileType(job.job_id)
+                    job_info["file_type"] = sess.query(Job).options(joinedload("file_type")).filter_by(job_id = job.job_id).one().file_type.name
```
cmc333333: Looks like we've already got a job here -- would it make sense to just walk over to
```python
job.file_type.name
```
? It's actually already preloaded due to `lazy='joined'`

dataactbroker/handlers/fileHandler.py
```diff
@@ -702,15 +745,16 @@ def get_lines_from_csv(self, file_path):
...
             full_file_path = "".join([CONFIG_BROKER['d_file_storage_path'], timestamped_name])
 
             _smx_logger.debug('Downloading file...')
             if not self.download_file(full_file_path, url):
                 # Error occurred while downloading file, mark job as failed and record error message
                 mark_job_status(job_id, "failed")
-                job = job_manager.getJobById(job_id)
-                file_type = job_manager.getFileType(job_id)
+                job = sess.query(Job).filter_by(job_id = job_id).one()
+                file_type = sess.query(Job).options(joinedload("file_type")).filter_by(job_id = job_id).one().file_type.name
```
cmc333333: Similar here, we can just do `job.file_type.name` as `Job`s prefetch `file_type`.

dataactcore/interfaces/function_bag.py
```diff
@@ -501,3 +378,225 @@ def check_job_dependencies(job_id):
...
+        if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
+            submission.publish_status_id = PUBLISH_STATUS_DICT['updated']
+        # submission is being updated, so turn off publishable flag
+        submission.publishable = False
+        for key in submission_values:
+            # update existing submission with any values provided
+            setattr(submission, key, submission_values[key])
+
+    sess.commit()
+    return submission
```
cmc333333: +1 to returning an object rather than an ID

dataactbroker/handlers/fileHandler.py
bsweger: @cm good catch--`upload_job_id` is something I was playing around with and ended up not using. I knew you'd like the named tuple!

dataactbroker/handlers/fileHandler.py
```diff
@@ -189,32 +188,59 @@ def submit(self,name,CreateCredentials):
...
+                    # convert incoming dates to Python date objects
+                    try:
+                        submission_data[value] = datetime.strptime(request_value, date_format)
+                    except ValueError:
+                        raise ResponseException("Date must be provided as MM/YYYY", StatusCode.CLIENT_ERROR,
+                                                ValueError)
+                elif request_value:
+                    submission_data[value] = request_value
+                else:
+                    if not existing_submission:
```
bsweger: Yes, I like that--will update. Thanks!

dataactbroker/handlers/fileHandler.py
```diff
@@ -307,9 +310,10 @@ def checkSubmissionById(self, submission_id, file_type):
 
         """
         error = None
+        sess = GlobalDB.db().session
 
         try:
-            submission = self.interfaces.jobDb.getSubmissionById(submission_id)
+            submission = sess.query(Submission).filter_by(submission_id = submission_id).one()
         except ResponseException as exc:
```
bsweger: Yes, absolutely--we can't just ignore that logic. Just submitted a change for this. It's not _quite_ the same thing but should ensure that we're sending the same info back to the front-end.

dataactbroker/handlers/jobHandler.py
```diff
@@ -247,11 +249,12 @@ def getSubmissionForJob(self,job):
 
     def sumNumberOfRowsForJobList(self, jobs):
         """ Given a list of job IDs, return the number of rows summed across jobs """
+        sess = GlobalDB.db().session
         # temporary fix until jobHandler.py is refactored away
         jobList = [j.job_id for j in jobs]
         rowSum = 0
         for jobId in jobList:
-            jobRows = self.getNumberOfRowsById(jobId)
+            jobRows = sess.query(Job).filter_by(job_id = jobId).one().number_of_rows
```
bsweger: Absolutely. This entire function will move to `function_bag` as we finish up this refactor, so I'll defer working on it now, if that's ok.

cmc333333:
Nice work @bsweger and @alburde1 !

PR #426, opened 2016-11-22
Set tas_id on imported SF133s

cmc333333:
Also fixes a regression with SF133 amount imports and updates import docs.

Easiest to review commit by commit.


dataactvalidator/scripts/load_sf133.py
```diff
@@ -61,6 +65,66 @@ def fill_blank_sf133_lines(data):
...
+    absolute_fiscal_month = 12 * fiscal_year + zero_based_period - 3
+    absolute_following_month = absolute_fiscal_month + 1
+    start_date = date(absolute_fiscal_month // 12,
+                      (absolute_fiscal_month % 12) + 1,     # 1-based
+                      1)
+    end_date = date(absolute_following_month // 12,
+                    (absolute_following_month % 12) + 1,    # 1-based
+                    1)
+
+    # Why min()?
```
nmonga91: So if there are multiple active tas entries, we just care about the one that was there first?

dataactcore/migrations/versions/a62138efd429_add_tas_id_to_sf133.py
```diff
@@ -0,0 +1,42 @@
...
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('sf_133', sa.Column('tas_id', sa.Integer(), nullable=True))
+    op.create_foreign_key('fk_tas', 'sf_133', 'tas_lookup', ['tas_id'], ['tas_id'])
```
cmc333333: I _think_ postgres scopes FK's to their table, but this is good advice. Will implement, thanks!

dataactvalidator/scripts/load_sf133.py
cmc333333: Yeah -- there _shouldn't_ be multiple active tas entries, but it's _possible_. Min is just a consistent tie breaker.

nmonga91:
Looks good, ready to merge

PR #428, opened 2016-11-23
Use the presence of a field rather than its value

cmc333333:
The values can be falsey, like is_quarter == False. I *think* this logic
matches our intent, but please double check me.

Thanks @bsweger for explaining what this was supposed to do!

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -231,16 +231,20 @@ class FileHandler:
             existing_submission_id = request_params.get('existing_submission_id')
             existing_submission = True if existing_submission_id else None
             for request_field, submission_field in request_submission_mapping.items():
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -231,16 +231,20 @@ def submit(self, user_id, create_credentials):
...
+                    if 'date' in request_field:
+                        # convert incoming dates to Python date objects
+                        try:
+                            request_value = datetime.strptime(request_value, date_format)
+                        except ValueError:
+                            raise ResponseException("Date must be provided as MM/YYYY", StatusCode.CLIENT_ERROR,
+                                                    ValueError)
+
                     submission_data[submission_field] = request_value
+                # All of those fields are required unless
```
bsweger: Good idea to comment this--thanks!

PR #429, opened 2016-11-28
Db 1385 permissions cleanup

nmonga91:
- Updating CGAC on login
- Allowing null permission to access help page
- Updating unauthorized message text

- [x] Reviewed by backend
- [ ] Merge coordinated with frontend


dataactbroker/handlers/accountHandler.py
```diff
@@ -231,15 +233,15 @@ def grant_highest_permission(self, session, user, group_list, cgac_group):
             permission_group = [g for g in group_list if g.startswith(cgac_group + "-PERM_")]
             # Check if a user has been placed in a specific group. If not, deny access
             if not permission_group:
-                raise ValueError("You have logged in with MAX but do not have permission to access the broker.")
```
cmc333333: I _think_ removing this line means we no longer need 221-223.

nmonga91: I was thinking we would still need it because of line 161 & 172

nmonga91:
@cmc333333 Yes, the goal is to allow users who are in MAX to still reach the site. But since their permission is `null`, the front-end would handle it differently and restrict access there (for viewing purposes). For actual execution in the API, I may need to account for that null permission in `permissions.py`

dataactbroker/handlers/accountHandler.py
cmc333333: Womp womp, good call

cmc333333:
All good!

PR #430, opened 2016-11-28
Remove job interfaces

bsweger:
This PR removes the last of the old job database interfaces. Since this was the last interface standing, we also remove the `InterfaceHolder` and `BaseInterface` classes. Oh happy day!


dataactbroker/handlers/fileHandler.py
```diff
@@ -702,7 +699,7 @@ def handleEmptyResponse(self, job, valJob):
             valJob.number_of_errors = 0
             valJob.number_of_warnings = 0
             valJob.filename = None
-            jobDb.session.commit()
+            sess.commit()
```
alburde1: Do we need a `sess.commit()` both here and on line 695? Can we remove the first one? Do we need it on line 687 either? Can we just `sess.commit()` at the very end of this function outside the `if`?

dataactbroker/scripts/createGenerationJobs.py
```diff
@@ -1,47 +0,0 @@
-from dataactcore.interfaces.interfaceHolder import InterfaceHolder
```
alburde1: Did we not use this file at all? Good catch if so

dataactbroker/handlers/domainHandler.py
```diff
@@ -7,23 +7,10 @@
 from dataactcore.utils.statusCode import StatusCode
 from dataactbroker.handlers.aws.session import LoginSession
 
-# todo: determine if we even need this DomainHandler if we don't have interfaces. It will only contain one static function
+# todo: determine if we even need this DomainHandler. It will only contain one static function
```
bsweger: You're right--will change this.

dataactbroker/handlers/fileHandler.py
```diff
@@ -434,13 +419,24 @@ def getStatus(self):
...
+            # Format submission reporting date
+            if submission.is_quarter_format:
+                reporting_date = 'Q{}/{}'.format(
+                    int(submission.reporting_fiscal_period / 3), submission.reporting_fiscal_year)
+            else:
+                reporting_date = submission.reporting_start_date.strftime("%m/%Y")
+            # Broker allows submission for a single quarter or a single month,
+            # so reporting_period start and end dates reported by check_status
+            # are always equal
+            submission_info["reporting_period_start_date"] = submission_info["reporting_period_end_date"] = reporting_date
```
bsweger: That's a good question. A long time ago, broker users could actually specify submission start/end dates that included a day (_e.g._, 7/13/2015 - 8/09/2016). Thus, our models are set up to accommodate this format. 

As the broker evolved, we realized that real-world submissions will only come in 2 flavors:
* A single quarter (this is the only submission type eligible for certification and import to usaspending). The` check_status` date format for this is `Q1/2017`, for example.
* A single month (for agencies to do internal data testing). The `check_status` date format for this is `10/2016`.

With this PR, we're now depending on the front-end to send us a start/end date that are both in the same quarter (for quarterly submission) and a start/end date in the same month (for monthly submissions). That makes the actual day irrelevant. 

You can see this in the current code by looking at the reporting start/end dates being returned by the `check_status` route. 

Should I try to clarify the comment? Do you think it makes sense to add some logic/testing around the submission reporting dates to make sure we don't get anything other than a single quarter or single month?

dataactbroker/handlers/fileHandler.py
bsweger: Good call. That's, uh, a lot of commits, especially considering that the `mark_job_status` function being called is also committing. Will remove some of these. 

I do think that a longer-term fun project would be to commit based on a sensible unit of work. The committing-at-the-bottom-of-every-function thing is an artifact of the old db setup.

dataactbroker/scripts/createGenerationJobs.py
bsweger: @alburde1 It's an old data migration from back in the day (before we did data migrations via alembic). So we don't need it going forward, and it will live on in repo history should we ever need to reference it.

dataactbroker/handlers/fileHandler.py
alburde1: If we're updating it to only allow those, then I do think it makes sense to do a little bit of testing to make sure they stay the same. Nothing major and we probably don't need to check for edge cases (whatever those may be), just some super basic "is this the same" stuff.

I think the comment makes sense for anyone that knows about the format change so it's fine.

bsweger: @alburde1 just pushed some updates for this that splits the date checking into a separate function and adds unit tests

PR #431, opened 2016-11-28
Return 0 for filesize if S3 file not found

bsweger:
This is a hotfix to address an issue we had with initial validations getting stuck on dev. There's more to unpack here (_e.g._, is there a way to prevent the app from getting stuck when there's an unhandled exception like this, why is this suddenly a problem), but this should at least get folks past that first validation screen.


nmonga91:
Looks good, merging!

PR #433, opened 2016-11-29
.codeclimate.yml

bsweger:
First stab at a custom .codeclimate.yml for this project. It explicitly excludes alembic migrations (which were previously included) and anything in`tests/` (no change...these were excluded before).

In an effort to make the reports more actionable, also disabled the similar code check. The duplication engine is still enabled, but the similar code check specifically was turning up what I consider to be distracting false positives.

Obviously, open for discussion!

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,5 @@
+---
+languages:
+  Python: true
+exclude_paths:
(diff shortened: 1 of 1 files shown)


cmc333333:
Seems like a good start; I think that skipping tests and migrations is a-okay.

PR #436, opened 2016-11-30
merge-dev-stable into development

nmonga91:
- [ ] Reviewed carefully to ensure that nothing was unintentionally overridden


cmc333333:
Looks good to me. We may want to tweak those logs to be less verbose, but that can be done later.

PR #438, opened 2016-12-01
Db 1430 b5 update

alburde1:
Updating B5 rules based on https://federal-spending-transparency.atlassian.net/browse/DB-1430


cmc333333:
Looks like it matches the requested logic to me. Note that the CSV diff is largely whitespace changes (maybe the newline?)

PR #441, opened 2016-12-05
Hotfix SF133 speed

cmc333333:
We're planning to re-import SF133, but the current load process takes ~8hrs. Swapping in this change reduces that to ~3hrs.

### DIFF
--- a/dataactcore/models/domainModels.py
+++ b/dataactcore/models/domainModels.py
@@ -85,7 +85,7 @@ class SF133(Base):
     ending_period_of_availabil = Column(Text)
     main_account_code = Column(Text, nullable=False)
     sub_account_code = Column(Text, nullable=False)
-    tas = Column(Text, nullable=False, default=concatTas, onupdate=concatTas)
+    tas = Column(Text, nullable=False, default=concatTas)
     fiscal_year = Column(Integer, nullable=False)
     period = Column(Integer, nullable=False)
     line = Column(Integer,nullable=False)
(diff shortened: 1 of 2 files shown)


dataactcore/models/domainModels.py
```diff
@@ -85,7 +85,7 @@ class SF133(Base):
     ending_period_of_availabil = Column(Text)
     main_account_code = Column(Text, nullable=False)
     sub_account_code = Column(Text, nullable=False)
-    tas = Column(Text, nullable=False, default=concatTas, onupdate=concatTas)
```
nmonga91: Are we not concatenating on update because we now have a `tas_id` associated with each entry?

dataactvalidator/scripts/load_sf133.py
```diff
@@ -81,47 +87,49 @@ def update_tas_id(fiscal_year, fiscal_period):
...
-    sess.execute(
-        sql, {'start_date': start_date, 'end_date': end_date,
-              'fiscal_year': fiscal_year, 'fiscal_period': fiscal_period}
-    )
+    subquery = sess.query(sa.func.min(TASLookup.tas_id))
+
+    # Filter to matching TAS components, accounting for NULLs
+    for field in _tas_fields_to_match:
+        tas_col = getattr(TASLookup, field)
+        sf133_col = getattr(SF133, field[:26])
```
nmonga91: Is there a reason we're only looking at the first 26 characters of the field to look up in the SF133 domain model?

dataactcore/models/domainModels.py
cmc333333: We don't actually update SF133s other than through the new update command which sets `tas_id`. Removing it allows us to avoid pulling each SF133 from the db, run `concatTas` (which would produce the same result), and inserting it again. Does that make sense?

nmonga91:
Looks good, merging

PR #442, opened 2016-12-06
Fill in the rest of the data for file F

cmc333333:
We had skipped a handful of fields until we had a good sense of the FSRS data. Now that time has come, so let's fill in the rest of the details.


cmc333333:
Updated! As the mapping logic has gone through a few phases, it's probably easiest to just review the final mapping code, look at the associated tests, and make sure that everything's sensible. There shouldn't be a need to worry about the exact mappings between fields unless you want to compare with the IDD.

PR #445, opened 2016-12-06
WIP - Associate multiple agencies with a single user

cmc333333:
So far, this removes the `SYS` pseudo-agency from the UI, creates a `website_admin` field (replacing the associated permission) and creates a new table to association users with agencies + permissions.

This can be merged as is whenever, but I'll keep chugging along. I wanted to post it to get any initial feedback from @nmonga91


cmc333333:
Per @nmonga91 closing this in favor of smaller PRs

PR #447, opened 2016-12-08
Separate out superuser logic

cmc333333:
We'll be allowing users to have multiple agencies with different levels of permission for each, so having the `website_admin` permission on the agency level doesn't make sense. This also removes two fictitious permission names.

### DIFF
--- a/dataactbroker/handlers/accountHandler.py
+++ b/dataactbroker/handlers/accountHandler.py
@@ -277,7 +277,7 @@ class AccountHandler:
                 filter(CGAC.cgac_code == cgac_code).\
                 one_or_none()
             agency_name = "Unknown" if agency_name is None else agency_name
-            for user in sess.query(User).filter_by(permission_type_id=PERMISSION_TYPE_DICT['website_admin']):
+            for user in sess.query(User).filter_by(website_admin=True):
                 email_template = {'[REG_NAME]': username, '[REG_TITLE]':title, '[REG_AGENCY_NAME]':agency_name,
                                  '[REG_CGAC_CODE]': cgac_code,'[REG_EMAIL]' : user_email,'[URL]':link}
                 new_email = sesEmail(user.email, system_email,templateType="account_creation",parameters=email_template)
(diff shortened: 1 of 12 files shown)


dataactbroker/handlers/accountHandler.py
```diff
@@ -884,3 +865,30 @@ def email_users(self, system_email, session):
...
+        user.permission_type_id = None
+    else:
+        perms = [perm[-1].lower() for perm in permission_group]
+        ordered_perms = sorted(
+            PERMISSION_MAP.items(), key=lambda pair: pair[1]['order'])
+        for key, permission in ordered_perms:
+            name = permission['name']
+            if key in perms:
+                user.permission_type_id = PERMISSION_TYPE_DICT[name]
+                break
```
alburde1: I know we were already using the "highest permission" before, but what's the plan when we add multiple permissions? Someone could be "write" for multiple agencies right? So will we have to update this then or will this just work?

alburde1: Or, if they're write for one and read for another, I assume they can only do read for the one they're read in, so how is that going to work?

dataactbroker/permissions.py
```diff
@@ -31,21 +24,11 @@ def decorated_function(*args, **kwargs):
...
-                if permission == "check_email_token":
-                    if LoginSession.isRegistering(session):
-                        return f(*args, **kwargs)
-                    else :
-                        error_message = "unauthorized"
-                elif permission == "check_password_token":
-                    if LoginSession.isResetingPassword(session):
-                        return f(*args, **kwargs)
-                    else :
-                        error_message  = "unauthorized"
```
alburde1: Just clarifying, all the changes above this are because we no longer need these because we aren't in charge of registration and such?

dataactbroker/permissions.py
```diff
@@ -31,21 +24,11 @@ def decorated_function(*args, **kwargs):
...
-                elif LoginSession.isLogin(session):
+                if LoginSession.isLogin(session):
```
alburde1: Where is "session" defined and do we still need to pass it around? How is this session different than the sess you defined above it? More of a clarification question for me.

dataactbroker/userRoutes.py
```diff
@@ -13,22 +17,26 @@ def add_user_routes(app,system_email,bcrypt):
...
-        accountManager = AccountHandler(request,bcrypt = bcrypt)
-        return accountManager.register(system_email, session)
+        # check the session to make sure register is set to prevent any one
+        # from using route
+        if LoginSession.isRegistering(session):
+            accountManager = AccountHandler(request,bcrypt = bcrypt)
+            return accountManager.register(system_email, session)
+        else:
+            return JsonResponse.create(StatusCode.LOGIN_REQUIRED,
+                                       {'message': 'unauthorized'})
```
alburde1: In terms of this function/route and the `update_user` and `delete_user` routes (and whatever other ones fall under this category), do we need to be updating them at all? I thought we were going to remove them entirely (can we do that yet?)

dataactbroker/handlers/accountHandler.py
cmc333333: Yep, that rewrite will come in a later PR. Right now we're passing just one permission in; I think in the future we'll pass multiple.

dataactbroker/permissions.py
cmc333333: Not quite -- I just moved the functionality to the respective routes. We only used each bit of logic in one place.

PR #448, opened 2016-12-08
Create UserAffiliations, a contextual user

cmc333333:
Builds on #447 ([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1386-1-separate-superuser...db-1386-2-start-affiliations)).

This
* Adds the current user to `g` and uses that in place of passing around user ids
* Creates the `UserAffiliation` model/table
* Populates that table from back data and whenever a user logs in


dataactbroker/fileRoutes.py
```diff
@@ -18,7 +17,7 @@ def add_file_routes(app,CreateCredentials,isLocal,serverPath,bcrypt):
     @permissions_check(permission="writer")
     def submit_files():
         fileManager = FileHandler(request,isLocal=IS_LOCAL, serverPath=SERVER_PATH)
-        return fileManager.submit(LoginSession.getName(session), CreateCredentials)
```
nmonga91: Good call. Even without the global changes, this could've been derived as such within the submit method instead of being passed as a parameter, no?

dataactbroker/handlers/accountHandler.py
```diff
@@ -584,9 +586,12 @@ def list_users(self):
     def list_user_emails(self):
         """ List user names and emails """
         sess = GlobalDB.db().session
-        user = sess.query(User).filter(User.user_id == LoginSession.getName(flaskSession)).one()
         try:
-            users = sess.query(User).filter(User.cgac_code == user.cgac_code, User.user_status_id == USER_STATUS_DICT["approved"], User.is_active == True).all()
+            users = sess.query(User).filter_by(
+                cgac_code=g.user.cgac_code,
+                user_status_id=USER_STATUS_DICT["approved"],
```
nmonga91: eventually, we'll need to move away from this `user_status` since we're not handling registration/approval in the broker anymore.

dataactbroker/handlers/accountHandler.py
```diff
@@ -802,13 +810,13 @@ def set_skip_guide(self, session):
                 )
             skip_guide = request_dict['skip_guide']
             if isinstance(skip_guide, bool):    # e.g. from JSON
-                user.skip_guide = skip_guide
+                g.user.skip_guide = skip_guide
             elif isinstance(skip_guide, str):
                 # param is a string, allow "true" or "false"
                 if skip_guide.lower() == "true":
```
nmonga91: Is it possible to simplify the assignment of `skip_guide` here? just curious if there's a way to handle these checks/conversions.

found some references here but nothing seems like a catch all for the situations here: http://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python

dataactbroker/handlers/accountHandler.py
```diff
@@ -892,3 +900,7 @@ def grant_highest_permission(user, group_list, cgac_group):
             if key in perms:
                 user.permission_type_id = PERMISSION_TYPE_DICT[name]
                 break
+    user.affiliations = [UserAffiliation(
+        cgac=sess.query(CGAC).filter_by(cgac_code=user.cgac_code).one(),
```
nmonga91: so this will update to add additional agency affiliations correct?

dataactbroker/handlers/aws/session.py
```diff
@@ -11,34 +11,6 @@ class LoginSession():
     """
     This class is a wrapper for the session object
     """
-
```
nmonga91: +1 for cleanup/simplification of getting the current user

dataactcore/migrations/versions/4260d9ddf25e_permission_data_migration.py
```diff
@@ -0,0 +1,36 @@
+"""Permission data migration
```
nmonga91: Great organization for having data migration as a separate revision file

dataactbroker/app.py
```diff
@@ -61,7 +62,12 @@ def teardown_appcontext(exception):
 
     @app.before_request
     def before_request():
-        GlobalDB.db()
+        sess = GlobalDB.db().session
+        # setup user
+        g.user = None
+        if session.get('name') is not None:
+            g.user = sess.query(User).filter_by(user_id=session['name']).\
```
cmc333333: Exactly. We can probably even use that to add the user id to all broker logs.

dataactbroker/fileRoutes.py
cmc333333: Indeed

dataactbroker/handlers/accountHandler.py
cmc333333: Agreed

dataactbroker/handlers/accountHandler.py
cmc333333: I think so. I'll take a swing

nmonga91:
Ready to merge once conflicts are resolved. I think this will need to coordinate with https://github.com/fedspendingtransparency/data-act-broker-backend/pull/454 because of the new session user referencing

dataactbroker/handlers/accountHandler.py
```diff
@@ -808,23 +808,13 @@ def set_skip_guide(self, session):
...
-                    g.user.skip_guide = True
-                elif skip_guide.lower() == "false":
-                    g.user.skip_guide = False
-                else:
-                    raise ResponseException(
-                        "skip_guide must be true or false",
-                        StatusCode.CLIENT_ERROR
-                    )
-            else:
+            skip_guide = str(request_dict['skip_guide']).lower()
```
nmonga91: ingenious! much cleaner, thank you for handling this

PR #450, opened 2016-12-08
Update SF-133 load script to pull from centralized SF-133 bucket

michaeldhess:
*Should use updated config


cmc333333:
Great idea

PR #451, opened 2016-12-08
Write unit test for the sf-133 files

michaeldhess:
*Update config_examples
*Update load_sf133 get_sf133_list() function to pull local files from config (needs updated config file)


tests/unit/dataactvalidator/test_sf133_content.py
```diff
@@ -0,0 +1,161 @@
...
+                    elif row.line=='2500':
+                        validate_two = float("{0:.2f}".format(row.amount))
+                sum_one = float("{0:.2f}".format(sum_one))
+                sum_two = float("{0:.2f}".format(sum_two))
+
+                validation_one = sum_one==validate_one
+                validation_two = sum_two==validate_two
+                validation_three = validate_one==validate_two
+
+                tas_m = TAS_RE.match(tas)
```
cmc333333: It feels off to smash together a bunch of fields to create the `tas`, then parse them out with a regular expression. Would it make sense to use a representative row or something like it to access the original field names?

tests/unit/dataactvalidator/test_sf133_content.py
```diff
@@ -0,0 +1,161 @@
...
+
+                # initial validations fail
+                if not (validation_one and validation_two) and (tas_m and tas_m.group('atc')!='X' and file_fiscal_year>int(tas_m.group('epoa'))):
+                    if not validation_three:
+                        failed_validations.append('{0},{1},{2},{3},{4},{5},{6},{7},1900!=2500,{8},{9}'.format(
+                            file_index, tas_m.group('ata'), tas_m.group('aid'), 
+                            tas_m.group('atc'), tas_m.group('bpoa'), tas_m.group('epoa'), 
+                            tas_m.group('mac'), tas_m.group('sac'), validate_one, validate_two)
+                        )
```
cmc333333: Small detail, but this might be easier to understand:
```python
failed_validations.append(",".join([
    file_index, .... , '1900!=2500', ...
]))
```

tests/unit/dataactvalidator/test_sf133_content.py
```diff
@@ -0,0 +1,161 @@
...
+                            file_index, tas_m.group('ata'), tas_m.group('aid'), 
+                            tas_m.group('atc'), tas_m.group('bpoa'), tas_m.group('epoa'), 
+                            tas_m.group('mac'), tas_m.group('sac'), sum_one, validate_one)
+                        )
+                    if not validation_two:
+                        failed_validations.append('{0},{1},{2},{3},{4},{5},{6},{7},sum!=2500,{8},{9}'.format(
+                            file_index, tas_m.group('ata'), tas_m.group('aid'), 
+                            tas_m.group('atc'), tas_m.group('bpoa'), tas_m.group('epoa'), 
+                            tas_m.group('mac'), tas_m.group('sac'), sum_two, validate_two)
+                        )
```
cmc333333: As we add more validations, lots of `if-append`s will be hard to understand. What do you think about moving these validations to separate functions and then running each in sequence?

cmc333333:
Bump @michaeldhess

dataactcore/config_example.yml
```diff
@@ -52,10 +52,16 @@ broker:
...
-
+    
     # S3 filenames for SF-133 file, only required if planning to load SF-133 table
     sf_133_folder: config
 
+    # The path where the broker will store sf133 files and emails
+    # when the broker is not using an AWS S3 bucket.
+    # This should be an absolute path. Note that this path
+    # is ignored if use_aws is true.
+    sf_133_files: /full/path/to/sf133/files
```
michaeldhess: Where do you see this? The existing import code, if there is no path specified and `use_aws` is false, returns an empty array.

tests/unit/dataactvalidator/test_sf133_content.py
michaeldhess: I switched back and forth when deciding which one would be easier to understand. I'm not sure why I decided on `format()` over `join()`, because looking over it again I agree that `join()` is more readable.

dataactcore/config_example.yml
cmc333333: Per chat, we're going with a subdir in the `CONFIG_BROKER["path"], "dataactvalidator", "config"` path

cmc333333:
Thanks for making these changes @michaeldhess ! Looks good

PR #449, opened 2016-12-08
Db 1443 detached d file generation

nmonga91:
- [ ] Reviewed by backend
- [ ] Merge coordinated with https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/390

PR #454, opened 2016-12-12
Db 1443 detached d file generation (part 2)

nmonga91:
- Cleaning up code by abstracting common code to functions
- Using python logger
- Implemented changes based on feedback on https://github.com/fedspendingtransparency/data-act-broker-backend/pull/449

- [x] merge coordinated with https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/390
- [x] tested on sandbox with https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/398


nmonga91:
[run tests]

nmonga91:
[run tests]

dataactbroker/handlers/fileHandler.py
```diff
@@ -758,13 +732,32 @@ def handleEmptyResponse(self, job, val_job):
...
+            # Create file D API URL with dates and callback URL
+            api_url = FileHandler.get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date)
+
+            logger.debug('Calling Detached D file API => %s', api_url)
+            try:
+                # Check for numFound = 0
+                if not ("numFound='0'" not in self.get_xml_response_content(api_url)):
+                    sess = GlobalDB.db().session
+                    # If the call to the external API wasn't successful, mark the job as failed
+                    job.error_message = "%s data unavailable for the specified date range" % file_type
```
nmonga91: We may need to reword this error if `numFound=0` can happen on any error when calling the external API.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1225,6 +1171,44 @@ def get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date):
...
+            sess.commit()
+            return new_job
+
+        # This will update the reference so no need to return the job, just the upload and timestamped file names
+        job.filename = upload_file_name
+        job.original_filename = timestamped_name
+        job.job_status_id = JOB_STATUS_DICT["running"]
+        sess.commit()
+
+        return upload_file_name, timestamped_name
```
cmc333333: This function returns two different types, which will make working with it confusing. Could it just always return the `job` (either the original or the new one)?

Relatedly, could we remove the `filename`, `original_filename`, `job_status_id` fields from the `Job` construction? Then we could pass both through the same code path:
```python
if job is None:
    job = Job(job_type_id=...)

job.filename = upload_file_name
...
sess.add(job)
sess.commit()

return job
```

We could always grab those fields from the created job.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1225,6 +1171,44 @@ def get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date):
         url = CONFIG_BROKER["".join([file_type_name, "_url"])].format(cgac_code, start_date, end_date, callback)
         return url
 
+    @staticmethod
+    def create_generation_task(job_id):
```
cmc333333: Nice extraction. Note that this could be a straight function (instead of a staticmethod) for further decoupling

dataactbroker/handlers/fileHandler.py
```diff
@@ -758,13 +732,32 @@ def handleEmptyResponse(self, job, val_job):
...
-        # Check for numFound = 0
-        return "numFound='0'" not in self.get_xml_response_content(api_url)
+        file_type = FILE_TYPE_DICT_LETTER[FILE_TYPE_DICT[file_type_name]]
+        task_key = FileHandler.create_generation_task(job.job_id)
+
+        if not self.isLocal:
+            # Create file D API URL with dates and callback URL
+            api_url = FileHandler.get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date)
+
+            logger.debug('Calling Detached D file API => %s', api_url)
```
cmc333333: This isn't necessarily Detached, hey?

dataactbroker/handlers/fileHandler.py
```diff
@@ -758,13 +732,32 @@ def handleEmptyResponse(self, job, val_job):
...
+            try:
+                # Check for numFound = 0
+                if not ("numFound='0'" not in self.get_xml_response_content(api_url)):
```
cmc333333: Should this be
```python
if "numFound='0'" in self.get_xml_response_content(api_url):
```
instead of the double negation?

dataactbroker/handlers/fileHandler.py
```diff
@@ -703,31 +694,14 @@ def add_job_info_for_d_file(self, upload_file_name, timestamped_name, submission
...
 
-        if not self.isLocal:
-            # Create file D API URL with dates and callback URL
-            get_url = FileHandler.get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date)
+        result = self.call_d_file_api(file_type_name, cgac_code, start_date, end_date, job)
 
-            _debug_logger.debug('Calling D file API => %s', get_url)
-            try:
-                if not self.call_d_file_api(get_url):
-                    self.handleEmptyResponse(job, val_job)
```
cmc333333: Do we still need to special case and empty response? If not, can we remove the `handleEmptyResponse` function?

dataactbroker/handlers/fileHandler.py
```diff
@@ -758,13 +732,32 @@ def handleEmptyResponse(self, job, val_job):
...
+                    sess = GlobalDB.db().session
+                    # If the call to the external API wasn't successful, mark the job as failed
+                    job.error_message = "%s data unavailable for the specified date range" % file_type
+                    job.job_status_id = JOB_STATUS_DICT['failed']
+                    sess.commit()
+            except Timeout as e:
+                exc = ResponseException(str(e), StatusCode.CLIENT_ERROR, Timeout)
+                return False, JsonResponse.error(e, exc.status, url="", start="", end="", file_type=file_type)
+        else:
+            self.complete_generation(task_key, file_type)
```
cmc333333: This function returns different types depending on code path: `None` or `(False, error)`. Should it just return the error? If it's `None` we can assume there was no error.

dataactbroker/handlers/fileHandler.py
```diff
@@ -703,31 +694,14 @@ def add_job_info_for_d_file(self, upload_file_name, timestamped_name, submission
...
-                if not self.call_d_file_api(get_url):
-                    self.handleEmptyResponse(job, val_job)
-            except Timeout as e:
-                exc = ResponseException(str(e), StatusCode.CLIENT_ERROR, Timeout)
-                return False, JsonResponse.error(e, exc.status, url="", start="", end="", file_type=file_type)
-        else:
-            self.complete_generation(task.generation_task_key, file_type)
-
-        return True, None
+        return True, None if result is None else result
```
cmc333333: I find this difficult to read -- perhaps switching the return value to the single error would help:
```python
return (not error, error)
```
Though here it'd probably make sense just to keep sending the error up.

dataactbroker/handlers/fileHandler.py
```diff
@@ -831,27 +824,35 @@ def load_d_file(self, url, upload_name, timestamped_name, job_id, isLocal):
...
 
         Returns: tuple of submission ID and file type
 
         """
-        requestDict = RequestDictionary(self.request)
-        if not (requestDict.exists("submission_id") and requestDict.exists("file_type")):
+        request_dict = RequestDictionary.derive(self.request)
+
+        if detached:
+            if not set(['file_type', 'cgac_code', 'start', 'end']).issubset(request_dict.keys()):
```
cmc333333: I like the subset check.

Minor, but there's a set literal syntax:
```python
{'file_type', 'cgac_code', 'start', 'end'}
```

dataactbroker/handlers/fileHandler.py
```diff
@@ -831,27 +824,35 @@ def load_d_file(self, url, upload_name, timestamped_name, job_id, isLocal):
...
+
+        if not set(['file_type', 'submission_id']).issubset(request_dict.keys()):
             raise ResponseException("Generate file route requires submission_id and file_type",
                                     StatusCode.CLIENT_ERROR)
 
-        submission_id = requestDict.getValue("submission_id")
-        file_type = requestDict.getValue("file_type")
+        submission_id = request_dict['submission_id']
+        file_type = request_dict['file_type']
         return submission_id, file_type
```
cmc333333: This function also returns multiple types of values. Can we standardize to a single return type?

dataactbroker/handlers/fileHandler.py
```diff
@@ -890,126 +891,73 @@ def generateFile(self):
...
-            except Timeout as e:
-                exc = ResponseException(str(e), StatusCode.CLIENT_ERROR, Timeout)
-                return False, JsonResponse.error(e, exc.status, url="", start="", end="", file_type=file_type)
-        else:
-            self.complete_generation(task.generation_task_key, file_type)
+        result = self.call_d_file_api(file_type_name, cgac_code, start_date, end_date, new_job)
 
         # Return same response as check generation route
-        return self.check_detached_generation(user_id, file_type)
+        return self.check_detached_generation(new_job.job_id) if result is None else result
```
cmc333333: I don't know if this is too cute, but you could do something like
```python
return result or self.check_detached_generation(new_job.job_id)
```
to make the logic flow a bit easier to read.

dataactbroker/handlers/fileHandler.py
```diff
@@ -890,126 +891,73 @@ def generateFile(self):
...
         sess = GlobalDB.db().session
 
-        # Get the last job created by the specified user (only care about the jobs that have a user id here
-        # since the generation is detached from a submission)
-        uploadJob = sess.query(Job).filter_by(
-            user_id=user_id,
-            file_type_id=FILE_TYPE_DICT_LETTER_ID[file_type],
-            job_type_id=JOB_TYPE_DICT['file_upload']
-        ).order_by(Job.created_at.desc()).first()
+        # We want to user first() here so we can see if the job is None so we can mark
```
cmc333333: Do you want `.one_or_none()` instead?

dataactbroker/handlers/fileHandler.py
```diff
@@ -1225,6 +1171,44 @@ def get_d_file_url(task_key, file_type_name, cgac_code, start_date, end_date):
...
+        task_key = uuid4()
+        task = FileGenerationTask(generation_task_key=task_key, job_id=job_id)
+        sess.add(task)
+        sess.commit()
+        return task.generation_task_key
+
+    def add_generation_job_info(self, file_type_name, job=None, dates=None):
+        # if job is None, that means the info being added is for detached d file generation
+        sess = GlobalDB.db().session
+        user_id = LoginSession.getName(session)
```
cmc333333: This line can be changed to
```python
user_id = g.user.user_id
```
after #448

dataactbroker/handlers/fileHandler.py
cmc333333: I think that's the advice around methods vs staticmethods, but I was suggesting moving it out of the class altogether. Either way, this is much nicer

dataactbroker/handlers/fileHandler.py
nmonga91: It has been removed. Thanks for pointing that out.

dataactbroker/handlers/fileHandler.py
nmonga91: I've updated the return value to only return an exception (if running into a Timeout exception) and then this statement will work and push the error upwards as it did before. It will look cleaner as well.

dataactbroker/handlers/fileHandler.py
nmonga91: Of course. That was a mistake from copying existing code. Thanks for the catch.

dataactbroker/handlers/fileHandler.py
nmonga91: Yes, it was updated to return either the error or `None`

dataactbroker/handlers/fileHandler.py
nmonga91: I thought it best to separate the functions since the were returning different types

cmc333333:
Nice work @nmonga91 and thanks for working through that!

PR #455, opened 2016-12-12
MAX account help page access

nmonga91:
Allowing anyone with a max account the ability to see the help pages by removing the CGAC check and only setting the CGAC if it's available.


cmc333333:
Simple solution, nice work

PR #458, opened 2016-12-12
Account for NAICS_Description field in file F

cmc333333:
Data for this field needed to be pulled from D1, so this adds a subquery to
grab that. It also converts the mapper interface to use a callable (rather
than an object with a `transform` method).

### DIFF
--- a/dataactcore/utils/fileF.py
+++ b/dataactcore/utils/fileF.py
@@ -6,7 +6,7 @@ import iso3166
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.models.fsrs import (
     FSRSGrant, FSRSProcurement, FSRSSubcontract, FSRSSubgrant)
-from dataactcore.models.stagingModels import AwardFinancial
+from dataactcore.models.stagingModels import AwardFinancial, AwardProcurement
 
 
 def _country_name(code):
@@ -34,7 +34,7 @@ class CopyValues():
         self.subgrant_field = subgrant
         self.award_field = award
 
-    def transform(self, models):
+    def __call__(self, models):
         if self.subcontract_field and models.subcontract:
             return getattr(models.subcontract, self.subcontract_field)
         elif self.subgrant_field and models.subgrant:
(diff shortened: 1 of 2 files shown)


klrBAH:
[run tests]

klrBAH:
[run tests]

dataactcore/utils/fileF.py
```diff
@@ -159,18 +153,27 @@ def transform(self, models):
...
+               AwardProcurement.naics == FSRSSubcontract.naics).\
+        order_by(AwardProcurement.award_procurement_id).limit(1).as_scalar()
+    results = sess.query(AwardFinancial, FSRSProcurement, FSRSSubcontract,
+                         naics_subquery).\
         filter(AwardFinancial.submission_id == submission_id).\
         filter(FSRSProcurement.contract_number == AwardFinancial.piid).\
         filter(FSRSSubcontract.parent_id == FSRSProcurement.id)
-    for award, proc, sub in triplets:
-        yield ModelRow(award, proc, sub, None, None)
+    for award, proc, sub, naics_desc in results:
```
alburde1: Why are we combining the filter in the query above and not in this one? Seems like we should pick one method and stick to it (at least when the two filters are so close to one another)

dataactcore/utils/fileF.py
```diff
@@ -191,5 +194,5 @@ def generate_f_rows(submission_id):
                                      submission_grants(submission_id)):
         result = OrderedDict()
         for key, mapper in mappings.items():
-            result[key] = mapper.transform(model_row) or ''
+            result[key] = mapper(model_row) or ''
```
alburde1: What is mappings, what does "mapper" vs "mapper.transform" do, and why did we remove the transform?

cmc333333: I'll add a comment, but a `mapper` is a callable that converts a `model_row` into a CSV cell value. `mappings` is a collection of `mapper`s for each of the CSV columns.

We don't need the `transform` anymore due to changing them `mapper`s to callables (rather than objects with a "transform" method)

dataactcore/utils/fileF.py
cmc333333: Good call

PR #459, opened 2016-12-12
Populate affiliations, filter available agencies

cmc333333:
Builds on #448 ([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1386-2-start-affiliations...db-1386-3-populate-affiliations)) and must be reconciled with #455

This sets multiple affiliations if a user logs in via MAX and begins to use those affiliations when retrieving agency listings.


dataactbroker/handlers/accountHandler.py
```diff
@@ -140,24 +150,16 @@ def max_login(self,session):
...
             ticket = safeDictionary.getValue("ticket")
             service = safeDictionary.getValue('service')
-            parent_group = CONFIG_BROKER['parent_group']
 
             # Call MAX's serviceValidate endpoint and retrieve the response
             max_dict = self.get_max_dict(ticket, service)
 
             if not 'cas:authenticationSuccess' in max_dict['cas:serviceResponse']:
                 raise ValueError("You have failed to login successfully with MAX")
+            cas_attrs = max_dict['cas:serviceResponse']['cas:authenticationSuccess']['cas:attributes']
```
nmonga91: +1 for the cleanup here

dataactbroker/handlers/accountHandler.py
```diff
@@ -867,28 +864,68 @@ def email_users(self, system_email, session):
...
-        ordered_perms = sorted(
-            PERMISSION_MAP.items(), key=lambda pair: pair[1]['order'])
-        for key, permission in ordered_perms:
-            name = permission['name']
-            if key in perms:
-                user.permission_type_id = PERMISSION_TYPE_DICT[name]
-                break
+        affiliations = list(best_affiliation(perms_to_affiliations(perms)))
+
+        if not affiliations:
```
nmonga91: PR https://github.com/fedspendingtransparency/data-act-broker-backend/pull/454 made some changes to allow folks that aren't in any groups whatsoever (so they don't need to be affiliated with an Agency) to access the broker with a null value for `permission_type_id` and `cgac_code`. This wouldn't be supported with this check for `if not affiliations` correct?

dataactbroker/domainRoutes.py
```diff
@@ -1,14 +1,30 @@
...
+        elif g.user.website_admin:
+            cgacs = sess.query(CGAC).all()
+        else:
+            cgac_ids = [affil.cgac_id for affil in g.user.affiliations]
+            cgacs = sess.query(CGAC).filter(CGAC.cgac_code.in_(cgac_ids)).all()
+        agency_list = [
+            {'agency_name': cgac.agencyname, 'cgac_code': cgac.cgac_code}
+            for cgac in cgacs
+        ]
+        return JsonResponse.create(StatusCode.OK,
```
cmc333333: There's certainly no new risk in inlining the logic this way (all the same imports are happening and functions being executed in the same order) -- the bigger question is how do we want to break up the logic. In this case, I can move the logic to a "handler" module, but then the routes module would be rather bare:
```python
from dataactbroker.handlers.domain import list_agencies

def add_domain_routes(app):
    app.route("/v1/list_agencies/")(list_agencies)
EOF
```

dataactbroker/handlers/accountHandler.py
cmc333333: Correct, this'll need to be rewritten to use the `None` strategy from #455

nmonga91:
Looks good, ready to merge once conflicts are resolved

cmc333333:
I'm going to rebase and build from `development` as there've been many changes upstream

cmc333333:
Closing this in favor of different PRs next sprint

PR #460, opened 2016-12-13
Fix CGAC importing

cmc333333:
When we added UserAffiliations, we didn't account for CGACs being dropped from the DB every deploy. This modifies that import script to modify the CGACs in place and to accommodate legitimate CGAC or User deletions by cascading.


dataactcore/migrations/versions/5de499ab5b62_cascade_useraffiliation_deletes.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint('user_affiliation_user_fk', 'user_affiliation', type_='foreignkey')
+    op.drop_constraint('user_affiliation_cgac_fk', 'user_affiliation', type_='foreignkey')
+    op.create_foreign_key('user_affiliation_user_fk', 'user_affiliation', 'users', ['user_id'], ['user_id'], ondelete='CASCADE')
+    op.create_foreign_key('user_affiliation_cgac_fk', 'user_affiliation', 'cgac', ['cgac_id'], ['cgac_id'], ondelete='CASCADE')
```
nmonga91: When a row in `user_affiliation` deleted, with `ondelete=cascade` both the user and the cgac would also be deleted as a result of the cascade (if i'm understanding the direction of the cascading correctly). I can see wanting to delete the users once they have no affiliations (if they come back into the broker they'll just be recreated with null permissions). 

But, do we want the CGAC entry also removed as a result? Wouldn't that CGAC then be unavailable elsewhere as well?

tests/unit/dataactvalidator/test_scripts_loadFile.py
```diff
@@ -0,0 +1,49 @@
+import pandas as pd
+
+from dataactcore.models.domainModels import CGAC
+from dataactvalidator.scripts import loadFile
+from tests.unit.dataactcore.factories.domain import CGACFactory
+
+
+def test_delete_missing_cgacs(database):
```
nmonga91: +1 for testing our load files, moving in a great direction, and fast, with our testing

dataactvalidator/scripts/loadFile.py
```diff
@@ -13,33 +13,53 @@
...
+    to_delete = set(models.keys()) - set(new_data['cgac_code'])
+    sess = GlobalDB.db().session
+    if to_delete:
+        sess.query(CGAC).filter(CGAC.cgac_code.in_(to_delete)).delete(
+            synchronize_session=False)
+    for cgac_code in to_delete:
+        del models[cgac_code]
+
+
+def update_cgacs(models, new_data):
```
nmonga91: Great use of dynamic attribute assignment

dataactcore/migrations/versions/5de499ab5b62_cascade_useraffiliation_deletes.py
cmc333333: I'm pretty sure the cascade works the other way: when a user is deleted, their affiliations disappear and when a CGAC is deleted, the affiliation also disappears.

I'll write a quick test for it, though

PR #463, opened 2016-12-13
Try two at fixing CGAC agency list

cmc333333:
Also adds tests!

### DIFF
--- a/tests/unit/dataactbroker/test_domainRoutes.py
+++ b/tests/unit/dataactbroker/test_domainRoutes.py
@@ -1,0 +1,51 @@
+import json
+from unittest.mock import Mock
+
+from flask import Flask
+import pytest
+
+from dataactbroker import domainRoutes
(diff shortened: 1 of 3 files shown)


tests/unit/dataactbroker/test_domainRoutes.py
```diff
@@ -0,0 +1,51 @@
...
+from tests.unit.dataactcore.factories.domain import CGACFactory
+from tests.unit.dataactcore.factories.user import UserFactory
+
+
+@pytest.fixture
+def domain_app():
+    app = Flask('test-app')
+    domainRoutes.add_domain_routes(app)
+    app.teardown_appcontext(lambda exception: GlobalDB.close())
+    yield app.test_client()
```
cmc333333: Could use dataactbroker.app:createApp here, but there's a lot of configurations there that I didn't want to mess with.

PR #467, opened 2016-12-15
Use affiliations to determine permissions, part 1

cmc333333:
This includes the first half of an effort to replace our permissions checks (which rely on a user's _single_ permission level) to a per-agency check. Easiest to step through changeset by changeset, but this is largely foundation for part 2, which makes use of these functions.

* Read in multiple user affiliations from MAX
* Remove unused parameters
* Add functions to check user permission for a specific agency

### DIFF
--- a/dataactbroker/handlers/accountHandler.py
+++ b/dataactbroker/handlers/accountHandler.py
@@ -467,6 +467,9 @@ def set_max_perms(user, max_group_list):
     or
     {parent-group}-CGAC_SYS to indicate website_admin"""
     prefix = CONFIG_BROKER['parent_group'] + '-CGAC_'
+
+    # Each group name that we care about begins with the prefix, but once we
+    # have that list, we don't need the prefix anymore, so trim it off.
     perms = [group_name[len(prefix):]
              for group_name in max_group_list.split(',')
              if group_name.startswith(prefix)]
--- a/dataactbroker/handlers/accountHandler.py
+++ b/dataactbroker/handlers/accountHandler.py
@@ -1,6 +1,8 @@
+import logging
+from operator import attrgetter
 import re
-import time
 import requests
+import time
 import xmltodict
 
 from dateutil.parser import parse
@@ -21,7 +23,11 @@ from dataactcore.models.jobModels import Submission
 from dataactcore.utils.statusCode import StatusCode
 from dataactcore.interfaces.function_bag import get_email_template, check_correct_password, updateLastLogin
 from dataactcore.config import CONFIG_BROKER
-from dataactcore.models.lookups import USER_STATUS_DICT, PERMISSION_TYPE_DICT, PERMISSION_MAP
+from dataactcore.models.lookups import (
+    PERMISSION_SHORT_DICT, PERMISSION_TYPE_DICT, USER_STATUS_DICT)
+
+
+logger = logging.getLogger(__name__)
 
 
 class AccountHandler:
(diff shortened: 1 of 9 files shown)


dataactbroker/handlers/accountHandler.py
```diff
@@ -21,7 +23,11 @@
 from dataactcore.utils.statusCode import StatusCode
 from dataactcore.interfaces.function_bag import get_email_template, check_correct_password, updateLastLogin
 from dataactcore.config import CONFIG_BROKER
-from dataactcore.models.lookups import USER_STATUS_DICT, PERMISSION_TYPE_DICT, PERMISSION_MAP
+from dataactcore.models.lookups import (
+    PERMISSION_SHORT_DICT, PERMISSION_TYPE_DICT, USER_STATUS_DICT)
```
alburde1: What do these parenthesis do and why do we have to add them?

dataactbroker/handlers/accountHandler.py
```diff
@@ -430,39 +423,67 @@ def email_users(self, system_email, session):
...
+    UserAffiliations and/or website_admin status.
+
+    Permissions are encoded as a comma-separated list of
+    {parent-group}-CGAC_{cgac-code}-PERM_{one-of-R-W-S}
+    or
+    {parent-group}-CGAC_SYS to indicate website_admin"""
+    prefix = CONFIG_BROKER['parent_group'] + '-CGAC_'
+    perms = [group_name[len(prefix):]
+             for group_name in max_group_list.split(',')
+             if group_name.startswith(prefix)]
```
alburde1: what exactly is being set here? I'm having trouble following this perms setting line for some reason.

dataactbroker/handlers/fileHandler.py
```diff
@@ -949,16 +949,15 @@ def check_detached_generation(self, job_id=None):
...
         """ Return information about file generation jobs
 
         Returns:
             Response object with keys status, file_type, url, message.
             If file_type is D1 or D2, also includes start and end.
         """
         sess = GlobalDB.db().session
-        if submission_id is None or file_type is None:
-            submission_id, file_type = self.get_request_params_for_generate()
+        submission_id, file_type = self.get_request_params_for_generate()
```
alburde1: So how does it know which submission this is for? What if someone is generating more than one submission at a time?

dataactbroker/handlers/accountHandler.py
cmc333333: The parens just allow us to use multiple lines. Pep8 has a 79-char line length, so I tend to use that when writing new code.

alburde1: Ok, so it's not like they're doing anything super important that I didn't know about. I was just confused because the function_bag imports aren't changed and they seem longer. I use PyCharm so it just lets me keep going. Doesn't even get to the random vertical line that probably denotes a "page width" of some sort.

dataactbroker/handlers/accountHandler.py
alburde1: ahhh, trimming, that's what's happening in the first line there. Not sure why my brain refused to understand that. Might be nice to have a comment if you don't mind.

dataactbroker/handlers/fileHandler.py
cmc333333: It grabs the `submission_id` from the request params (see `get_request_params_for_generate()`), which is what this function was doing during all of our executions -- turns out, we never called `checkGeneration()` with an argument.

I don't understand the question about multiple submissions. Maybe you're thinking of multiple file types? If so, the browser makes one request per file type.

dataactbroker/handlers/fileHandler.py
```diff
@@ -875,7 +875,7 @@ def generateFile(self):
             return error_response
 
         # Return same response as check generation route
-        return self.checkGeneration(submission_id, file_type)
+        return self.checkGeneration()
```
alburde1: This was a `checkGeneration` with an argument that was removed

PR #468, opened 2016-12-16
Handful of fixups

cmc333333:
* Avoid a DB error killing all future tests
* Delete runTests.py (now that Jenkins doesn't need it)
* Fix logging of config values
* More useful messages every 100 rows
* Remove all of the deprecated loggers (now that the ELK stack understands json logs)


tests/unit/conftest.py
```diff
@@ -43,6 +43,7 @@ def database(full_database_setup):
     db, tables_in_drop_order = full_database_setup
     yield db
     db.session.expire_all()
+    db.Session.rollback()
```
alburde1: Does capitalization matter here? If so, why is this the only one capitalized?

cmc333333: Ah good question, that's definitely worth a comment!

PR #469, opened 2016-12-16
Use affiliations to determine permissions, part 2

cmc333333:
Builds on #467 ([diff](https://github.com/fedspendingtransparency/data-act-broker-backend/compare/db-1386-3-permission-setup...db-1386-4-use-permissions))

This adds goes through each of the `fileHandler` endpoints and switches each over to inspect user affiliations for permissions. A lot of code changes, but it's relatively repetitive after the foundation is laid in the first few commits.

Tested this locally as a `website_admin` and on sandbox as a `writer` (but not admin).

### DIFF
--- a/dataactbroker/README.md
+++ b/dataactbroker/README.md
@@ -22,7 +22,15 @@ dataactbroker/
 The `/dataactbroker/scripts` folder contains the install scripts needed to setup the broker API for a local install. For complete instructions on running your own copy of the API and other DATA Act broker components, please refer to the [documentation in the DATA Act core responsitory](https://github.com/fedspendingtransparency/data-act-broker-backend/blob/master/doc/INSTALL.md "DATA Act broker installation guide").
 
 ### Handlers
-The `dataactbroker/handlers` folder contains the logic to handle requests that are dispatched from the `loginRoutes.py`, `fileRoutes.py`, and 'userRoutes.py' files. Routes defined in these files may include the `@permissions_check` tag to the route definition. This tag adds a wrapper that checks if there exists a session for the current user and if the user is logged in, as well as checking the user's permissions to determine if the user has access to this route. If user is not logged in to the system or does not have access to the route, a 401 HTTP error will be returned. This tag is defined in `dataactbroker/permissions.py`.
+The `dataactbroker/handlers` folder contains the logic to handle requests that
+are dispatched from the `loginRoutes.py`, `fileRoutes.py`, and 'userRoutes.py'
+files. Routes defined in these files may include the `@requires_login` and
+`@requires_submission_perms` tags to the route definition. This tag adds a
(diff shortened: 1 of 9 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -151,18 +151,13 @@ def getErrorReportURLsForSubmission(self, is_warning = False):
             # Unexpected exception, this is a 500 server error
             return JsonResponse.error(e, StatusCode.INTERNAL_ERROR)
 
-    def get_signed_url_for_submission_file(self):
+    def get_signed_url_for_submission_file(self, submission):
```
alburde1: Do we use this anywhere in this function? You took out the `user_agency_must_match` which I think was where the submission was needed.

dataactbroker/README.md
```diff
@@ -22,7 +22,15 @@ dataactbroker/
...
+The `dataactbroker/handlers` folder contains the logic to handle requests that
+are dispatched from the `loginRoutes.py`, `fileRoutes.py`, and 'userRoutes.py'
+files. Routes defined in these files may include the `@requires_login` and
+`@requires_submission_perms` tags to the route definition. This tag adds a
+wrapper that checks if there exists a session for the current user and if the
+user is logged in, as well as checking the user's permissions to determine if
+the user has access to this route. If user is not logged in to the system or
+does not have access to the route, a 401 HTTP error will be returned. This
+tags are defined in `dataactbroker/permissions.py`.
```
cmc333333: Good call

dataactbroker/fileRoutes.py
```diff
@@ -139,15 +141,15 @@ def sign_submission_file():
...
-    @permissions_check(permission='writer')
-    def post_submission_narratives(submission_id):
+    @requires_submission_perms('writer')
+    def post_submission_narratives(submission):
         json = request.json or {}
         # clean input
         json = {key.upper():value.strip() for key, value in json.items()
                 if isinstance(value, str) and value.strip()}
-        return update_narratives(int(submission_id), json)
+        return update_narratives(submission, json)
```
cmc333333: Nah, we were wrapping that function with `@requires_submission_perms` before which did the conversion. We've just bumped that conversion into the routes now.

dataactbroker/handlers/fileHandler.py
cmc333333: Nice catch, I think you've found a security vulnerability -- it looks like a user can request _any_ file.

cmc333333: We've got DB-1538 in the sprint now to address this; salright to leave this until then? We'll need the submission for the file verification.

alburde1: Why did we remove `user_agency_must_match` from this function in the first place? It was doing that exact check for vulnerability.

PR #472, opened 2016-12-16
Hotfix: Removing part of rule B18 that checks if the reimbursable flag is present for 3 digit object class code.

nmonga91:
Hotfix needed due to NRC validation issues


cmc333333:
:+1: @nmonga91 sneaking in next sprint's tickets tsk tsk

PR #477, opened 2016-12-20
Filter by affiliation, remove permission/cgac from user

cmc333333:
Final step in allowing users to have more than one agency:
* Filter submissions and user emails by the current user's affiliations (rather than cgac_code)
* Remove cgac_code and permission_type_id from user (we defer to affiliation values everywhere)

Relies on fedspendingtransparency/data-act-broker-web-app#406


dataactbroker/handlers/fileHandler.py
```diff
@@ -1179,3 +1145,55 @@ def get_error_metrics(submission):
...
+                      Submission.user_id == g.user.user_id))
+    if certified != 'mixed':
+        query = query.filter_by(publishable=certified)
+    submissions = query.order_by(Submission.updated_at.desc()).\
+            limit(limit).offset(offset)
+
+    return JsonResponse.create(StatusCode.OK, {
+        "submissions": [serialize_submission(s) for s in submissions],
+        "total": query.count()
+    })
```
alburde1: I realize we aren't doing this right now, but should we be accounting for website_admin in this filter? Can they see all submissions? If so, it doesn't seem to be reflected here

dataactbroker/scripts/initialize.py
```diff
@@ -1,20 +1,16 @@
-from dataactvalidator.app import createApp
 from dataactbroker.scripts.setupEmails import setupEmails
```
alburde1: This whole file is being murdered in my PR, so don't worry about editing it any more than this. We don't actually use this initialize.

dataactbroker/handlers/accountHandler.py
```diff
@@ -235,19 +235,19 @@ def logout(self,session):
...
-            this_info = {"id":user.user_id, "name": user.name, "email": user.email}
-            user_info.append(this_info)
+        users = sess.query(User).\
+            filter_by(user_status_id=USER_STATUS_DICT['approved']).\
+            filter_by(is_active=True)
+        if not g.user.website_admin:
+            relevant_cgacs = [aff.cgac_id for aff in g.user.affiliations]
+            subquery = sess.query(UserAffiliation.user_id).\
+                filter(UserAffiliation.cgac_id.in_(relevant_cgacs))
+            users = users.filter(User.user_id.in_(subquery))
```
cmc333333: `my_query.all()` is the same thing as `list(my_query)`; it just converts an iterator into a list. We're still doing that materialization in line 247, so we don't need the `.all()` here.

dataactbroker/handlers/fileHandler.py
cmc333333: Good idea, will add

PR #479, opened 2016-12-21

alburde1:
General cleanup of backend code. Do this one commit at a time (although I admit there is more than one commit where I forgot to save the changes to the tests so the commit after it is related directly to the last one)

- Removing a lot of functions we didn't use anymore, either because of the route cleanup or just in general.
- Removing anything to do with locking/expiring an account
- Linting cleanup
- Making a few functions static
- Removing empty/unused files (including yml and initialize files)
- Cleaning up the README a little bit

Next steps include:
- Update models (double-check all of these to make sure we really don't need them anymore)
  - Remove `is_active`, `incorrect_password_attempts`, and `user_status_id` fields from `User` model. Possibly also `last_login_date` field. Uncouple from `UserStatus`
  - Remove `UserStatus` model completely
- Clean up code based on model updates
- Go through and do a test refactor now that we removed so many routes and to reflect models
- Final runthrough for linting/README/etc


dataactbroker/handlers/accountHandler.py
```diff
@@ -168,7 +168,6 @@ def max_login(self,session):
         except Exception as e:
             # Return 500
             return JsonResponse.error(e,StatusCode.INTERNAL_ERROR)
-        return self.response
```
cmc333333: Good catch!

dataactbroker/handlers/aws/session.py
```diff
@@ -197,7 +139,7 @@ def save_session(self, app, session, response):
                 session["session_check"] = False
             else:
                 expiration = datetime.utcnow() + timedelta(seconds=SessionTable.TIME_OUT_LIMIT)
-        if(not "_uid" in session):
+        if not "_uid" in session:
```
cmc333333: Super small, but PEP8 prefers `if x not in y` to `if not x in y`.

dataactbroker/scripts/setupEmails.py
```diff
@@ -11,95 +11,46 @@ def setupEmails():
...
-            "[EMAIL]."
-            "<br /><br />"
-            "The DATA Act Implementation Team <br />"
-            "[EMAIL]")
-        loadEmailTemplate(sess, "DATA Act Broker - Registration", template, "account_creation_user")
-
-        #Unlock account email
-        template = "Your account has been unlocked and requires your password to be reset. Please click the following link <a href='[URL]'>here</a> to start the processs. The link will expire in 24 hours. <br/> <br/> If you did not request your account to be unlocked, please notify the DATA Act Broker Helpdesk (DATABroker@fiscal.treasury.gov) <br /><br />DATA Act Broker Helpdesk<br /><br />DATABroker@fiscal.treasury.gov"
-        loadEmailTemplate(sess, "DATA Act Broker - Password Reset", template, "unlock_account")
-
```
cmc333333: Nice!

PR #482, opened 2016-12-28
Convert D1/2 federal_action_obligation fields to floats

cmc333333:
We were comparing the string "0.0000" to "0", which is silly. Convert everything to floats (which is safe according to production data) instead.


dataactcore/migrations/versions/6affb1acf439_convert_federal_action_obligation_to_.py
```diff
@@ -0,0 +1,58 @@
...
+    op.alter_column('award_procurement', 'federal_action_obligation',
+               existing_type=sa.Numeric(),
+               type_=sa.TEXT(),
+               existing_nullable=True)
+    op.alter_column('award_financial_assistance', 'federal_action_obligation',
+               existing_type=sa.Numeric(),
+               type_=sa.TEXT(),
+               existing_nullable=True)
+    ### end Alembic commands ###
+
```
alburde1: This may be a silly question but why is the downgrade `op.alter_column` while upgrade is `op.execute` with an alter inside it?

cmc333333: Alembic doesn't output the "USING" statement, which is required when moving from string to numeric but not when moving in reverse (numeric to string)

PR #483, opened 2016-12-28
More linting

cmc333333:
Removes 36 exceptions to the flake8 rules. I figured it was a good time to make a bunch of these fixes while most of the devs were out to avoid merge conflicts.

These are largely whitespace changes and were selected due to having only a few exceptions per rule.


tests/integration/mixedFileTests.py
```diff
@@ -210,7 +210,7 @@ def test_program_mixed(self):
         """Test mixed job with some rows failing."""
         jobId = self.jobDict["programMixed"]
         self.passed = self.run_test(
-        jobId, 200, "finished", 8268, 5, "complete", 46, 24, 7697)
+            jobId, 200, "finished", 8268, 5, "complete", 46, 24, 7697)
```
alburde1: why change from camelCase to underscored in `jobTests.py` but not here?

cmc333333: No reason; happy to fix one moment.

alburde1:
Woo! Lots of under_scores :)

PR #485, opened 2016-12-30
Proposal: Add webargs

cmc333333:
We currently have a loose net of custom HTTP argument parsing code which sometimes checks the presence of fields, sometimes checks the field format, etc. etc. In other situations, we assume a field is present or of a particular format and crash if not.

This PR introduces [webargs](https://webargs.readthedocs.io/en/latest/), a somewhat standard HTTP argument parsing library. The expected arguments are defined declaratively and error messages are provided by the library (in addition to type conversions). This PR doesn't rewrite everything, but if we like this approach the path forward is just to remove the remaining uses of `RequestDictionary`


alburde1:
My concern with this method is it removes some of our ability to make specialized messages (as an example, we no longer tell the user that the page number or size limit is the specific problem or what it's supposed to be in `list_submissions`). Do we want to lose that flexibility?

cmc333333:
@alburde1 I _think_ we can make specialized messages, let me test it out. This _does_ send back messages like "job_id: Missing data for required field.", but we could probably make those messages easier to read.

cmc333333:
@alburde1 looks like we can explicitly set the `error_messages` in the `webargs_fields` constructors. However, I'm realizing that none of these messages make it to the end user (the JS UI will send well-formed parameters), so I'm less inclined to swap them all out (the defaults are good enough for developers to understand). Was there anything else?

alburde1:
Nope, but since this is a proposal I'd say a second opinion beyond just mine would be nice. Perhaps @nmonga91 could put in his 2 cents?

nmonga91:
Overall, I agree with this effort to move away from using `RequestDictionary`. Handling these kinds of type and requirement checks with webargs is a great way to extract out trivial checks. And since we can customize the error messages, I'm ok with moving forward with this.

Can we confirm that the HTTP status codes that get returned are still the ones we expect? The front-end I know looks for specific codes for certain errors. For example, permission errors we expected a 401 but changed the way we were raising the exception at some point, and were then getting 403.

cmc333333:
@nmonga91 good thought -- the webargs library actually uses a different error code (422) but we convert it to a 400 before sending it back to the user. I've added a test to verify that conversion.

PR #489, opened 2017-01-04
Check submission permissions when generating report urls

cmc333333:
This replaces the `/v1/sign_submission_file` endpoint, which would load any file name, with `/v1/submission/<int:submission_id>/report_url`, which is more rigid. This new method reduces security concerns but also allows us to change the file url in the future, if desired.

It must be merged concurrently with fedspendingtransparency/data-act-broker-web-app#422


dataactcore/utils/report.py
```diff
@@ -9,18 +9,32 @@ def get_report_path(job, report_type):
     Return the filename for the error report.
     Does not include the folder to avoid conflicting with the S3 getSignedUrl method.
     """
-    path = 'submission_{}_{}_{}_report.csv'.format(job.submission_id, job.file_type.name, report_type)
-    return path
+    return report_file_name(job.submission_id, report_type == 'warning',
```
nmonga91: If we're just forwarding the data to `report_file_name`, do still need this encompassing function as opposed to calling `report_file_name` directly?

cmc333333: Can do -- I didn't want to worry too much about the rest of the app, but if there aren't too many calls to these other functions, I'll replace them.

dataactcore/utils/report.py
```diff
@@ -4,25 +4,6 @@
...
-    """Return filename for a cross file error report."""
-    return report_file_name(submission_id, False, source_file, target_file)
-
-
-def get_cross_warning_report_name(submission_id, source_file, target_file):
-    """Return filename for a cross file warning report."""
-    return report_file_name(submission_id, True, source_file, target_file)
-
-
 def report_file_name(submission_id, warning, file_type, cross_type=None):
```
nmonga91: here `file_type` is equivalent to `source_file` and `cross_type` is equivalent to `target_file`, correct?

cmc333333: Exactly; I renamed to use the same interface for normal file reports and cross file reports.

PR #491, opened 2017-01-04

alburde1:
Cleaning up tests, especially baseTestAPI.py


tests/integration/baseTestAPI.py
```diff
@@ -78,79 +69,36 @@ def setUpClass(cls):
...
-            add_status_user('waiting@agency.gov', 'awaiting_approval')
-            add_status_user('impatient@agency.gov', 'awaiting_approval')
-            add_status_user('watchingPaintDry@agency.gov', 'awaiting_approval')
-            add_status_user(test_users['admin_email'], 'approved', True)
-            add_status_user(test_users['approved_email'], 'approved')
-            add_status_user('nefarious@agency.gov', 'denied')
+                )
+                user.salt, user.password_hash = getPasswordHash(user_password, Bcrypt())
+                user.name = name
+                user.username = username
```
cmc333333: Small detail, but would it make sense to set these two attributes in the `UserFactory` constructor?

tests/integration/baseTestAPI.py
```diff
@@ -173,85 +121,30 @@ def tearDownClass(cls):
...
-        try:
-            self.session_id = response.headers["x-session-id"]
-        except KeyError:
-            # Session ID doesn't come back for inactive user, set to empty
-            self.session_id = ""
-        return response
-
-    def login_password_locked_user(self):
-        """Force user to have their account locked then attempt to login again"""
+    def login_extra_user(self):
```
cmc333333: Looks like login_user and login_extra_user are almost identical -- should the username just be a method parameter?

Since `login_extra_user` is only called once, we could default that param to the "agency_user".

cmc333333:
Whoot whoot, nice work @alburde1 !

PR #492, opened 2017-01-05
Hotfix: No longer marking job as failed if there are no results when calling …

nmonga91:
…the D file API. Marking job, and dependent validation job, as finished and proceeding without errors. No error message will be displayed for all types of D file generation if there are not results from the API.

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -21,11 +21,11 @@ from dataactcore.config import CONFIG_BROKER, CONFIG_SERVICES
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.models.errorModels import File
 from dataactcore.models.jobModels import (
-    FileGenerationTask, Job, Submission, SubmissionNarrative)
+    FileGenerationTask, Job, Submission, SubmissionNarrative, JobDependency)
 from dataactcore.models.userModel import User
 from dataactcore.models.lookups import (
     FILE_TYPE_DICT, FILE_TYPE_DICT_LETTER, FILE_TYPE_DICT_LETTER_ID,
-    JOB_STATUS_DICT, JOB_TYPE_DICT, RULE_SEVERITY_DICT, FILE_TYPE_DICT_ID, JOB_STATUS_DICT_ID)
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -726,9 +726,30 @@ def call_d_file_api(self, file_type_name, cgac_code, start_date, end_date, job):
                 # Check for numFound = 0
                 if "numFound='0'" in self.get_xml_response_content(api_url):
                     sess = GlobalDB.db().session
-                    # If the call to the external API wasn't successful, mark the job as failed
-                    job.error_message = "%s data unavailable for the specified date range" % file_type
-                    job.job_status_id = JOB_STATUS_DICT['failed']
+                    # No results found, skip validation and mark as finished
+                    sess.query(JobDependency). \
+                        filter(JobDependency.prerequisite_id == job.job_id). \
+                        delete(synchronize_session='fetch')
```
cmc333333: Why are we deleting jobs that depend on this one?

dataactbroker/handlers/fileHandler.py
```diff
@@ -726,9 +726,30 @@ def call_d_file_api(self, file_type_name, cgac_code, start_date, end_date, job):
...
+                    mark_job_status(job.job_id, "finished")
+                    job.filename = None
+
+                    val_job = sess.query(Job).filter_by(
+                        submission_id=job.submission_id,
+                        file_type_id=FILE_TYPE_DICT[file_type_name],
+                        job_type_id=JOB_TYPE_DICT['csv_record_validation']
+                    ).one_or_none()
```
cmc333333: Looks like we've already found the validation job in `add_job_info_for_d_file` -- would it make sense to just pass that in rather than re-finding it?

dataactbroker/handlers/fileHandler.py
nmonga91: This was there previously as well to ensure that these empty D file didn't have any pending dependencies that would halt the validation process.

dataactbroker/handlers/fileHandler.py
nmonga91: Added an optional param to account for the detached generator's usage of the function.

cmc333333:
This makes sense, has been manually tested, and isn't very much code. Let's do it.

PR #494, opened 2017-01-05
Hotfix: Handling empty response for d file generation by not displaying an er…

nmonga91:
…ror, but instead allowing a user to continue


cmc333333:
Take 2

PR #498, opened 2017-01-06
Speed up setting tas_id on staging models

cmc333333:
We were still using "is not distinct from" when setting the `tas_id` of staging models. This can't make use of `tas_lookup`'s indexes, so swap it out for the same logic used in SF133 imports.

Also standardizes the TAS component field names (tas_lookup was an outlier).


cmc333333:
@alburde1 - not a super scientific test, but I ran a rather large group of files which took ~551s without this change and ~459s with it (~17% speedup). These were pretty large files, though.

PR #499, opened 2017-01-06

alburde1:
This is another one that looks like a lot of changes but really isn't.

Just a bunch of linting fixes, then removing ignores for the linter and fixing the errors that pop up. Probably easier to go one at a time because each commit is its own rule


dataactvalidator/scripts/loadTas.py
```diff
@@ -136,8 +136,11 @@ def add_existing_id(data):
 
 
 def existing_id(row, existing):
-    """Check for a TASLookup which matches this `row` in the `existing` data.
-    :param existing: Dict[account_num, List[TASLookup]]"""
```
cmc333333: No biggie, but heads up that `:param` is a [sphinx](https://en.wikipedia.org/wiki/Sphinx_(documentation_generator)). No need to keep it here, but that's where the format comes from.

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -5,12 +5,12 @@
...
     """Define some data-munging functions that can be applied to pandas
     dataframes as necessary"""
-    return str(field).lower().strip().replace(" ","_").replace(",","_")
+    return str(field).lower().strip().replace(" ", "_").replace(",", "_")
 
 
 class LoaderUtils:
     @classmethod
-    def padFunction(self, field, padTo, keepNull):
+    def padFunction(cls, field, padTo, keepNull):
```
cmc333333: I think you said you'll do another round in a bit, but if we're not using the `cls` argument, we can turn this into a `staticmethod` (or just move it out of the class altogether)

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -10,7 +10,7 @@ def clean_col_names(field):
...
```
nmonga91: do we even need `cls` here since it's not referenced at all? This could instead be a static method, no?

tests/unit/conftest.py
```diff
@@ -6,9 +6,7 @@
 
 import dataactcore.config
 # Load all models so we can access them through baseModel.Base.metadata
-from dataactcore.models import (    # noqa
```
cmc333333: If this isn't needed anymore, we should probably remove the comment immediately proceeding.

tests/integration/baseTestValidator.py
```diff
@@ -104,7 +104,6 @@ def run_test(self, jobId, statusId, statusName, fileSize, stagingRows,
             stagingRows: Expected number of rows in validation db staging tables. False if no rows are expected
             errorStatus: Expected status in file table of error DB, False if file object should not exist
             numErrors: Expected number of errors
-            rowErrorsPresent: Checks flag for whether row errors occurred, None to skip the check
```
nmonga91: these kinds of docstrings are being removed because they are no longer provided as part of the input to the function call correct?

If we're removing the ones that aren't used, shouldn't we also add the ones that aren't present? For example, `numWarnings` and `warningFileSize`

dataactvalidator/scripts/loaderUtils.py
alburde1: Yeah, I was going to do all the moving of functions and fixing function names (under_score instead of camelCase) in the next round :)

tests/unit/conftest.py
alburde1: Oops... my bad :p

dataactvalidator/scripts/loaderUtils.py
alburde1: I know CM already said it, but I'll reply here too. I planned to do a bunch of static method moving in the next round, I just wanted it to not get too big.

tests/integration/baseTestValidator.py
alburde1: Hmm... weird, I was just using PyCharm to tell me when things were missing/extra. I wonder why it didn't tell me that those two needed to be added.

nmonga91:
Looks good. once @cmc333333 also confirms, we can merge

cmc333333:
Let's do it!

PR #500, opened 2017-01-06
Update report_url API docs

cmc333333:
I missed this in #489

### DIFF
--- a/dataactbroker/README.md
+++ b/dataactbroker/README.md
@@ -600,15 +600,16 @@ permissions (write access for the agency of the submission or SYS).
 {}
 ```
 
-#### POST "/v1/sign_submission_file/"
-This route sends a request to the backend with the submission ID and file name that is desired. The backend generates a signed S3 URL or the path to the local location of the file and returns a JSON response containing that url.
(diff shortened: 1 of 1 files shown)


dataactbroker/README.md
```diff
@@ -621,9 +622,15 @@ This route sends a request to the backend with the submission ID and file name t
...
-  * file - a string indicating the file to generate (file name without extension)
-  * submission - the integer submission ID
-  
+  * warning - Whether or not the requested report is a warning (or error)
+    report. Defaults to False if this parameter isn't present.
+  * file\_type - One of 'appropriations', 'program\_activity',
+    'award\_financial', 'award', 'award\_procurement', 'awardee\_attributes'
+    or 'sub\_award'. Designates the type of report you're seeking.
+  * cross\_type - If present, indicates that we're looking for a
+    cross-validation report between `file_type` and this parameter. It accepts the
```
alburde1: I don't think you have to escape out the `_`, check and see? No need to add those extra `\` if you don't.

cmc333333: Good call. Also looks like I do need to escape those angle brackets, though.

PR #502, opened 2017-01-09
Bug fix for B9/10 Validation for File C only

mab6bf:
Updating B9/10 for File C to exclude those rows with both a null program activity code and null program activity name, updated tests

### DIFF
--- a/tests/unit/dataactvalidator/test_b9_award_financial.py
+++ b/tests/unit/dataactvalidator/test_b9_award_financial.py
@@ -78,4 +78,4 @@ def test_failure_program_activity_null(database):
     pa = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
                                 account_number='test')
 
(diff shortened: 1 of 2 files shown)


tests/unit/dataactvalidator/test_b9_award_financial.py
```diff
@@ -78,4 +78,4 @@ def test_failure_program_activity_null(database):
     pa = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
                                 account_number='test')
 
-    assert number_of_errors(_FILE, database, models=[af, pa]) == 1
+    assert number_of_errors(_FILE, database, models=[af, pa]) == 0
```
bsweger: Very minor nit: since this is not longer testing for a failure, what do you think about renaming the function to `test_success_null_program_activity` or something like that?

mab6bf: Good call

PR #503, opened 2017-01-10
Add back all of the model types

cmc333333:
Alembic needs to know about all of the app's models, which means we need to
import them. This line was taken out during a recent refactor, but we still
need it until we can load those models dynamically.

This fixes alembic migration generation; it _may_ be required to upgrade the database.


alburde1:
My bad, this is good

PR #509, opened 2017-01-23
Db 1484 job manager sqs

nmonga91:
Transitioning the job manager from Celery/RabbitMQ to SQS.

- [x] SQS queue config and polling implementation verified & approved by @mab6bf
- [ ] Code reviewed by backend developer


nmonga91:
[run tests]

nmonga91:
[run tests]

nmonga91:
[run tests]

nmonga91:
[run tests]

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -416,7 +416,7 @@ def run_cross_validation(self, job):
         # Mark validation complete
         mark_file_complete(job_id)
 
-    def validate_job(self, request):
+    def validate_job(self, job_id):
```
alburde1: Are we calling this anywhere else? Does this mess with that in any way?

dataactvalidator/app.py
```diff
@@ -1,111 +1,59 @@
...
-                # insert file-level error info to the database
-                write_file_error(job.job_id, job.filename, error.errorType, error.extraInfo)
-            if error.errorType != ValidationError.jobError:
-                # job pass prerequisites for validation, but an error
-                # happened somewhere. mark job as 'invalid'
-                mark_job_status(job.job_id, 'invalid')
-        return JsonResponse.error(error, error.status)
-
-    @flask_app.errorhandler(Exception)
-    def handle_validation_exception(error):
```
alburde1: How are we handling validation exceptions now? This was catching all of our exceptions and sending them off to the app, where do they get caught now when they're just raised?

dataactvalidator/app.py
```diff
@@ -1,111 +1,59 @@
...
-    @flask_app.teardown_appcontext
-    def teardown_appcontext(exception):
-        GlobalDB.close()
-
-    @flask_app.before_request
-    def before_request():
-        GlobalDB.db()
-
-    @flask_app.errorhandler(ResponseException)
-    def handle_response_exception(error):
```
alburde1: How do we handle response exceptions that simply get raised?

dataactbroker/app.py
```diff
@@ -63,6 +63,8 @@ def teardown_appcontext(exception):
 
     @flask_app.before_request
     def before_request():
+        # Set global value for local
+        g.is_local = local
```
alburde1: if this is global now, can we stop passing it around or setting it in an init and just use the global value?

dataactcore/interfaces/function_bag.py
```diff
@@ -334,13 +336,19 @@ def check_job_dependencies(job_id):
...
                 mark_job_status(dep_job_id, 'ready')
-                # add dep_job_id to the SQS job queue
 
-                logger.info('Sending job %s to job manager in sqs', dep_job_id)
-                sqs = boto3.resource('sqs', region_name=CONFIG_BROKER['aws_region'])
-                queue = sqs.get_queue_by_name(QueueName='job-manager')
-                response = queue.send_message(MessageBody=str(dep_job_id))
-                logger.info('Send message response: %s', response)
+                # If local, utilize local db to mock queue
+                if g.is_local:
```
alburde1: Why are we not logging anything if it's local? I would think we still need some indication just in case something goes wrong.

dataactvalidator/validation_handlers/validationManager.py
nmonga91: only in the integration tests, but that's currently being worked on.

dataactvalidator/app.py
nmonga91: Any exception that gets raised will be logged, but we don't get an HTTP response any more so we don't need to explicitly handle response exceptions

dataactvalidator/app.py
nmonga91: After speaking with CM, we concluded it's ok to let python handle the exceptions since they don't get to the user anyways. The logger will still have them but we shouldn't need to handle it explicitly anymore since we don't have an endpoint to deal with.

nmonga91:
1. will we have to change how we create jobs in other functions or does the general process stay the same and it's just the backend stuff that changes?

- The process for creating jobs, etc will stay the same since, yes, this is just backend changes. The way we're communicating with the validator changes a bit since we're pushing to SQS and the validator is now just polling instead of publishing an endpoint.

2. Are more tests being added (even basic ones) to test this SQS stuff since the final commit tore out a lot of the tests?

- yup! currently working on this

dataactvalidator/app.py
```diff
@@ -1,111 +1,73 @@
...
-from dataactcore.utils.responseException import ResponseException
-from dataactvalidator.validation_handlers.validationError import ValidationError
 from dataactvalidator.validation_handlers.validationManager import ValidationManager
+from dataactcore.models.jobModels import SQS
+from dataactcore.aws.sqsHandler import get_queue
 
 
 logger = logging.getLogger(__name__)
+logging.getLogger('boto3').setLevel(logging.CRITICAL)
+logging.getLogger('botocore').setLevel(logging.CRITICAL)
```
cmc333333: Probably want to put this in the config file (or add it to `dataactcore.logging`). Just be sure that that `configure_logging` is called somewhere.

dataactvalidator/app.py
```diff
@@ -1,111 +1,73 @@
...
+
+        @current_app.before_request
+        def before_request():
+            GlobalDB.db()
+
+        # If local, utilize local db to mock queue
+        if local:
+            sess = GlobalDB.db().session
+
+            while 1:
```
cmc333333: I'd bet that hitting the local database like this will have less-than-ideal consequences. What if you added a 10 second sleep (to match the SQS bits below) instead?

Minor: `while True:` is a smidge more Pythonic.

dataactvalidator/config/supervisord.conf
```diff
@@ -1,17 +1,18 @@
...
 autostart=true
 autorestart=true
 stopasgroup=true
 killasgroup=true
+numprocs=13
 directory=/data-act/backend/dataactvalidator/
 startsecs=1
 startretries=10
-user=ec2-user
+user=root
```
cmc333333: This doesn't seem right -- is there a way to run as ec2-user instead? I think that's how all of the ansible stuff is working.

dataactvalidator/config/supervisord.conf
```diff
@@ -20,6 +21,8 @@ directory=/data-act/backend/dataactvalidator/
 startsecs=1
 startretries=10
 user=root
+environment=PYTHONPATH=%(ENV_PATH)s:/data-act/backend
```
cmc333333: Is the `s` supposed to be here?

dataactcore/interfaces/function_bag.py
```diff
@@ -332,11 +335,18 @@ def check_job_dependencies(job_id):
...
-                logger.info('Sending job %s to job manager', dep_job_id)
-                # will move this later
-                from dataactcore.utils.jobQueue import enqueue
-                enqueue.delay(dep_job_id)
+
+                # If local, utilize local db to mock queue
+                if g.is_local:
+                    new_queue_entry = SQS(job_id=dep_job_id)
+                    sess.add(new_queue_entry)
+                    sess.commit()
```
cmc333333: One way to avoid these logic branches here is to create an interface that matches SQS's. Something like:
```python
class SQSMockQueue:
    @staticmethod
    def send_message(MessageBody):    # noqa
        sess = GlobalDB.db().session
        sess.add(SQS(job_id=MessageBody))
        sess.commit()

    @staticmethod
    def receive_messages(WaitTimeSeconds):  # noqa
        for sqs in sess.query(SQS):
            yield SQSMockMessage(sqs)

class SQSMockMessage:
    def __init__(self, sqs):
        self.sqs = sqs
        self.body = sqs.job_id

    def delete(self):
        sess = GlobalDB.db().session
        sess.delete(self.sqs)
        sess.commit()


def sqs_queue():
    if CONFIG_SOMETHING['local']:
        return SQSMockQueue
    else:
        # stuff that's in get_queue
        return queue
```

dataactvalidator/app.py
cmc333333: One note here -- an exception will go all the way up to the Python run time. You've got supervisor set up, though, so those processes _should_ get restarted.

dataactvalidator/config/supervisord.conf
nmonga91: I had issues trying to start it while it was set to `ec2-user` but I can try to set it to that again

dataactvalidator/config/supervisord.conf
nmonga91: The other configs had it so I had not messed with out of fear that it may break something xD

dataactcore/interfaces/function_bag.py
nmonga91: Excellent suggestion! will definitely incorporate

dataactvalidator/config/supervisord.conf
cmc333333: Ooooh, this is Python templating, not sh templating! Never mind my comment, then!

dataactvalidator/app.py
```diff
@@ -1,111 +1,73 @@
...
+        def teardown_appcontext(exception):
+            GlobalDB.close()
+
+        @current_app.before_request
+        def before_request():
+            GlobalDB.db()
+
+        # If local, utilize local db to mock queue
+        if local:
+            sess = GlobalDB.db().session
```
alburde1: One of my comments somehow managed to not get added (or if it did even I can't find it).... I meant to also ask (here and everywhere else that we do SQS stuff differently for local vs remote) shouldn't we be logging locally too? I think it may help if something goes wrong locally, at least just saying that SQS polling is starting or something.

nmonga91:
Going to squash this into one commit so it's easy to revert back to the job manager (just in case)

PR #511, opened 2017-01-24
Attempting to improve performance of B9

mab6bf:


mab6bf:
Hey Becky, I'm working with Nipun on an alternative way to perform this query due to the performance hit we're taking from the main JOIN on this, so I'm going to close this PR

PR #512, opened 2017-01-24
Adding SQS integration

nmonga91:
Squashing PR 509's commits into one as a precaution in case of need to revert

- [x] SQS config/implementation approved by @mab6bf 
- [x] Reviewed by backend


alburde1:
Second question I just realized. Does this affect any of our docs in any way? Is there anything we need to update about what to run? I know this boots celery out, do we need to document that anywhere to avoid confusion? Is there something that takes its place that needs to be running? All important things to document especially as new people are onboarded

nmonga91:
I left the test files in intentionally in case we want to incorporate a full submission/validation using those. We can clean those up if we decide otherwise or remove whichever ones don't get utilized.

This will affect our docs and I'll update them.

PR #513, opened 2017-01-25
Only validate program activity for the years that we have it

kkuratsu:


klrBAH:

nmonga91:
@kkuratsu Can you also add a test where the AwardFinancial or ObjectClassProgramActivity objects have a year that doesn't match the ProgramActivity object? It should then pass if it tries to validate against a year we don't have, which is the main thing we'd want to check with this update.

kkuratsu:
I will work on this.

kkuratsu:
Don't think we need line #10.

`AND CAST(COALESCE(af.beginning_period_of_availa,'0') AS integer) IN (SELECT DISTINCT CAST(budget_year AS integer) FROM program_activity)
`

dataactvalidator/config/sqlrules/b9_award_financial.sql
```diff
@@ -7,7 +7,6 @@ SELECT af.row_number,
 	af.program_activity_code
 FROM award_financial as af
 WHERE af.submission_id = {0}
-	AND CAST(COALESCE(af.beginning_period_of_availa,'0') AS integer) >= 2016
```
nmonga91: If we remove this we're not making sure we're only checking against years we have. The change that was originally here should've been kept. Why was that removed?

dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
```diff
@@ -7,7 +7,6 @@ SELECT op.row_number,
 	op.program_activity_code
 FROM object_class_program_activity as op
 WHERE op.submission_id = {0}
-	AND CAST(COALESCE(op.beginning_period_of_availa,'0') AS integer) >= 2016
```
nmonga91: Same question as in award financial's b9 sql file

nmonga91:
I think we could keep those lines in order to short circuit the check instead of waiting to check the year once a join has already occurred. Would this negatively impact the performance?

Short circuiting might not save too much for File B, but for File C the dataset can be pretty large and that query is already pretty slow to begin with.

nmonga91:
Looks good, merging!

PR #514, opened 2017-01-25
Db 1487 upload detached files

alburde1:
Uploading FABS file. Creates new models and routes for this process.

- [x] Merged before or at the same time as https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/429


dataactbroker/fileRoutes.py
```diff
@@ -136,6 +136,11 @@ def complete_generation(generation_id):
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
         return file_manager.complete_generation(generation_id)
 
+    @app.route("/v1/upload_detached_file/", methods=["POST"])
+    def upload_detached_file():
```
nmonga91: Are there POST parameters we expect as part of this route?

dataactbroker/handlers/fileHandler.py
```diff
@@ -642,6 +643,130 @@ def generate_detached_file(self, file_type, cgac_code, start, end):
...
+                        upload_name=upload_name,
+                        file_name=file_name,
+                        file_letter=FILE_TYPE_DICT_LETTER[FILE_TYPE_DICT[file_type]]
+                    ))
+
+            file_job_dict = create_jobs(upload_files, submission, False)
+            for file_type in file_job_dict.keys():
+                if "submission_id" not in file_type:
+                    response_dict[file_type+"_id"] = file_job_dict[file_type]
+            if create_credentials and not self.isLocal:
```
nmonga91: This response is the same as `submit` and so is the build fileNameMap stuff, right? Can we abstract it out to a method call if possible instead of duplicating the same code?

dataactbroker/handlers/fileHandler.py
```diff
@@ -1015,10 +1140,10 @@ def list_submissions(page, limit, certified):
     offset = limit * (page - 1)
 
     cgac_codes = [aff.cgac.cgac_code for aff in g.user.affiliations]
-    query = sess.query(Submission)
+    query = sess.query(Submission).filter_by(d2_submission=False)
```
nmonga91: +1

dataactbroker/handlers/fileHandler.py
```diff
@@ -1015,10 +1140,10 @@ def list_submissions(page, limit, certified):
...
     if not g.user.website_admin:
         query = query.filter(sa.or_(Submission.cgac_code.in_(cgac_codes),
-                                    Submission.user_id == g.user.user_id))
+                                    Submission.user_id == g.user.user_id),)
```
nmonga91: What's the extra comma here used for?

dataactcore/migrations/versions/4bf29ae16467_.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
```
nmonga91: This was used to merge 2 revisions i'm assuming. Can we still put a meaningful message/file name? Maybe something like `[version #]_merging_[branch 1 version #]_and_[branch 2 version #]`?

dataactcore/migrations/versions/98e33bf8ff23_.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
```
nmonga91: Same thing here as before

dataactcore/migrations/versions/fc6be41471a3_populate_submission_reporting_period_.py
```diff
@@ -12,56 +12,16 @@
 branch_labels = None
 depends_on = None
 
-from alembic import op
```
nmonga91: Is this revision still necessary even though it's just passing? Would it be problematic to remove it altogether?

dataactvalidator/scripts/loadFile.py
```diff
@@ -36,20 +36,20 @@ def update_cgacs(models, new_data):
...
 
         # read CGAC values from csv
-        data = pd.read_csv(filename, dtype=str)
+        data = pd.read_csv(file_name, dtype=str)
         # clean data
         data = clean_data(
             data,
             CGAC,
-            {"cgac": "cgac_code", "agency": "agency_name"},
+            {"cgac_agency_code": "cgac_code", "agency_name": "agency_name"},
```
nmonga91: This rename is to match the agency list csv we're getting from S3?

dataactbroker/handlers/fileHandler.py
alburde1: oops... I must have missed it

dataactcore/migrations/versions/fc6be41471a3_populate_submission_reporting_period_.py
alburde1: I talked about it with CM and he thinks it's better to just have it there rather than alter the history/order

dataactvalidator/scripts/loadFile.py
alburde1: Hess did this one, but I think yes.

dataactbroker/fileRoutes.py
alburde1: Yes, agency code, quarter, reporting period, etc

nmonga91: Should this route also incorporate the argument checks like below, using `@use_kwargs`?

alburde1: I was debating updating it to use them but I never quite did it. Do you think we should do that or should that be for the next ticket?

nmonga91: I think it's worthwhile to follow the standard put in place since it can used simply for assessing which fields are required and what their type should be. We probably don't need the `validate` part though

nmonga91:
Ready to merge, but will be merged _before_ the front-end component.

PR #525, opened 2017-02-01
Updating integration test for SQS

nmonga91:
Updating SQS integration test to not purge since we're limited to 1 purge per minute

### DIFF
--- a/tests/integration/sqsTests.py
+++ b/tests/integration/sqsTests.py
@@ -14,7 +14,3 @@ class SQSTests(BaseTestValidator):
         self.assertNotEqual(messages, [])
         for message in messages:
             message.delete()
-
-        queue.purge()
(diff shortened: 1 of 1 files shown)


nmonga91:
So instead of purging, we set the retention period to 1 minute for all messages. They'll go away on their own and the focus of this initial test is to just test the connection to the queue.

alburde1:
Does this setting affect the regular queue and is this a problem? In case a message gets hung up for longer than a minute for any reason in the regular process.

nmonga91:
These tests have their own PR queue and since it's just adding then deleting the message there shouldn't be any hangup. If there is, there's a 30 second visibility timeout and once that expires, it just goes back on the queue and won't be used anymore. And then since the retention is 1 minute, it'll go away on its own.

PR #526, opened 2017-02-01
Db 1614 d2 fain validation rules

alburde1:
Creating fain validation rules for FABS


dataactvalidator/config/sqlrules/d1_detached_award_financial_assistance.sql
```diff
@@ -0,0 +1,8 @@
+SELECT
+    row_number,
+    record_type,
+    fain
+FROM detached_award_financial_assistance as dafa
+WHERE dafa.submission_id = {0}
+AND dafa.record_type = 2
```
nmonga91: We should probably indent the lines with `AND` here to keep formatting consistent and easy to view

tests/unit/dataactcore/factories/staging.py
```diff
@@ -218,6 +218,59 @@ class Meta:
...
+class DetachedAwardFinancialAssistanceFactory(factory.Factory):
+    class Meta:
+        model = stagingModels.DetachedAwardFinancialAssistance
+
+    detached_award_financial_assistance_id = None
+    submission_id = fuzzy.FuzzyInteger(9999)
+    job_id = fuzzy.FuzzyInteger(9999)
+    row_number = fuzzy.FuzzyInteger(9999)
+    action_date = fuzzy.FuzzyText()
+    action_type = fuzzy.FuzzyText()
```
nmonga91: Do we want to add default values for `action_type` also? The acceptable values should be [A, B, C, D, blank] but not sure if we need to get that deep for the default values for these.

Asked because I saw `record_type` was limited to 1 or 2.

alburde1: I wasn't sure what the default values for this were. We can add them but do we need to?

dataactvalidator/config/sqlrules/d1_detached_award_financial_assistance.sql
alburde1: I was actually thinking about it but in other sql rules the AND statements for the WHERE aren't indented (see a1, a2, etc). Do we want to start doing it?

nmonga91: I think we should. It definitely helps with readability

tests/unit/dataactcore/factories/staging.py
nmonga91: I don't think we have that for the reimbursable fund columns that can only be 'D', 'R', or blank so we should be fine

nmonga91:
Looks good, merging!

PR #527, opened 2017-02-01
Adding agency column to the recent activity data query

kimminahm:
This must be changed before the web app changes. Confirm that the new request for the recent activity table contains an agency field

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1059,7 +1059,6 @@ def serialize_submission(submission):
         "last_modified": submission.updated_at.strftime('%Y-%m-%d'),
         "size": total_size,
         "status": status,
-        # "errors": submission.number_of_errors,
         "agency": agency_name,
         # @todo why are these a different format?
         "reporting_start_date": str(submission.reporting_start_date),
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1049,17 +1049,13 @@ def serialize_submission(submission):
 
     cgac = sess.query(CGAC).\
         filter_by(cgac_code=submission.cgac_code).one_or_none()
-    if cgac:
-        agency_name = cgac.agency_name
-    else:
-        agency_name = 'N/A'
 
     return {
         "submission_id": submission.submission_id,
         "last_modified": submission.updated_at.strftime('%Y-%m-%d'),
         "size": total_size,
         "status": status,
-        "agency": agency_name,
+        "agency": cgac.agency_name if cgac else 'N/A',
         # @todo why are these a different format?
         "reporting_start_date": str(submission.reporting_start_date),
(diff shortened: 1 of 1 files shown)


klrBAH:

dataactbroker/handlers/fileHandler.py
```diff
@@ -1046,12 +1046,21 @@ def serialize_submission(submission):
...
+        agency_name = ''
+
     return {
         "submission_id": submission.submission_id,
         "last_modified": submission.updated_at.strftime('%Y-%m-%d'),
         "size": total_size,
         "status": status,
-        "errors": submission.number_of_errors,
+        # "errors": submission.number_of_errors,
+        "agency": agency_name,
```
nmonga91: We can shorten the above if/else block here to be `"agency": cgac.agency_name if cgac.agency_name else ''`. Also, I'm not sure we want to display just a blank value if there's no agency associated with a submission (although that shouldn't be the case for a submission, since it _should_ always have a submission), but we may want to return `None` instead of `''`  which will translate to `null` in the response JSON. 

Either way, that value needs to probably be translated to something more user friendly. I'd refer to Kaitlin on what they'd like to see in that scenario.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1046,12 +1046,20 @@ def serialize_submission(submission):
...
+        agency_name = cgac.agency_name
+    else:
+        agency_name = ''
+
     return {
         "submission_id": submission.submission_id,
         "last_modified": submission.updated_at.strftime('%Y-%m-%d'),
         "size": total_size,
         "status": status,
-        "errors": submission.number_of_errors,
```
nmonga91: Is there a reason this being removed?

kimminahm: Talked with Kaitlin and added screenshots of the table with agency added. The table was crowded and hard to read so we discussed it and she suggested removing the number of errors. This gave the rest of the table space to become readable.
screen shot 2017-02-01 at 11 55 37 am
screen shot 2017-02-01 at 2 09 08 pm

dataactbroker/handlers/fileHandler.py
```diff
@@ -1046,12 +1046,20 @@ def serialize_submission(submission):
...
+        "agency": agency_name,
```
nmonga91: We can shorten the above if/else block here to be "agency": cgac.agency_name if cgac.agency_name else ''. Also, I'm not sure we want to display just a blank value if there's no agency associated with a submission (although that shouldn't be the case for a submission, since it should always have a submission), but we may want to return None instead of '' which will translate to null in the response JSON.

Either way, that value needs to probably be translated to something more user friendly. I'd refer to Kaitlin on what they'd like to see in that scenario.

kimminahm: Talked to Kaitlin, decided to use N/A in the case that agency does not exist, used a ternary operator to replace the if/else block

PR #536, opened 2017-02-03
Db-1594 - 20 rule change (padding)

kkuratsu:
Current code seem to validate '0' or '0000' without warnings. Added another line to validate '0' just in case.

### DIFF
--- a/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
+++ b/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
@@ -19,6 +19,7 @@ SELECT op.row_number,
 FROM object_class_program_activity as op
 WHERE op.submission_id = {0}
     AND CAST(COALESCE(op.beginning_period_of_availa,'0') AS integer) IN (SELECT DISTINCT CAST(budget_year AS integer) FROM program_activity)
+	AND LOWER(op.program_activity_code) NOT IN ('0000', 'unknown/other')
 	AND op.row_number NOT IN (
 		SELECT op.row_number
 		FROM object_class_program_activity_b9_{0} as op
--- a/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
+++ b/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
@@ -19,7 +19,7 @@ SELECT op.row_number,
 FROM object_class_program_activity as op
 WHERE op.submission_id = {0}
     AND CAST(COALESCE(op.beginning_period_of_availa,'0') AS integer) IN (SELECT DISTINCT CAST(budget_year AS integer) FROM program_activity)
-	AND op.program_activity_code <> '0000'
+    AND op.program_activity_code <> '0000'
     AND LOWER(op.program_activity_name) <> 'unknown/other'
 	AND op.row_number NOT IN (
 		SELECT op.row_number
--- a/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
+++ b/dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
@@ -19,7 +19,8 @@ SELECT op.row_number,
 FROM object_class_program_activity as op
 WHERE op.submission_id = {0}
     AND CAST(COALESCE(op.beginning_period_of_availa,'0') AS integer) IN (SELECT DISTINCT CAST(budget_year AS integer) FROM program_activity)
-	AND LOWER(op.program_activity_code) NOT IN ('0000', 'unknown/other')
+	AND op.program_activity_code <> '0000'
+    AND LOWER(op.program_activity_name) <> 'unknown/other'
 	AND op.row_number NOT IN (
 		SELECT op.row_number
 		FROM object_class_program_activity_b9_{0} as op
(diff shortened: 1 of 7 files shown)


nmonga91:
The acceptance criteria had the following:
- Program activities with the code '0000' and title 'Unknown/Other' will validate without warnings
- Program activities with just '0' will get padded to '0000'

The SQL is checking for '0' when it should be padding. This would require looking at how the data is loaded into the FileColumns model because that handles the padding for the fields for a given file type. It has a `padded_flag` and `length` which will auto-pad for you if the file it's loading that data from (start at schemaLoader.py). 

The query is also not checking for `Unknown/Other` in addition to `0000`. The description said "will validate without warnings", I believe that would mean we don't need to continue the query if the program activity code for object_class_program_activity is one of those values (also look at SQL `IN` clause). I would clarify with Kristine on that one.

kkuratsu:
Thanks for feedback. Awaiting Kristine's response for the clarification. Will make suggested changes.

kkuratsu:
It appears that the program activity code already gets padded, so '0' would become '0000'. Even "" would become '0000'.

```
def load_program_activity(filename):
    """Load program activity lookup table."""
    model = ProgramActivity

        sess = GlobalDB.db().session

        # for program activity, delete and replace values??
        sess.query(model).delete()

        data = pd.read_csv(filename, dtype=str)
        data = clean_data(
            data,
            model,
            {"year": "budget_year", "agency_id": "agency_id", "alloc_id": "allocation_transfer_id",
             "account": "account_number", "pa_code": "program_activity_code", "pa_name": "program_activity_name"},
            {"program_activity_code": {"pad_to_length": 4}, "agency_id": {"pad_to_length": 3},
             "allocation_transfer_id": {"pad_to_length": 3, "keep_null": True}, "account_number": {"pad_to_length": 4}}
        )
        # because we're only loading a subset of program activity info,
        # there will be duplicate records in the dataframe. this is ok,
        # but need to de-duped before the db load.
        data.drop_duplicates(inplace=True)
        # insert to db
        table_name = model.__table__.name
        num = insert_dataframe(data, table_name, sess.connection())
        sess.commit()

    logger.info('{} records inserted to {}'.format(num, table_name))
```

dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
```diff
@@ -19,6 +19,8 @@ SELECT op.row_number,
 FROM object_class_program_activity as op
 WHERE op.submission_id = {0}
     AND CAST(COALESCE(op.beginning_period_of_availa,'0') AS integer) IN (SELECT DISTINCT CAST(budget_year AS integer) FROM program_activity)
+	AND op.program_activity_code <> '0000'
```
nmonga91: I would fix the indentation here

dataactvalidator/config/sqlrules/b9_object_class_program_activity.sql
```diff
@@ -19,6 +19,8 @@ SELECT op.row_number,
...
+    AND LOWER(op.program_activity_name) <> 'unknown/other'
 	AND op.row_number NOT IN (
```
nmonga91: Same here with the indentation, it's best to keep it clean and consistent

tests/unit/dataactvalidator/test_b9_award_financial.py
```diff
@@ -55,6 +55,18 @@ def test_success_mismatched_year(database):
...
 
 
+def test_success_unknown_value(database):
+    """Program activity name/code as null"""
+    af_1 = AwardFinancialFactory(row_number=1, beginning_period_of_availa=2016, agency_identifier='test',
+                                 allocation_transfer_agency='test', main_account_code='test',
+                                 program_activity_name='test', program_activity_code='0000')
+    af_2 = AwardFinancialFactory(row_number=1, beginning_period_of_availa=2016, agency_identifier='test',
+                                 allocation_transfer_agency='test', main_account_code='test',
+                                 program_activity_name='test', program_activity_code='Unknown/Other')
```
nmonga91: So the allowable values for the related JIRA ticket are as follows:

- `program_activity_name`: "Unknown/Other"
- `program_activity_code`: "0000"

Here, `program_activity_code` is being set to "Unknown/Other" so we're not testing what the SQL is testing (at least for success). `program_activity_code='Unknown/Other'` _should_ fail if i'm not mistaken

tests/unit/dataactvalidator/test_b9_object_class_program_activity.py
```diff
@@ -42,6 +42,17 @@ def test_success_mismatched_year(database):
...
 
 
+def test_success_unknown_value(database):
+    op_1 = ObjectClassProgramActivityFactory(row_number=1, beginning_period_of_availa=2017, agency_identifier='test',
+                                             allocation_transfer_agency='test', main_account_code='test',
+                                             program_activity_name='test', program_activity_code='0000')
+
+    op_2 = ObjectClassProgramActivityFactory(row_number=1, beginning_period_of_availa=2017, agency_identifier='test',
+                                             allocation_transfer_agency='test', main_account_code='test',
+                                             program_activity_name='test', program_activity_code='Unknown/Other')
```
nmonga91: Same comment as above

PR #537, opened 2017-02-03
Db-1595 - Object class value auto padding

kkuratsu:
- Need the broker to auto pad 0 values to 000 for the object class values
- Object class values with '0' will pass validation

### DIFF
--- a/dataactvalidator/config/programActivityFields.csv
+++ b/dataactvalidator/config/programActivityFields.csv
@@ -13,7 +13,7 @@ grossoutlaysdeliveredorderspaidtotal_fyb,gross_outlays_delivered_or_fyb,TRUE,flo
 grossoutlaysundeliveredordersprepaidtotal_cpe,gross_outlays_undelivered_cpe,TRUE,float,21,"B5, B6, A18",
 grossoutlaysundeliveredordersprepaidtotal_fyb,gross_outlays_undelivered_fyb,TRUE,float,21,"A16, B5, B6",
 mainaccountcode,main_account_code,TRUE,str,4,A1,TRUE
-objectclass,object_class,TRUE,str,4,B11,TRUE
+objectclass,object_class,TRUE,str,4,B11
 obligationsdeliveredordersunpaidtotal_cpe,obligations_delivered_orde_cpe,TRUE,float,21,B4,
 obligationsdeliveredordersunpaidtotal_fyb,obligations_delivered_orde_fyb,TRUE,float,21,"A16, B4",
 obligationsincurredbyprogramobjectclass_cpe,obligations_incurred_by_pr_cpe,TRUE,float,21,A19,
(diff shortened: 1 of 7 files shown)


dataactvalidator/config/programActivityFields.csv
```diff
@@ -13,7 +13,7 @@ grossoutlaysdeliveredorderspaidtotal_fyb,gross_outlays_delivered_or_fyb,TRUE,flo
 grossoutlaysundeliveredordersprepaidtotal_cpe,gross_outlays_undelivered_cpe,TRUE,float,21,"B5, B6, A18",
 grossoutlaysundeliveredordersprepaidtotal_fyb,gross_outlays_undelivered_fyb,TRUE,float,21,"A16, B5, B6",
 mainaccountcode,main_account_code,TRUE,str,4,A1,TRUE
-objectclass,object_class,TRUE,str,4,B11,
+objectclass,object_class,TRUE,str,4,B11,TRUE
```
alburde1: You probably don't want to add the "padded_flag" field here. For some reason the field_length is 4 (still not sure why since they all seem to be 3 long, but you can ask someone that knows more about the rules than me about that). This means that every single objectclass field will be padded to 4 long with a leading 0. You should double-check that with several values just to be sure.

dataactvalidator/config/programActivityFields.csv
```diff
@@ -13,7 +13,7 @@ grossoutlaysdeliveredorderspaidtotal_fyb,gross_outlays_delivered_or_fyb,TRUE,flo
 grossoutlaysundeliveredordersprepaidtotal_cpe,gross_outlays_undelivered_cpe,TRUE,float,21,"B5, B6, A18",
 grossoutlaysundeliveredordersprepaidtotal_fyb,gross_outlays_undelivered_fyb,TRUE,float,21,"A16, B5, B6",
 mainaccountcode,main_account_code,TRUE,str,4,A1,TRUE
-objectclass,object_class,TRUE,str,4,B11,
+objectclass,object_class,TRUE,str,4,B11
```
alburde1: This is going to error, we need that comma

dataactvalidator/config/sqlrules/b11_object_class_program_activity_3.sql
```diff
@@ -3,4 +3,5 @@ SELECT
     op.object_class
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
-AND op.object_class = '000'
+AND LENGTH(op.object_class) <> 3
+AND op.object_class <> '0'
```
alburde1: It should never be "0" because of our padding. We can also probably just change this back since we still need it to be a warning.

dataactvalidator/config/sqlrules/b11_award_financial_3.sql
```diff
@@ -3,4 +3,5 @@ SELECT
     af.object_class
 FROM award_financial AS af
 WHERE af.submission_id = {}
-AND af.object_class = '000'
+AND LENGTH(af.object_class) <> 3
+AND af.object_class <> '0'
```
alburde1: Should never be 0 because of our padding. We can also probably just change this back since we still need it to be a warning.

tests/unit/dataactvalidator/test_b11_object_class_program_activity_3.py
```diff
@@ -15,15 +15,25 @@ def test_column_headers(database):
 def test_success(database):
```
alburde1: We can probably revert this file to the old tests

tests/unit/dataactvalidator/test_b11_award_financial_3.py
```diff
@@ -15,16 +15,25 @@ def test_column_headers(database):
 def test_success(database):
```
alburde1: We can probably revert this file to the old tests

PR #538, opened 2017-02-03
Adding Rule D3 SQL and tests

nmonga91:


nmonga91:
[run tests]

dataactvalidator/config/sqlrules/d3_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,8 @@
+-- Action type is required for non-aggregate records (i.e., when RecordType = 2)
+SELECT
+    dafa.row_number,
+    dafa.action_type,
+    dafa.record_type
+FROM detached_award_financial_assistance as dafa
+WHERE dafa.record_type = 2
+    AND COALESCE(dafa.action_type, '') = '' ;
```
alburde1: Why this method instead of `AND action_type IS NULL`? If this is faster then that's fine, but I'm just wondering since it seems like this would require a function call and therefore be slower?
Also, just a note, you don't actually have to do `dafa.` if there's only one table in the list. You can just put the column names, just a clean thing.

tests/unit/dataactvalidator/test_d3_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,27 @@
...
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "action_type", "record_type"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset == actual
+
+
+def test_success(database):
+    """ Tests success for when Action type is required for non-aggregate records (i.e., when RecordType = 2) """
+    det_award = DetachedAwardFinancialAssistanceFactory(record_type="2")
```
alburde1: Should there be a test here that proves that other record types (1) don't break it if they don't have action types?

nmonga91: adding that now

dataactvalidator/config/sqlrules/d3_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,8 @@
+-- Action type is required for non-aggregate records (i.e., when RecordType = 2)
+SELECT
+    row_number,
+    action_type,
+    record_type
+FROM detached_award_financial_assistance
+WHERE record_type = 2
+    AND (action_type = '' OR action_type IS NULL);
```
alburde1: I think if you don't enter anything into the column it automatically casts it to a null. Double-check that? It might alleviate your concerns about the query being complicated.

```
WHERE dafa.submission_id = {0}
    AND dafa.record_type = 2
    AND dafa.fain IS NULL
```
worked just fine for me in D1.

alburde1: Also... just noticed, you aren't checking for submission_id anywhere in this query

dataactvalidator/config/sqlrules/d3_detached_award_financial_assistance_2.sql
```diff
@@ -0,0 +1,7 @@
+-- ActionType field must contain A, B, C, D, or blank
+SELECT
+    row_number,
+    action_type,
+    record_type
+FROM detached_award_financial_assistance
+WHERE LOWER(COALESCE(action_type, '')) NOT IN ('', 'a', 'b', 'c', 'd') ;
```
alburde1: not checking for submission_id here either

PR #544, opened 2017-02-03
Adding rule D7 and related tests

nmonga91:


dataactvalidator/config/sqlrules/d7_detached_award_financial_assistance.sql
```diff
@@ -0,0 +1,8 @@
+-- URI is a required field for aggregate records (i.e., when RecordType = 1)
+SELECT
+    row_number,
```
alburde1: Do we want to select the record_type too just so they know what it is?

nmonga91: for b12_award_financial_1 we don't return all the rows just the one the rule is specifically checking so I did the same here. No harm in including it though, what do you think?

alburde1: Ahh, I included it in d1 for the same error for the fain because technically we're checking against record_type as well. It's not an error if record_type is 2 so I thought it would be good to have that there for them to see

nmonga91: Let's add it

PR #547, opened 2017-02-07
Add exception for '0'/'00'

kkuratsu:
We can't pad the 0 value to 000 for the object class because it currently accept 3 and 4 digits.  The change to this commit is to add an exception for `0` and `00` and add test cases.

### DIFF
--- a/dataactvalidator/config/sqlrules/b11_award_financial_1.sql
+++ b/dataactvalidator/config/sqlrules/b11_award_financial_1.sql
@@ -4,4 +4,4 @@ SELECT
 FROM award_financial AS af
 WHERE af.submission_id = {}
 AND af.object_class NOT IN (SELECT object_class_code FROM object_class)
-AND af.object_class <> '000'
+AND af.object_class NOT IN ('000', '00', '0')
--- a/dataactvalidator/config/sqlrules/b11_award_financial_3.sql
+++ b/dataactvalidator/config/sqlrules/b11_award_financial_3.sql
@@ -3,4 +3,4 @@ SELECT
     af.object_class
 FROM award_financial AS af
 WHERE af.submission_id = {}
-AND af.object_class = '000'
+AND af.object_class IN ('000', '00', '0')
--- a/dataactvalidator/config/sqlrules/b11_object_class_program_activity_1.sql
+++ b/dataactvalidator/config/sqlrules/b11_object_class_program_activity_1.sql
@@ -4,4 +4,4 @@ SELECT
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
 AND op.object_class NOT IN (SELECT object_class_code FROM object_class)
-AND op.object_class <> '000'
+AND op.object_class NOT IN ('000', '00', '0')
--- a/dataactvalidator/config/sqlrules/b11_object_class_program_activity_3.sql
+++ b/dataactvalidator/config/sqlrules/b11_object_class_program_activity_3.sql
@@ -3,4 +3,5 @@ SELECT
     op.object_class
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
-AND op.object_class = '000'
+AND op.object_class IN ('000', '00', '0')
+
--- a/tests/unit/dataactvalidator/test_b11_award_financial_3.py
+++ b/tests/unit/dataactvalidator/test_b11_award_financial_3.py
@@ -24,7 +24,12 @@ def test_success(database):
 def test_failure(database):
     """ Test invalid object class code (3 digits) """
 
-    # This should return because if it's '000' a warning should be returned
+    # This should return because if it's '000', '00', '0' a warning should be returned
     af = AwardFinancialFactory(object_class='000')
+    assert number_of_errors(_FILE, database, models=[af]) == 1
 
+    af = AwardFinancialFactory(object_class='00')
+    assert number_of_errors(_FILE, database, models=[af]) == 1
+
+    af = AwardFinancialFactory(object_class='0')
     assert number_of_errors(_FILE, database, models=[af]) == 1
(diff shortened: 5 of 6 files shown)


dataactvalidator/config/sqlrules/b11_object_class_program_activity_3.sql
```diff
@@ -3,4 +3,7 @@ SELECT
     op.object_class
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
-AND op.object_class = '000'
+AND LENGTH(op.object_class) <> 3
+AND op.object_class <> '0'
```
alburde1: We were only looking for '000' so we want to keep looking for that along with any permutation of it. In this case, we want to use
```
AND (op.object_class = '000' OR op.object_class = '00' OR op.object_class = '0')
```
only formatted nicer

dataactvalidator/config/sqlrules/b11_award_financial_3.sql
```diff
@@ -3,4 +3,6 @@ SELECT
     af.object_class
 FROM award_financial AS af
 WHERE af.submission_id = {}
-AND af.object_class = '000'
+AND LENGTH(af.object_class) <> 3
```
alburde1: We were only looking for '000' so we want to keep looking for that along with any permutation of it. In this case, we want to use

```
AND (op.object_class = '000' OR op.object_class = '00' OR op.object_class = '0')
```

only formatted nicer

dataactvalidator/config/sqlrules/b11_award_financial_1.sql
```diff
@@ -4,4 +4,5 @@ SELECT
 FROM award_financial AS af
 WHERE af.submission_id = {}
 AND af.object_class NOT IN (SELECT object_class_code FROM object_class)
-AND af.object_class <> '000'
+AND af.object_class <> '0'
+AND af.object_class <> '00'
```
alburde1: We still want to look for `af.object_class <> '000'` since it's the same warning as before

dataactvalidator/config/sqlrules/b11_object_class_program_activity_1.sql
```diff
@@ -4,4 +4,5 @@ SELECT
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
 AND op.object_class NOT IN (SELECT object_class_code FROM object_class)
-AND op.object_class <> '000'
+AND op.object_class <> '0'
+AND op.object_class <> '00'
```
alburde1: We still want to look for `af.object_class <> '000'` since it's the same warning as before

tests/unit/dataactvalidator/test_b11_award_financial_3.py
```diff
@@ -12,19 +11,20 @@ def test_column_headers(database):
...
+    """ Test invalid object class code (3 digits) """
 
-    af = AwardFinancialFactory(object_class='object_class')
-    oc = ObjectClassFactory(object_class_code='object_class')
+    # This should return because if it's '000' a warning should be returned
+    af_1 = AwardFinancialFactory(object_class='0')
+    af_2 = AwardFinancialFactory(object_class='00')
 
-    assert number_of_errors(_FILE, database, models=[af, oc]) == 0
+    assert number_of_errors(_FILE, database, models=[af_1, af_2]) == 0
```
alburde1: It should give the same warning as before for "0", "00", or "000". The success case shouldn't change

tests/unit/dataactvalidator/test_b11_award_financial_3.py
```diff
@@ -12,19 +11,20 @@ def test_column_headers(database):
...
 
 
 def test_failure(database):
     """ Test invalid object class code (3 digits) """
 
     # This should return because if it's '000' a warning should be returned
-    af = AwardFinancialFactory(object_class='000')
+    af = AwardFinancialFactory(object_class='12345')
```
alburde1: We should still be getting a warning for "000" but we should also now get it for "0" and "00"

tests/unit/dataactvalidator/test_b11_object_class_program_activity_3.py
```diff
@@ -12,18 +11,18 @@ def test_column_headers(database):
...
-    assert number_of_errors(_FILE, database, models=[op, oc]) == 0
+    assert number_of_errors(_FILE, database, models=[op_1, op_2]) == 0
 
 
 def test_failure(database):
-    """ Test invalid object class code (3 digits) """
+    """ Test invalid object class code """
 
-    op = ObjectClassProgramActivityFactory(object_class='000')
+    op = ObjectClassProgramActivityFactory(object_class='12345')
```
alburde1: See comments above for this test file too.

tests/unit/dataactvalidator/test_b11_object_class_program_activity_3.py
```diff
@@ -24,6 +24,12 @@ def test_success(database):
...
     """ Test invalid object class code (3 digits) """
 
+    # This should return because if it's '000', '00', '0' a warning should be returned
     op = ObjectClassProgramActivityFactory(object_class='000')
+    assert number_of_errors(_FILE, database, models=[op]) == 1
+
+    op = ObjectClassProgramActivityFactory(object_class='00')
+    assert number_of_errors(_FILE, database, models=[op]) == 1
 
+    op = ObjectClassProgramActivityFactory(object_class='0')
```
alburde1: We don't need to separate these into different asserts. You can just set each of them to op_1, op_2, and op_3 and then do this:
`assert number_of_errors(_FILE, database, models=[op_1, op_2, op_3]) == 3`
Same for the other test file

kkuratsu: This will give 3 warnings instead of 1 for each of '0' input. I think it's confusing if there is 3 warnings for one invalid input?

PR #549, opened 2017-02-09
db-1618-d9-multiple-recipients

alburde1:
Validation to make sure that records with a record_type of 1 only contain "multiple recipients" in the AwardeeOrRecipientLegalEntityName column

Technical Changes:
- New sql rules


kkuratsu:
I need to make sure my records have a recipient name

Acceptance Criteria:
- [ ] AwardeeOrRecipientLegalEntityName is a required field for all submissions and cannot be blank.
- [ ] Must contain "MULTIPLE RECIPIENTS" for aggregate records (i.e., when RecordType = 1).

The A/C 1 is missing for a AwardeeOrRecipientLegalEntityName required field and cannot be blank.

alburde1:
Nope, it's a required field and whether you put in nothing or spaces it still throws a "missing required field" error

kkuratsu:
Do we need a test case for the Null value to test against the A/C?

PR #550, opened 2017-02-09
DB-1670-fabs-intermediate-table

alburde1:
Creates a table to hold valid published rows for FABS submissions. Rows are not associated with a specific submission, job, or row number

Technical changes:
- New table, will need to run `alembic upgrade head`


dataactcore/models/stagingModels.py
```diff
@@ -534,3 +534,58 @@ def __init__(self, **kwargs):
...
+    period_of_performance_curr = Column(Text)
+    period_of_performance_star = Column(Text)
+    place_of_performance_code = Column(Text)
+    place_of_performance_congr = Column(Text)
+    place_of_perform_country_c = Column(Text)
+    place_of_performance_forei = Column(Text)
+    place_of_performance_zip4a = Column(Text)
+    record_type = Column(Integer)
+    sai_number = Column(Text)
+    uri = Column(Text, index=True)
```
nmonga91: Only FAIN+URI determines a unique row? I thought you had mentioned 4 attributes that should make up the key

alburde1: They're FAIN, AwardModificationAmendmentNumber, URI, and AwardingSubTierAgencyCode but I didn't think index had to be unique. I honestly only kept that because the previous two award financial assistance tables had index on these columns.

The PK for this table is the `published_award_financial_assistance_id` anyway in this case. Would it make more sense to figure out how to create a composite key as the PK?

nmonga91: There are 2 options:

1. create a composite key (http://docs.sqlalchemy.org/en/latest/core/constraints.html#sqlalchemy.schema.PrimaryKeyConstraint): This only works if we're not going to reference these values as a foreign key anywhere else.
2. create a unique constraint (http://docs.sqlalchemy.org/en/latest/core/constraints.html#unique-constraint): This is desirable when we want to reference a row here in another table. That way we can use the single primary key (the id column) instead of the entire composite key while maintaining the constraint that those 4 fields should be unique

The index should be on whatever will be used for filtering the most, so the 4 columns you mentioned since we'll always be looking for something that matches that set of values.

dataactcore/migrations/versions/88e9b634ca1a_create_published_award_financial_.py
```diff
@@ -0,0 +1,96 @@
...
+    sa.Column('place_of_perform_country_c', sa.Text(), nullable=True),
+    sa.Column('place_of_performance_forei', sa.Text(), nullable=True),
+    sa.Column('place_of_performance_zip4a', sa.Text(), nullable=True),
+    sa.Column('record_type', sa.Integer(), nullable=True),
+    sa.Column('sai_number', sa.Text(), nullable=True),
+    sa.Column('uri', sa.Text(), nullable=True),
+    sa.PrimaryKeyConstraint('published_award_financial_assistance_id'),
+    sa.UniqueConstraint('awarding_sub_tier_agency_c', 'award_modification_amendme', 'fain', 'uri', name='uniq_award_mod_sub_tier_fain_uri')
+    )
+    op.create_index(op.f('ix_published_award_financial_assistance_award_modification_amendme'), 'published_award_financial_assistance', ['award_modification_amendme'], unique=False)
```
nmonga91: These were auto-generated by alembic correct?

alburde1: Yes, I didn't touch the generation file in any way. The only part I modified was the model.

nmonga91:
Looks good, merging!

PR #557, opened 2017-02-13
Adding implementation and tests for Rule D4 for FABS submissions

nmonga91:
Adding implementation and tests for Rule D4 for FABS submissions


kkuratsu:
Do we need to have a test case for null?
A/C: * This is a required field for all submissions and cannot be blank.

nmonga91:
The requirement check happens at an earlier stage and therefore does not need to happen at the SQL level

nmonga91:
We should be using `COALESCE` though. I will add that.

kkuratsu:
Why the test for nulll in d4_2 and d4_3 passes the test? Should it give an error if the action_date is null?

nmonga91:
d4_2 and d4_3 should pass for Null because they operate assuming they operate on an actual date. If it's not a date, there's no point in running them. We can have @kaitlin confirm that that functionality/assumption is acceptable though.

kaitlin:
Makes sense to me. We shouldn't run this validation if the value is null.

PR #560, opened 2017-02-15
Db 1668 submit fabs file

alburde1:
This PR creates a submit/publish route for FABS submissions. It also updates the `check_status` route to return the `publish_status` of the submission.

May need a second PR if updates are needed for the front end.

Technical Updates:
None


nmonga91:
[run tests]

nmonga91:
[run tests]

klrBAH:
[run tests]

dataactbroker/handlers/fileHandler.py
```diff
@@ -779,6 +780,75 @@ def check_generation(submission, file_type):
 
         return JsonResponse.create(StatusCode.OK, response_dict)
 
+    @staticmethod
+    def submit_detached_file(submission):
+        """ Submits the FABS upload file associated with the submission ID """
+        # Check to make sure it's a d2 submission
+        if not submission.d2_submission:
+            raise ResponseException("This is not a FABS submission", StatusCode.CLIENT_ERROR)
```
nmonga91: We may want to reword these error messages. using "This" is very specific. We may want to say something like "Submission is not a FABS submission". Should probably consult Kaitlin about this depending on how the message is displayed on the front-end as well

dataactbroker/handlers/fileHandler.py
```diff
@@ -779,6 +780,75 @@ def check_generation(submission, file_type):
...
+            # get all valid lines for this submission
+            query = sess.query(DetachedAwardFinancialAssistance).\
+                filter_by(is_valid=True, submission_id=submission_id).all()
+
+            for row in query:
+                # if it is not a delete row
+                if row.correction_late_delete_ind is None or row.correction_late_delete_ind.upper() != "D":
+                    # remove all keys in the row that are not in the intermediate table
+                    temp_obj = row.__dict__
+                    temp_obj.pop('detached_award_financial_assistance_id', None)
```
nmonga91: Is there a way to pop a list of attributes instead of doing this one by one?

alburde1: Not that I've seen. I could make a for loop and a list of all the keys I want out but all that would do would be to make it look nicer. It wouldn't actually change how it's working.

tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -267,8 +268,12 @@ def test_submission_bad_dates(start_date, end_date, quarter_flag, submission):
 
 
 def test_submission_to_dict_for_status(database):
+    for ps in PUBLISH_STATUS:
```
nmonga91: There's an easier way to add this by default instead of doing it manually, using fixtures. Take a look at https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/tests/unit/conftest.py#L49-L51

nmonga91:
Looks good, merging!

PR #562, opened 2017-02-15
db-1209 Performed only w a transaction obligated amount value

kkuratsu:
Acceptance Criteria:
- C8/C9 validation is only performed on rows in file C that have a transaction obligated amount value in the field
- C11/C12 validation is only performed on rows in file C that have a transaction obligated amount value in the field

Technical Changes:
SQL Rule Updates

### DIFF
--- a/dataactvalidator/config/sqlrules/c12_cross_file.sql
+++ b/dataactvalidator/config/sqlrules/c12_cross_file.sql
@@ -12,4 +12,8 @@ WHERE ap.submission_id = {}
 		WHERE af.submission_id = ap.submission_id
 		    AND af.piid = ap.piid
 		    AND af.parent_award_id IS NOT DISTINCT FROM ap.parent_award_id
+	) AND NOT EXISTS(
+	  SELECT af.transaction_obligated_amou
+	  FROM award_financial AS af
+	  WHERE af.transaction_obligated_amou IS NULL
 	)
--- a/tests/unit/dataactvalidator/test_c11_cross_file.py
+++ b/tests/unit/dataactvalidator/test_c11_cross_file.py
@@ -15,24 +15,41 @@ def test_column_headers(database):
 def test_success(database):
     """ Unique PIID, ParentAwardId from file C exists in file D1 during the same reporting period. """
 
-    af = AwardFinancialFactory(piid='some_piid', parent_award_id='some_parent_award_id')
+    af = AwardFinancialFactory(piid='some_piid', parent_award_id='some_parent_award_id',
+                               transaction_obligated_amou='12345')
     ap = AwardProcurementFactory(piid='some_piid', parent_award_id='some_parent_award_id')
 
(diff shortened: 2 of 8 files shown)


klrBAH:
[run tests]

dataactvalidator/config/sqlrules/c12_cross_file.sql
```diff
@@ -12,4 +12,8 @@ WHERE ap.submission_id = {}
 		WHERE af.submission_id = ap.submission_id
 		    AND af.piid = ap.piid
 		    AND af.parent_award_id IS NOT DISTINCT FROM ap.parent_award_id
+	) AND NOT EXISTS(
+	  SELECT transaction_obligated_amou
+	  FROM award_financial AS af
+	  WHERE transaction_obligated_amou IS NULL
```
alburde1: I'm not the best at SQL so maybe this wouldn't work, but why not just add `AND af.transaction_obligated_amou IS NULL` to the NOT EXISTS above?

tests/unit/dataactvalidator/test_c11_cross_file.py
```diff
@@ -15,39 +15,58 @@ def test_column_headers(database):
...
 
-    af = AwardFinancialFactory(piid='some_piid', parent_award_id=None)
+    af = AwardFinancialFactory(piid='some_piid', parent_award_id=None, transaction_obligated_amou='12345')
     ap = AwardProcurementFactory(piid='some_piid', parent_award_id=None)
 
     assert number_of_errors(_FILE, database, models=[af, ap]) == 0
 
+    # Not perform when no transaction obligated amount value in the field
+    af = AwardFinancialFactory(piid='some_piid', parent_award_id='some_parent_award_id',
+                               allocation_transfer_agency=None, transaction_obligated_amou=None)
```
alburde1: Is allocation_transfer_agency needed here? What is it testing that isn't being tested by the rest of the parameters?

dataactvalidator/config/sqlrules/c12_cross_file.sql
kkuratsu: Originally, I had the `af.transaction_obligated_amou IS NULL` inside the `NOT EXISTS` and it doesn't work. The test for PIID, ParentAwardID from file D1 exists in file C fails. So I have to move that outside.

tests/unit/dataactvalidator/test_c11_cross_file.py
kkuratsu: It tests if the `transction_obligated_award_id is NULL`, it should pass even with af.PIID is not the same with ap.PIID. We don't need that, but I added to show that it meet the acceptance criteria only performed on rows that have a `transction_obligated_award_id`.

I can remove that if you think we don't need to test that.

alburde1: Right, that testing is important, I was just asking specifically why `allocation_transfer_agency` is being set

PR #566, opened 2017-02-15
Adding ability to delete a submission and ALL related data

nmonga91:
Added a new route to allow deleting a submission and all related data. This is a somewhat dangerous implementation so it should be confirmed/approved prior to merging.

- [x] Reviewed by backend
- [x] Approved by @kaitlin 
- [x] Integration with front-end tested (coordinated with @kimminahm) 

Technical Release Notes:
- New alembic migration revision added to include cascading on delete for all models associated with a submission (must run `alembic upgrade head`)


nmonga91:
[run tests]

dataactcore/models/stagingModels.py
```diff
@@ -1,4 +1,5 @@
-from sqlalchemy import Column, Integer, Text, Numeric, Index, Boolean, UniqueConstraint
+from sqlalchemy import Column, Integer, Text, Numeric, Index, Boolean, UniqueConstraint, ForeignKey
+from sqlalchemy.orm import relationship
```
alburde1: the DetachedAwardFinancialAssistance table probably also needs one of these FKs

nmonga91: Left it out originally cuz it was messing with another branch. Thanks for the reminder. Added it back in.

dataactbroker/fileRoutes.py
```diff
@@ -174,6 +176,20 @@ def post_submission_narratives(submission):
...
+    @convert_to_submission_id
+    @requires_submission_perms('writer')
+    def delete_submission(submission):
+        """ Deletes all data associated with the specified submission
+        NOTE: THERE IS NO WAY TO UNDO THIS """
+
+        sess = GlobalDB.db().session
+        sess.query(Submission).filter(Submission.submission_id == submission.submission_id).delete(
+            synchronize_session=False)
+        sess.expire_all()
```
alburde1: Meant to ask this before, does this expire all things with a relationship to this particular submission? For example, if the session has an awardfinancial object related to the submission, will it think it isn't expired? (don't know if this is relevant to how we do anything, but worth double-checking)

dataactcore/models/jobModels.py
```diff
@@ -117,10 +117,10 @@ class JobDependency(Base):
     __tablename__ = "job_dependency"
```
alburde1: Missed this last time, should Job also have a cascade on delete or no? Since we're cascading a delete on job_id for a lot of them.

dataactbroker/fileRoutes.py
nmonga91: This basically expunges any stale references to deleted objects from the session. So this way, we're not actually referencing the deleted objects anywhere else.

dataactcore/models/jobModels.py
nmonga91: Yeah we should. A job can't exist without a submission (because then it doesn't really serve a purpose) and if we're deleting everything related to a submission, that includes the related Jobs and their dependencies.

alburde1: Oh, it already had one. I somehow missed it and thought it didn't have one.

kaitlin:
Looks good to me! I approve this PR.

nmonga91:
Integration was tested locally and works correctly. There is some cleanup needed for the frontend but backend seems to be working as expected.

PR #570, opened 2017-02-16
Db 1626 c23 validation parent award id piid

alburde1:
Update C23 rules to compare just PIID if parent_award_id is null in file C or both PIID and parent_award_id if it isn't.

Technical Changes:
- SQL Rule Updates


nmonga91:
Can we add comments to the top of each indicating what part of C23 is being checked for that specific SQL file?

nmonga91:
Looks good, ready to merge

PR #581, opened 2017-02-23
Bug - Remove D/R validation from B11

kkuratsu:
Background:
Agency is populating an object class for all lines in File C so object class is not blank but they are still failing validation B11 because of blank reimbursable/direct indicator.

Error message says:
B11 - Failed rule: If object class is 4 digits, check that 1 digit prefix is consistent with value for ByDirectReimbursableFundingSource

*The D/R check and message should not be part of B11. B11 should only check for Object Class validity.*

Technical changes:
• Removed `b11_2` sql files and `b11_2` test files
• Renamed `b11_3` sql files and `b11_3` test files to `b11_2`

Note: The changes you see in the files are a result of renaming files. No code changes.


nmonga91:
Looks good, merging

PR #585, opened 2017-02-27
DB 1682 Flex Fields Cross File

alburde1:
Add flex fields to cross validation error/warning files

Technical Updates:
* DB update


dataactcore/models/stagingModels.py
```diff
@@ -19,6 +19,8 @@ class FlexField(Base):
     row_number = Column(Integer, nullable=False)
     header = Column(Text)
     cell = Column(Text)
+    file_type_id = Column(Integer)
+    file_letter_name = Column(Text)
```
alburde1: I can look into the lookup thing, that would make sense. Adding a relationship seems like overkill since it would require a foreign key

dataactvalidator/validation_handlers/validator.py
```diff
@@ -168,13 +168,24 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict):
...
+
             for row in failed_rows:
                 # get list of values for each column
                 values = ["{}: {}".format(short_to_long_dict[c], str(row[c])) if c in short_to_long_dict else
                           "{}: {}".format(c, str(row[c])) for c in cols]
                 values = ", ".join(values)
+                full_column_string = column_string
+                # go through all flex fields in this row and add to the columns and values
+                for field in flex_data[row['row_number']]:
+                    full_column_string += ", " + field.header + "_file" + field.file_letter_name.lower()
```
alburde1: I'm not sure if we check for duplicates. If we do, that's not what the issue is here. It's possible to have a flex field called `flex_divOp` in file A and also have one called `flex_divOp` in file B and that's not disallowed because it might be needed. We need to differentiate between the two somehow in the error/warning files for cross-file validations since columns from both files are in the output files

nmonga91:
Looks good, merging

PR #586, opened 2017-02-28
Db 1535 publish submission

nmonga91:
Added functionality to certify a submission as well as restart validation for a submission that is older than 02/22/17. 

- [x] Tested & approved by @kaitlin on Sandbox
- [x] Reviewed by backend

Technical Release Notes:
- Added check to restrict deleting a certified submission
- Added new route: `/v1/certify_submission`
- Added new route: `/v1/restart_validation`


nmonga91:
[run tests]

nmonga91:
[run tests]

dataactcore/migrations/versions/56f78c0fc729_added_last_validated_to_submission.py
```diff
@@ -0,0 +1,42 @@
...
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('submission', sa.Column('last_validated', sa.DateTime(), nullable=True))
+    op.drop_column('submission', 'datetime_utc')
```
alburde1: Why are we removing this datetime? Just wondering.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -305,6 +306,11 @@ def run_validation(self, job):
             error_list.write_all_row_errors(job_id)
             # Update error info for submission
             populate_submission_error_info(submission_id)
+
+            # Update last validated date
+            job.submission.last_validated = datetime.utcnow()
+            sess.commit()
```
alburde1: I can't comment higher up because GitHub, but do we need the sess.commit on line 304 if we're committing right here?

dataactbroker/fileRoutes.py
```diff
@@ -190,13 +190,35 @@ def delete_submission(submission):
...
+    @requires_submission_perms('submitter', check_owner=False)
+    def certify_submission(submission):
+        if submission.publishable:
+            sess = GlobalDB.db().session
+            submission.publish_status_id = PUBLISH_STATUS_DICT['published']
+            sess.commit()
+        else:
+            return JsonResponse.error(ValueError("Submission cannot be certified due to critical errors"),
+                                      StatusCode.CLIENT_ERROR)
+        return JsonResponse.create(StatusCode.OK, {"message": "Success"})
```
nmonga91: Correct we can move that up

dataactcore/migrations/versions/56f78c0fc729_added_last_validated_to_submission.py
nmonga91: It does the same thing as `created_at` so this field is deprecated.

dataactbroker/handlers/fileHandler.py
```diff
@@ -999,6 +999,34 @@ def create_response_dict_for_submission(self, upload_files, submission, existing
...
+        sess = GlobalDB.db().session
+        initial_file_types = [FILE_TYPE_DICT['appropriations'], FILE_TYPE_DICT['program_activity'],
+                              FILE_TYPE_DICT['award_financial']]
+
+        jobs = sess.query(Job).filter(Job.submission_id == submission.submission_id).all()
+
+        # set all jobs to their initial status of "waiting"
+        for job in jobs:
+            job.job_status_id = JOB_STATUS_DICT['waiting']
+        sess.commit()
```
nmonga91: Nope we don't. The commit below will suffice.

dataactvalidator/validation_handlers/validationManager.py
nmonga91: Possibly, depending on whether or not the methods called after perform queries that would rely on the updated information. I'll look into that

nmonga91: yup, we need 2 here because `populate_submission_error_info` calls `sum_number_of_errors_for_job_list` which does a query on the Job table

alburde1: Ok, thanks for checking

dataactcore/interfaces/function_bag.py
```diff
@@ -609,3 +609,26 @@ def get_submission_status(submission):
...
     return status
+
+
+def get_last_validated_date(submission_id):
+    """ Return the oldest last validated date for validation jobs """
+    sess = GlobalDB.db().session
+
+    validation_job_types = [JOB_TYPE_DICT['csv_record_validation'], JOB_TYPE_DICT['validation']]
+
+    jobs = sess.query(Job).filter(and_(Job.submission_id == submission_id,
```
alburde1: You don't need to use "and" here. You can just put the comma and it implies it

nmonga91:
[run tests]

PR #591, opened 2017-03-01
Db 1690 d24 place of performance country code

alburde1:
Rule D24, PrimaryPlaceOfPerformanceCountryCode matches a code from the country code list

Also updating the country_codes.csv file to contain the 3 entities mentioned in the ticket

Technical Release Notes:
- Updated country codes table contents (will need a re-initialize)
- New sql rules


kkuratsu:
On of the A/C is `PrimaryPlaceOfPerformanceCountryCode is a required field for all submissions and cannot be blank.`. Can we check for a blank here as well?

alburde1:
As with every "always required" value, it isn't checked by the rules but rather by the column requirements in `detachedAwardsFields.csv`

PR #598, opened 2017-03-03
Db 1632 sortable submissions

kimminahm:
Added logic content for the api routes `/list_submission` to be able to sort and order

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1176,12 +1176,14 @@ def get_error_metrics(submission):
         return JsonResponse.error(e, StatusCode.INTERNAL_ERROR)
 
 
-def list_submissions(page, limit, certified):
+def list_submissions(page, limit, certified, sort, order):
     """ List submission based on current page and amount to display. If provided, filter based on
     certification status """
     sess = GlobalDB.db().session
 
     offset = limit * (page - 1)
+
+    postSort = ''
 
     cgac_codes = [aff.cgac.cgac_code for aff in g.user.affiliations]
     query = sess.query(Submission).filter_by(d2_submission=False)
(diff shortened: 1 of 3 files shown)


tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -20,6 +20,53 @@ def list_submissions_result():
...
+    sub3 = SubmissionFactory(user_id=2, submission_id=3, number_of_warnings=1, reporting_start_date=date(2010, 1, 3))
+    sub4 = SubmissionFactory(user_id=2, submission_id=4, number_of_warnings=1, reporting_start_date=date(2010, 1, 4))
+    sub5 = SubmissionFactory(user_id=2, submission_id=5, number_of_warnings=1, reporting_start_date=date(2010, 1, 5))
+    add_models(database, [user1, user2, sub1, sub2, sub3, sub4, sub5])
+
+    monkeypatch.setattr(fileHandler, 'g', Mock(user=user1))
+    result = list_submissions_sort('reporting', 'desc')
+    assert result['total'] == 5
+    sub = result['submissions'][0]
+    for subit in result['submissions']:
```
nmonga91: is it possible to use set comparison here potentially since we're controlling the number of submissions available/returned? I was thinking that we could compare both sets instead of looping through them to check each submission individually

dataactbroker/handlers/fileHandler.py
```diff
@@ -1176,7 +1176,7 @@ def get_error_metrics(submission):
         return JsonResponse.error(e, StatusCode.INTERNAL_ERROR)
 
 
-def list_submissions(page, limit, certified):
+def list_submissions(page, limit, certified, sort='modified', order='desc'):
```
kimminahm: good call, fileRoutes.py has default values set incase of missing args so it should be fine to remove

tests/unit/dataactbroker/test_file_handler.py
kimminahm: I do not think so, using set comparisons can determine existence within the set but not the order, which is what this should be testing for

dataactbroker/handlers/fileHandler.py
kimminahm: actually, if i remove the default values here then the py.test will fail because it calls the function directly

nmonga91:
Looks good, ready to merge

PR #599, opened 2017-03-03
Db 1694 d30 business funds indicator

alburde1:
- Making BusinessFundsIndicator required
- Rule to test that BusinessFundsIndicator is either "non" or "rec"

Technical Release Notes:
- New sql rules
- Update to file_columns table contents


dataactvalidator/config/sqlrules/d30_detached_award_financial_assistance.sql
```diff
@@ -0,0 +1,8 @@
+-- BusinessFundsIndicator must contain one of the following values: REC or NON.
+SELECT
+    row_number,
+    business_funds_indicator
+FROM detached_award_financial_assistance
+WHERE submission_id = {0}
+    AND UPPER(business_funds_indicator) != 'REC'
```
nmonga91: It's a little cleaner if we combine these checks into one with something like `UPPER(business_funds_indicator) NOT IN ('REC', 'NON')`

nmonga91:
Looks good, merging!

PR #604, opened 2017-03-09
Db 1692 d32-d34 period of performance dates

alburde1:
Creates rules D32-D34 pertaining to PeriodOfPerformanceStartDate and PeriodOfPerformanceCurrentEndDate.

Technical Release Notes:
- New sql rules


dataactvalidator/config/sqlrules/d32_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,17 @@
...
+  return FALSE;
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT
+    row_number,
+    period_of_performance_star
+FROM detached_award_financial_assistance
+WHERE submission_id = {0}
+    AND period_of_performance_star IS NOT NULL
```
nmonga91: do we want this to fail if it's blank? or should we treat blank and NULL the same?

dataactvalidator/config/sqlrules/d33_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,17 @@
...
+  return FALSE;
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT
+    row_number,
+    period_of_performance_curr
+FROM detached_award_financial_assistance
+WHERE submission_id = {0}
+    AND period_of_performance_curr IS NOT NULL
```
nmonga91: Same as the previous question for blanks vs nulls

tests/unit/dataactvalidator/test_d32_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "period_of_performance_star"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset == actual
+
+
+def test_success(database):
+    """ PeriodOfPerformanceStartDate is an optional field, but when provided, must follow YYYYMMDD format """
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(period_of_performance_star="19990131")
```
nmonga91: we should probably add a test case for a null value to make sure it's omitted. And if we decide to avoid blanks, a blank test case should also be added.

tests/unit/dataactvalidator/test_d33_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_column_headers(database):
+    expected_subset = {"row_number", "period_of_performance_curr"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset == actual
+
+
+def test_success(database):
+    """ PeriodOfPerformanceCurrentEndDate is an optional field, but when provided, must follow YYYYMMDD format """
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(period_of_performance_curr="19990131")
```
nmonga91: Same comment as D32_1 for nulls/blanks

tests/unit/dataactvalidator/test_d33_detached_award_financial_assistance_2.py
```diff
@@ -0,0 +1,31 @@
...
+    expected_subset = {"row_number", "period_of_performance_curr"}
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset == actual
+
+
+def test_success(database):
+    """ When provided, PeriodOfPerformanceCurrentEndDate must be a valid date between 19991001 and 20991231.
+        (i.e., a date between 10/01/1999 and 12/31/2099) """
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(period_of_performance_curr="20120725")
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(period_of_performance_curr=None)
```
nmonga91: We should also probably add a test case for blank here too, as well as other tests

dataactvalidator/config/sqlrules/d32_detached_award_financial_assistance_1.sql
alburde1: I did this a while ago but I think I tested and if someone enters blank or just spaces it defaults to null in the DB

nmonga91: Ok cool. If the broker inserts it properly when uploaded we may be good. But I know that for the unit tests, we can force a blank value so not sure how we want to handle that in the SQL

alburde1: I can just put it in. I was just trying to simplify but it certainly won't hurt anything.

nmonga91:
Looks good, ready to merge once checks have completed.

PR #605, opened 2017-03-09
Db 1760 Deleting published submissions

alburde1:
Submissions should only be deletable if they are unpublished (updated and published submissions are not deletable)

### DIFF
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -192,7 +192,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         """ Deletes all data associated with the specified submission
         NOTE: THERE IS NO WAY TO UNDO THIS """
 
-        if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
+        if submission.publish_status_id != PUBLISH_STATUS_DICT['unpublished']:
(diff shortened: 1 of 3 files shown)


dataactbroker/fileRoutes.py
```diff
@@ -192,7 +192,7 @@ def delete_submission(submission):
         """ Deletes all data associated with the specified submission
         NOTE: THERE IS NO WAY TO UNDO THIS """
 
-        if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
+        if submission.publish_status_id != PUBLISH_STATUS_DICT['unpublished']:
             return JsonResponse.error(ValueError("Certified submissions cannot be deleted"), StatusCode.CLIENT_ERROR)
```
nmonga91: We may want to update this message. Can be misleading if someone sees "Updated" as the publish status and the submission status no longer shows up as "Certified" on the front-end.

nmonga91:
We should also update the API docs here: https://github.com/fedspendingtransparency/data-act-broker-backend/tree/development/dataactbroker#post-v1delete_submission to include the change that any submission that was certified at some point, cannot be deleted (i.e., any submission with a publish status of `published` or `updated`)

nmonga91:
Looks good, merging!

PR #609, opened 2017-03-13
DB-1759 I need to see my previous C12 warnings

kkuratsu:
As an agency user, I need to see my previous C12 warnings
Acceptance Criteria:
If a PIID and PAID combination exists in D1, it should show a warning if it does not exist in C
If a PIID (with null PAID) is in D1, it should show a warning if it does not exist in C

Technical changes: 
SQL rules (Change the rule back to previous commit)

### DIFF
--- a/dataactvalidator/config/sqlrules/c12_cross_file.sql
+++ b/dataactvalidator/config/sqlrules/c12_cross_file.sql
@@ -5,25 +5,16 @@ SELECT
     ap.parent_award_id
 FROM award_procurement AS ap
 WHERE ap.submission_id = {}
-    AND COALESCE(ap.federal_action_obligation, 0) <> 0
     AND ap.piid IS NOT NULL
-    AND ((ap.parent_award_id IS NULL
-          AND NOT EXISTS (
-              SELECT 1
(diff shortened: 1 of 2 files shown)


dataactvalidator/config/sqlrules/c12_cross_file.sql
```diff
@@ -5,26 +5,17 @@ SELECT
...
-    )
+    AND COALESCE(ap.federal_action_obligation, 0) <> 0
     AND NOT EXISTS (
+        SELECT 1
+        FROM award_financial AS af
+        WHERE af.submission_id = ap.submission_id
+            AND af.piid = ap.piid
+            AND af.parent_award_id IS NOT DISTINCT FROM ap.parent_award_id
+    )
+    AND NOT EXISTS(
```
nmonga91: @alburde1 was kind enough to notify me that this block is no longer required to be part of this check since we're comparing D1 to C here, and `transaction_obligated_amou` is not in D1.

kkuratsu:
Remove transaction_obligated_amou.

nmonga91:
Looks good, merging

PR #614, opened 2017-03-15
Db 1762 certified files in separate location

alburde1:
Copying certified files to the certified folder for that environment.

Technical Release Notes:
N/A

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1039,8 +1039,8 @@ class FileHandler:
         new_bucket = CONFIG_BROKER['certified_bucket']
         for job in jobs:
             old_path_sections = job.filename.split("/")
-            new_path = '{}/{}/{}/{}'.format(submission.cgac_code, submission.reporting_fiscal_period // 3,
-                                            submission.reporting_fiscal_year, old_path_sections[-1])
+            new_path = '{}/{}/{}/{}'.format(submission.cgac_code, submission.reporting_fiscal_year,
+                                            submission.reporting_fiscal_period // 3, old_path_sections[-1])
             self.s3manager.copy_file(original_bucket=original_bucket,
                                      new_bucket=new_bucket,
                                      original_path=job.filename, new_path=new_path)
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1039,7 +1039,9 @@ class FileHandler:
             old_path_sections = job.filename.split("/")
             new_path = '{}/{}/{}/{}'.format(submission.cgac_code, submission.reporting_fiscal_period // 3,
                                             submission.reporting_fiscal_year, old_path_sections[-1])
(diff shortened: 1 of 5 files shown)


alburde1:
[run tests]

alburde1:
[run tests]

alburde1:
[run tests]

dataactcore/aws/s3UrlHandler.py
```diff
@@ -132,3 +132,14 @@ def get_file_urls(self, bucket_name, path):
                 urls[file_name] = url
 
         return urls
+
+    @staticmethod
```
nmonga91: I think it's time we rename this file since it's doing more than simple URL manipulation.

dataactcore/config_example.yml
```diff
@@ -137,6 +137,7 @@ logging:
             - use_aws
             - local
             - aws_bucket
+            - certified_bucket
```
nmonga91: Do we want to put this key in as an example above in the broker section as well?

dataactbroker/handlers/fileHandler.py
```diff
@@ -1029,6 +1029,22 @@ def restart_validation(submission):
...
+        sess = GlobalDB.db().session
+        # Putting this here for now, get the uploads list
+        jobs = sess.query(Job).filter(Job.submission_id == submission.submission_id,
+                                      Job.job_type_id == JOB_TYPE_DICT['file_upload'],
+                                      Job.filename.isnot(None)).all()
+        original_bucket = CONFIG_BROKER['aws_bucket']
+        new_bucket = CONFIG_BROKER['certified_bucket']
+        for job in jobs:
+            old_path_sections = job.filename.split("/")
+            new_path = '{}/{}/{}/{}'.format(submission.cgac_code, submission.reporting_fiscal_period // 3,
```
nmonga91: This may be better to organize as [CGAC]/[Fiscal Year]/[Quarter] (i believe that's how it's outlined in JIRA as well)

alburde1: Oops, that was my bad

dataactcore/config_example.yml
alburde1: aws_bucket wasn't in there either so I just followed that, do we want to add it?

dataactcore/aws/s3UrlHandler.py
alburde1: Do we want to do that in this PR (there are other PRs this will almost definitely break) or do we want to do it first thing in the next sprint?

nmonga91: I don't think it's mandatory now to fix it but the naming should be relevant. If you can write up a ticket to ensure we at least are aware of it, that should suffice.

dataactcore/config_example.yml
nmonga91: if that's not in there then it's fine to leave it out.

nmonga91:
Looks good, merging

PR #615, opened 2017-03-15
Db 1699 submission file organization

alburde1:
Updating the keys for storing submissions in S3 (use submission_id instead of user_id)

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -932,20 +932,19 @@ class FileHandler:
     def add_generation_job_info(self, file_type_name, job=None, dates=None):
         # if job is None, that means the info being added is for detached d file generation
         sess = GlobalDB.db().session
-        user_id = g.user.user_id
+
+        if job is None:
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -969,7 +969,7 @@ def build_file_map(self, request_params, file_type_list, response_dict, upload_f
             if file_name:
                 if not self.isLocal:
                     upload_name = "{}/{}".format(
-                        g.user.user_id,
+                        submission.submission_id,
```
nmonga91: This handles updating file paths for an existing submission correct?

alburde1: No, that's what the one-time script I'm working on will do (although if someone uploads a new file in an existing submission this SHOULD work as intended as well)

nmonga91: Yeah the latter was my concern, when a user updates a submission, this'll put the new file in the right place. Looks good to me 👍

nmonga91:
Looks good, merging

PR #617, opened 2017-03-20
Db mk 1667 update submission

kimminahm:
Adding code so that existing code continues to work while allowing for users to upload new d2 files for FABS
[Dependent PR](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/479)
Acceptance Criteria
- Users can upload a replacement file in a FABS submission


dataactbroker/handlers/fileHandler.py
```diff
@@ -639,6 +639,123 @@ def upload_detached_file(self, create_credentials):
...
+        except ResponseException as e:
+            # call error route directly, status code depends on exception
+            return JsonResponse.error(e, e.status)
+        except Exception as e:
+            # unexpected exception, this is a 500 server error
+            return JsonResponse.error(e, StatusCode.INTERNAL_ERROR)
+        except:
+            return JsonResponse.error(Exception("Failed to catch exception"), StatusCode.INTERNAL_ERROR)
+
+    def update_detached_file(self, create_credentials):
```
alburde1: Why are we created an entire new function that does 99% of the same stuff? `upload_detached_files` has a check for existing submissions. You'll have to update how it works slightly but in general you can just use the existing code. If you look at the regular upload, I believe the code calls the same function whether it's an existing submission or a new one.

kimminahm: This was back from when I was first trying to figure out how to make this work. Removing now

dataactbroker/handlers/fileHandler.py
```diff
@@ -639,26 +639,54 @@ def upload_detached_file(self, create_credentials):
...
+                existing_submission = None
+                existing_submission_obj = None
+
             for request_field, submission_field in request_job_mapping.items():
                 if request_field in request_params:
                     job_data[submission_field] = request_params[request_field]
                 # all of those fields are required
-                else:
+                # unless existing_submission_id exists
+                elif 'existing_submission_id' not in request_params:
```
alburde1: Are the other fields suddenly not required? We can still pass them even if they aren't changing since we have all of them can't we?

dataactbroker/handlers/fileHandler.py
```diff
@@ -639,26 +639,54 @@ def upload_detached_file(self, create_credentials):
...
 
             # convert submission start/end dates from the request into Python date objects
-            date_format = '%d/%m/%Y'
             try:
-                job_data['reporting_start_date'] = datetime.strptime(job_data['reporting_start_date'],
-                                                                     date_format).date()
-                job_data['reporting_end_date'] = datetime.strptime(job_data['reporting_end_date'],
-                                                                   date_format).date()
+                job_data['reporting_start_date'] = formatted_start_date
+                job_data['reporting_end_date'] = formatted_end_date
```
alburde1: This doesn't really need a "try" anymore does it?

dataactbroker/handlers/fileHandler.py
```diff
@@ -673,18 +701,21 @@ def upload_detached_file(self, create_credentials):
...
+                sub_tier_agency_id = sub_tier_agency.sub_tier_agency_id
+                sub_tier_affiliation = SubmissionSubTierAffiliation(submission_id=submission.submission_id,
+                                                                    sub_tier_agency_id=sub_tier_agency_id)
+                sess.add(sub_tier_affiliation)
+                sess.commit()
 
-            # build fileNameMap to be used in creating jobs
-            self.build_file_map(request_params, ['detached_award'], response_dict, upload_files, submission)
+                # build fileNameMap to be used in creating jobs
+            self.build_file_map(request_params, ['detached_award'], response_dict, upload_files)
```
alburde1: Why was submission removed here? It's used in the function
```
                if not self.isLocal:
                    upload_name = "{}/{}".format(
                        submission.submission_id,
                        S3UrlHandler.get_timestamped_filename(file_name)
                    )
```

dataactbroker/handlers/fileHandler.py
```diff
@@ -673,18 +701,21 @@ def upload_detached_file(self, create_credentials):
...
+            if existing_submission_obj is None:
+                sub_tier_agency_id = sub_tier_agency.sub_tier_agency_id
+                sub_tier_affiliation = SubmissionSubTierAffiliation(submission_id=submission.submission_id,
+                                                                    sub_tier_agency_id=sub_tier_agency_id)
+                sess.add(sub_tier_affiliation)
+                sess.commit()
 
-            # build fileNameMap to be used in creating jobs
-            self.build_file_map(request_params, ['detached_award'], response_dict, upload_files, submission)
+                # build fileNameMap to be used in creating jobs
```
alburde1: Random new indent that unaligns the comment?

dataactbroker/handlers/fileHandler.py
```diff
@@ -639,28 +639,52 @@ def upload_detached_file(self, create_credentials):
...
+                existing_submission = None
+                existing_submission_obj = None
+
             for request_field, submission_field in request_job_mapping.items():
                 if request_field in request_params:
                     job_data[submission_field] = request_params[request_field]
                 # all of those fields are required
-                else:
+                # unless existing_submission_id exists
+                elif 'existing_submission_id' not in request_params and request_field in request_params:
```
alburde1: If it's always required, couldn't you just add `existing_submission_id` to `request_job_mapping`? Rather than having this strange combination of both?

dataactbroker/handlers/fileHandler.py
```diff
@@ -639,28 +639,52 @@ def upload_detached_file(self, create_credentials):
...
+                    raise ResponseException("Date must be provided as DD/MM/YYYY", StatusCode.CLIENT_ERROR, ValueError)
+
             # get the cgac code associated with this sub tier agency
-            sub_tier_agency = sess.query(SubTierAgency).\
-                filter_by(sub_tier_agency_code=request_params["agency_code"]).one()
-            job_data["cgac_code"] = sub_tier_agency.cgac.cgac_code
+            job_data["cgac_code"] = cgac_code
             job_data["d2_submission"] = True
 
             # convert submission start/end dates from the request into Python date objects
```
alburde1: Fix the comment? This should be above the "try" you moved, you probably don't actually even need this comment here and can just smush the lines below it up right under the d2_submission getting set

dataactbroker/handlers/fileHandler.py
kimminahm: its not required. If it is a first time submission this will not exist

dataactbroker/handlers/fileHandler.py
kimminahm: fixed

PR #618, opened 2017-03-20
Db 1770 fix c9 warnings

alburde1:
Stop using transaction_obligated_amou in this check.

Also making sure that any combination of fain/uri (including nulls) from D2 must be replicated exactly in C for this warning to not go off.

Adding a "with" to speed things up.

### DIFF
--- a/dataactvalidator/config/sqlrules/c9_award_financial.sql
+++ b/dataactvalidator/config/sqlrules/c9_award_financial.sql
@@ -1,18 +1,28 @@
-SELECT afa.row_number, afa.fain, afa.uri
-FROM award_financial_assistance AS afa
-WHERE afa.submission_id = {0}
-    AND (COALESCE(afa.federal_action_obligation, 0) <> 0
+WITH award_financial_assistance_c9_{0} AS
+    (SELECT submission_id,
+		row_number,
+		federal_action_obligation,
+		original_loan_subsidy_cost,
+		fain,
+		uri
+	FROM award_financial_assistance
+	WHERE submission_id = {0})
+SELECT
+    afa.row_number,
+    afa.fain,
+    afa.uri
+FROM award_financial_assistance_c9_{0} AS afa
+WHERE (COALESCE(afa.federal_action_obligation, 0) <> 0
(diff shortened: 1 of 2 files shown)


dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,27 +1,28 @@
-SELECT afa.row_number, afa.fain, afa.uri
-FROM award_financial_assistance AS afa
-WHERE afa.submission_id = {0}
-    AND (COALESCE(afa.federal_action_obligation, 0) <> 0
+WITH award_financial_assistance_c9_{0} AS
```
nmonga91: When I select "View" to see the file in Github, the tabs look a bit off.

dataactvalidator/config/sqlrules/c9_award_financial.sql
```diff
@@ -1,27 +1,28 @@
...
-                    ON afa.submission_id = af.submission_id
-                        AND afa.uri = af.uri
-            WHERE afa.submission_id = {0}
-        )
-    ) AND NOT EXISTS(
-	      SELECT transaction_obligated_amou
-	      FROM award_financial AS af
-	      WHERE transaction_obligated_amou IS NULL
-    );
+                        AND (afa.fain = af.fain
```
nmonga91: For `afa.fain = af.fain OR (afa.fain IS NULL AND af.fain IS NULL)`, couldn't that be replaced with `afa.fain IS NOT DISTINCE FROM af.fain`?

dataactvalidator/config/sqlrules/c9_award_financial.sql
alburde1: Looking into it

dataactvalidator/config/sqlrules/c9_award_financial.sql
alburde1: No, but I COULD replace it with `afa.fain IS NOT DISTINCT FROM af.fain` ;)

nmonga91:
looks good, merging

PR #619, opened 2017-03-22
Pass MonthString and Year instead of MonthNumber and Year

kimminahm:
Format the date IF monthly from numerical to string (eg. `01 => Jan`)


alburde1:
[run tests]

PR #627, opened 2017-03-28
Db 1763 prevent submit period certified

kkuratsu:
**Broker-backend** - Add a new route to check certified submission with same period.

[**Web-App**](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/494) - Make API request to check existing submission with the same period. Add warning message and prevent creating a new submission with the same period that already been certified.

**Acceptance Criteria**:
- If user tries to create a new submission for the same period of a submission that has already been certified, they get a warning with a link to the existing submission.
- If multiple submissions have been created for the same period, once one is submitted the rest can no longer be submitted.
- If a user tries to publish a submission for a FY and quarter that has already been certified, they see a warning message and are directed to the certified submission to update it.


kkuratsu:
It fails codeclimate for a duplicated code. Currently, it outputs the same error message for both `check_year_quarter/` and `certify_submission/`, so it looks like we have a duplicated code. 

It's best to keep a separate error message for each route instead of combine them. Having the error message of it own will allow us to easily change each individual message in the future.

PR #628, opened 2017-03-29
Db 1704 agency list order

alburde1:
Adding a priority column to the sub_tier_agency table and passing it up when listing the sub tier agencies.

Technical Release Notes:
- Adding a column to a table, will need to upgrade alembic head


dataactvalidator/scripts/loadFile.py
```diff
@@ -99,6 +99,12 @@ def load_sub_tier_agencies(file_name):
 
         # read Sub Tier Agency values from csv
         data = pd.read_csv(file_name, dtype=str)
+
+        data["PRIORITY"] = 0
```
nmonga91: Does it break if we don't initialize that column with a default value of 0?

dataactvalidator/scripts/loadFile.py
```diff
@@ -99,6 +99,12 @@ def load_sub_tier_agencies(file_name):
...
+        condition = data["FPDS DEPARTMENT ID"] == data["SUB TIER CODE"]
+        data.loc[condition, "PRIORITY"] = 1
+        data.loc[~condition, "PRIORITY"] = 2
```
nmonga91: Also do we need this line if the default on insertion is already set to 2 via the database migration?

dataactvalidator/scripts/loadFile.py
alburde1: Doesn't seem like it, removed.

dataactvalidator/scripts/loadFile.py
alburde1: I wanted to make sure in case a new one was added or the data got messed up somehow. It'll cover any accidental manual edits.

nmonga91:
Looks good, ready to merge

PR #632, opened 2017-03-30
Db 1763 prevent submit period certified

kkuratsu:
**Broker-backend** - Add a new route to check certified submission with same period.

**[Web-App]**(https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/494) - Make API request to check existing submission with the same period. Add warning message and prevent creating a new submission with the same period that already been certified.

**Acceptance Criteria:**

- If user tries to create a new submission for the same period of a submission that has already been certified, they get a warning with a link to the existing submission.
- If multiple submissions have been created for the same period, once one is submitted the rest can no longer be submitted.
- If a user tries to publish a submission for a FY and quarter that has already been certified, they see a warning message and are directed to the certified submission to update it.


dataactbroker/fileRoutes.py
```diff
@@ -260,3 +320,12 @@ def wrapped(*args, **kwargs):
                 "submission_id is required", StatusCode.CLIENT_ERROR)
         return fn(submission_id, *args, **kwargs)
     return wrapped
+
+
+def _find_existing_submissions(sess, submission_id, cgac_code, reporting_fiscal_year, reporting_fiscal_period):
+    return sess.query(Submission).filter(
```
nmonga91: This should return the most recent submission (may need to be confirmed with @kaitlin) that matches this criteria. Currently it returns the first one it finds (based on running it locally).

dataactbroker/fileRoutes.py
```diff
@@ -27,6 +27,32 @@ def add_file_routes(app, create_credentials, is_local, server_path):
     @requires_login
     def submit_files():
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
+
+        sess = GlobalDB.db().session
+
+        start_date = request.json.get('reporting_period_start_date')
```
kkuratsu: I follow how `@app.route("/v1/delete_submission/", methods=['POST'])` above retrieving parameters. Do you want me change this to use one of the two methods? We might need to change `delete_submission/` too in this case?

nmonga91: That's using a decorator that still utilizes `webargs`. It may be beneficial to utilize webargs the way other routes are using it. Not sure what the best practice is in this situation though.

alburde1:
Just wanted to note that it might make sense to make this entire function a decorator so we don't have nearly identical code in 3 different places. Or at least put as much as possible into a decorator instead of a function that seems to just create a single query.

dataactbroker/fileRoutes.py
kkuratsu: All three routes are similar, but there are slightly different params we can use.  For example, `/submit_files/` stores the state values of cgac_code, reporting_start_date, reporting_end_date, but these are not supplied when we try to update files for certified submission. We might want to reconsider adding a check on `submit_files` since we already have a check when users try to certify submission

I changed the `/v1/check_year_quarter/` to utilize `kwargs`. For the `certify_submission/` there really no benefit of using `kwargs`, adding kwargs also makes the code harder to work with. In my opinion, it's best to keep it as it is. 

I did, however, change the return error to function to avoid repeated code. Although that again not really best practice in my opinion since all three routes should be able to return different messages. It will be hard to work with in the future if we combine 3 return errors in to one. Again, I will leave it up to you @nmonga91.

dataactbroker/fileRoutes.py
```diff
@@ -265,3 +312,23 @@ def wrapped(*args, **kwargs):
                 "submission_id is required", StatusCode.CLIENT_ERROR)
         return fn(submission_id, *args, **kwargs)
     return wrapped
+
+
+def _find_existing_submissions_in_period(sess, submission_id, cgac_code, reporting_fiscal_year,
```
nmonga91: Alisa mentioned turning this into a decorator. That'll consolidate the extraction of arguments from request and the check this is doing. So instead a a call in each function, you'll just use `@check_existing_submission_in_period` or something similar. That seems like a good route to take.

dataactbroker/fileRoutes.py
nmonga91: Other decorators used on multiple routes return the same message, so that's not a concern since we know, by the message, what's actual causing it.

dataactbroker/fileRoutes.py
```diff
@@ -265,3 +315,25 @@ def wrapped(*args, **kwargs):
...
+    submission_query = sess.query(Submission).filter(
+        Submission.cgac_code == cgac_code,
+        Submission.reporting_fiscal_year == reporting_fiscal_year,
+        Submission.reporting_fiscal_period == reporting_fiscal_period,
+        Submission.publish_status_id == PUBLISH_STATUS_DICT['published'])
+    if submission_id:
+        submission_query = submission_query.filter(
+            Submission.submission_id != submission_id)
+
+    submission_query = submission_query.order_by(desc(Submission.created_at))
```
nmonga91: instead of using `desc` you can also do `Submission.created_at.desc()`

nmonga91: This is ok as is though

nmonga91:
Looks good, merging

PR #636, opened 2017-03-31
Db 1748 b rule fixes zeroes allowed

alburde1:
For TAS in file B that have zeros for each of the financial data elements (USSGL and related subtotals):
- Accept blanks for program activity name (B9) without warnings
- Accept 0000 for program activity code (B10) without warnings (note, it is ok for the 0 to be padded)
- Accept 000/0000 for object class (B11) without warnings (note, it is ok for the 0 to be padded)
- Accept D, R, or blank for D/R field regardless of whether a 3 or 4 digit object class is given (B12, B18) without warnings

Some of these already happened, but those that didn't have been updated.

Technical Release Notes:
- Updates to validation rules, need to reload them to the db

### DIFF
--- a/dataactvalidator/config/sqlrules/b11_object_class_program_activity_2.sql
+++ b/dataactvalidator/config/sqlrules/b11_object_class_program_activity_2.sql
@@ -1,7 +1,35 @@
+CREATE OR REPLACE function pg_temp.is_zero(numeric) returns integer AS $$
+BEGIN
+    perform CAST($1 AS numeric);
+    CASE WHEN $1 <> 0
+        THEN return 1;
+        ELSE return 0;
+    END CASE;
+EXCEPTION WHEN others THEN
+    return 0;
+END;
+$$ LANGUAGE plpgsql;
+
 SELECT
     op.row_number,
     op.object_class
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
-AND op.object_class IN ('000', '00', '0')
+AND op.object_class IN ('0000', '000', '00', '0')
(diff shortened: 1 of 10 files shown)


alburde1:
[run tests]

dataactvalidator/config/sqlrules/b11_object_class_program_activity_2.sql
```diff
@@ -1,7 +1,35 @@
...
+
 SELECT
     op.row_number,
     op.object_class
 FROM object_class_program_activity AS op
 WHERE op.submission_id = {}
-AND op.object_class IN ('000', '00', '0')
+AND op.object_class IN ('0000', '000', '00', '0')
+AND pg_temp.is_zero(op.deobligations_recov_by_pro_cpe) + pg_temp.is_zero(op.gross_outlay_amount_by_pro_cpe) +
+    pg_temp.is_zero(op.gross_outlay_amount_by_pro_fyb) + pg_temp.is_zero(op.gross_outlays_delivered_or_cpe) +
```
nmonga91: Is there a reason `COALESCE` wasn't enough here? Can these fields have non-numeric strings?

alburde1: You mean instead of doing `is_zero`? They can potentially have negative numbers I think. Adding together the positives and negatives, in rare cases, might result in 0.

nmonga91: Gotcha, makes sense

nmonga91:
Looks good, ready to merge

PR #637, opened 2017-03-31
DB 1792 correct end date for non-d2 submissions

kimminahm:
Wrote a script to fix the end date of all non-d2 submissions


nmonga91:
This should stay in this specific branch and _not_ merged into dev since it's a one time script.

PR #638, opened 2017-03-31
 Hotfix: Submission dashboard performance

nmonga91:
Updating logic for list_submissions to increase performance by only serializing on set being returned and sorting via the query object.


dataactbroker/handlers/fileHandler.py
```diff
@@ -1262,73 +1262,76 @@ def list_submissions(page, limit, certified, sort='modified', order='desc'):
...
                                     Submission.user_id == g.user.user_id))
     if certified != 'mixed':
         if certified == 'true':
-            query = query.filter(Submission.publish_status_id == PUBLISH_STATUS_DICT['published'])
+            query = query.filter(Submission.publish_status_id != PUBLISH_STATUS_DICT['unpublished'])
         else:
-            query = query.filter(Submission.publish_status_id != PUBLISH_STATUS_DICT['published'])
+            query = query.filter(Submission.publish_status_id == PUBLISH_STATUS_DICT['unpublished'])
+
+    print(query)
```
alburde1: We probably don't want to keep printing this

dataactbroker/handlers/fileHandler.py
```diff
@@ -1262,73 +1262,76 @@ def list_submissions(page, limit, certified, sort='modified', order='desc'):
...
+    total_submissions = query.count()
 
     options = {
-        'modified': 'last_modified',
-        'reporting': 'reporting_start_date',
-        'status': 'status',
-        'agency': 'agency'
+        'modified': {'model': Submission, 'col': 'updated_at'},
+        'reporting': {'model': Submission, 'col': 'reporting_start_date'},
+        # 'status': 'status',
```
alburde1: We removing this sort? If so, do we really need this commented out still? Could just delete it.

tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -55,14 +55,14 @@ def test_list_submissions_sort_success(database, job_constants, monkeypatch):
     assert result['total'] == 5
     sub = result['submissions'][0]
     for subit in result['submissions']:
-        assert subit['user']['name'] <= sub['user']['name']
+        assert subit['user']['name'] >= sub['user']['name']
```
alburde1: Why did these tests flip? Were they wrong to begin with? How were they passing?

nmonga91: Yeah they didn't seem correct. It was testing `submitted_by` in ascending order so `sub['user']['name']` in this case is `Barry Allen` which is actually less than `Oliver Queen` when doing such a comparison with strings. I'm not sure how it was working before. Strange that I was only seeing this failure now.

PR #645, opened 2017-04-06
DB 1792 accurate submission dates

kimminahm:
Updates the end date of a submission to use the last date of the month rather than the default first date


kimminahm:
https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/481/files

The hotfix will need to be removed. The script needs to be run to update all the existing dates

PR #646, opened 2017-04-07
Hot Fix: PA name to case insensitive

kkuratsu:
**Technical Release Notes**:
- New sql rules
- Lowering program activity case on loading

Store `program_activity_name`as lowercase and just lowercase the value in the agency's file.


nmonga91:
Does `program_activity_name` already get stored as lower case when we load the CSV?

kkuratsu:
Yes, the program_activity_name is stored as lowercase so we don't have to lowercase thousands of time in one validation. I already uploaded the lowercase PA table to S3.

nmonga91:
The concern is that it's lowercased manually in the CSV vs doing it programmatically in the loader. Is that the best way? It might be better to update the loader to lowercase that column to avoid human error when updating the CSV.

kkuratsu:
I think we should actually make the entire table case insensitive instead of pick a column to lowercase in the loader. Currently we have to manually update PA table, so I think it's best to do this way now and discuss how we want to handle all tables that need to lowercase. We then should just do it in the loader for all tables.  

Right now I don't have that information which column require to have mixed case letters. If the answer is none. We should go back to all places that we lowercase value.

What is yr thought on this?

nmonga91:
In program_activity, only the `program_activity_name` has non-numerical characters so lowercasing the entire table wouldn't do much. That column would be the only one that needs to be lowercased. 

If other tables need to be lowercased (a separate effort), that should be done wherever the data is being loaded instead of in the CSV directly.

dataactvalidator/scripts/loadFile.py
```diff
@@ -177,6 +177,8 @@ def load_program_activity(filename):
             {"program_activity_code": {"pad_to_length": 4}, "agency_id": {"pad_to_length": 3},
              "allocation_transfer_id": {"pad_to_length": 3, "keep_null": True}, "account_number": {"pad_to_length": 4}}
         )
+        # Lowercase Program Activity Name
+        data['program_activity_name'] = data['program_activity_name'].apply(lambda x: x.lower())
```
nmonga91: Just curious, does the `apply` not update `data['program_activity_name']` in place? It returns a new object?

kkuratsu:
It's actually update in place.

PR #651, opened 2017-04-12
Hotfix: Update the validation to also include the 2016 

kkuratsu:
Hotfix: Update the validation to also include the 2016 values in the validation.


dataactvalidator/config/sqlrules/b9_award_financial.sql
```diff
@@ -1,18 +1,14 @@
 WITH award_financial_b9_{0} AS
 	(SELECT submission_id,
 		row_number,
-		beginning_period_of_availa,
```
nmonga91: I thought these were already removed, why are these showing up as being removed here?

dataactvalidator/config/sqlrules/b9_award_financial.sql
```diff
@@ -28,7 +24,8 @@ WHERE af.submission_id = {0}
                 AND af.main_account_code IS NOT DISTINCT FROM pa.account_number
                 AND LOWER(af.program_activity_name) IS NOT DISTINCT FROM pa.program_activity_name
                 AND af.program_activity_code IS NOT DISTINCT FROM pa.program_activity_code
-                AND CAST(pa.budget_year as integer) = (SELECT reporting_fiscal_year
-                                                            FROM submission
-                                                            WHERE submission_id = af.submission_id))
-	);
\ No newline at end of file
+                AND (CAST(pa.budget_year as integer) = 2016
```
nmonga91: You could also do something like:
```
CAST(pa.budget_year as integer) in (2016, (SELECT reporting_fiscal_year
						  FROM submission
						  WHERE submission_id = af.submission_id))
```

Not sure if reducing to use a single CAST for `budget_year` would affect performance, but this is just another option to checking multiple years.

tests/unit/dataactvalidator/test_b9_award_financial.py
```diff
@@ -44,45 +44,51 @@ def test_success_null(database):
...
+                                 main_account_code='test2', program_activity_name='test2',
+                                 program_activity_code='test2')
 
     pa_1 = ProgramActivityFactory(budget_year=2016, agency_id='test', allocation_transfer_id='test',
                                   account_number='test', program_activity_name='test', program_activity_code='test')
 
     pa_2 = ProgramActivityFactory(budget_year=2017, agency_id='test2', allocation_transfer_id='test2',
                                   account_number='test2', program_activity_name='test2', program_activity_code='test2')
 
-    submission = SubmissionFactory(submission_id='1', reporting_fiscal_year='2016')
```
nmonga91: Wouldn't we want to include tests for 2016 since we're adding that back in as a valid budget_year?

dataactvalidator/config/sqlrules/b9_award_financial.sql
kkuratsu: Yes it should be removed already, I forgot to take remove that from select statement. So I clean this up with this hotfix.

tests/unit/dataactvalidator/test_b9_award_financial.py
kkuratsu: If they submit on `reporting_fiscal_year 2017` and the PA name doesn't exist in the 2017 PA name, it should look for PA name in 2016 instead. 

So we want to test that PA name (test) for `butget_year=2016` which should pass even for the submission `reporting_fiscal_year 2017`

PR #654, opened 2017-04-12
DB 1810 restrict submission by agency permissions

kimminahm:
Restrict the agency list to agencies that user has submit/create permissions


kimminahm:
Need to add more generic message

kimminahm:
I talked with kaitlin and she said that removing the agency from the list was enough that way it would prevent users beforehand. On updating submissions, there is already an error message when a user tries to edit a submission that is not theirs

PR #656, opened 2017-04-13
Db 1819 allow large files to generate

alburde1:
Hotfix to stream D files non-locally instead of saving to memory. Fixes the Memory Error we kept getting when DoD tried to generate their D1 file.

Also uses threads to continue the streaming so we don't make MetroStar wait too long on their callback response. If we do, it sends it again and we end up downloading multiple 1GB+ files

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -502,7 +502,7 @@ class FileHandler:
                 if response.status_code != 200:
                     # Could not download the file, return False
                     return False
-                # write to file
+                # write (stream) to file
                 response.encoding = "utf-8"
                 for chunk in response.iter_content(chunk_size=1024):
                     if chunk:
@@ -540,9 +540,13 @@ class FileHandler:
                 job.error_message = "A problem occurred receiving data from {}".format(source)
 
                 raise ResponseException(job.error_message, StatusCode.CLIENT_ERROR)
-            # lines = get_lines_from_csv(full_file_path)
-            #
-            # write_csv(timestamped_name, upload_name, is_local, lines[0], lines[1:])
+
+            # we're streaming non-locally but locally we still need to write as a csv
+            # because copyfile doesn't do it
+            if self.isLocal:
+                lines = get_lines_from_csv(full_file_path)
+
+                write_csv(timestamped_name, upload_name, is_local, lines[0], lines[1:])
 
             logger.debug('Marking job id of %s', job_id)
             mark_job_status(job_id, "finished")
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -1,11 +1,14 @@
 import os
+import boto
```
nmonga91: Is there a reason we're not using Boto3? just `boto` is version 2.x.x, it'd be best to use the latest if possible

dataactbroker/handlers/fileHandler.py
```diff
@@ -488,18 +491,25 @@ def call_d_file_api(self, file_type_name, cgac_code, start_date, end_date, job,
...
         """ Download a file locally from the specified URL, returns True if successful """
         if not self.isLocal:
-            with open(local_file_path, "w") as file:
-                # get request
-                response = requests.get(file_url)
-                if response.status_code != 200:
-                    # Could not download the file, return False
-                    return False
-                # write to file
+            bucket = CONFIG_BROKER['aws_bucket']
```
nmonga91: in the S3 Handler we have, is there a method that can return the connection to the S3 bucket? if not it may be beneficial to abstract out this part of code there for reuse.

dataactbroker/handlers/fileHandler.py
```diff
@@ -488,18 +491,25 @@ def call_d_file_api(self, file_type_name, cgac_code, start_date, end_date, job,
...
+                if not response:
+                    response = requests.get(file_url, stream=True)
+                    # we only need to run this check if we haven't already
+                    if response.status_code != 200:
+                        # Could not download the file, return False
+                        return False
+                # write (stream) to file
                 response.encoding = "utf-8"
-                file.write(response.text)
+                for chunk in response.iter_content(chunk_size=1024):
```
nmonga91: Can you put something in the comments why the chunk size is set to 1024? Also, we may want to make it a constant variable (so just an all caps variable below the imports) so it's easily manipulated.

dataactbroker/handlers/fileHandler.py
alburde1: I just used the one that we use in the most places. Will it continue working with everything or will we have to run another test if I update?

nmonga91: We should test one more time just to make sure there aren't any issues, especially when switching versions since we don't know exactly what is different

alburde1: Update: now it's inside s3Handler so I'm just using the one already in there.

nmonga91:
Looks good, merging

PR #660, opened 2017-04-18
Hotfix: Closing extra DB connections on D files

alburde1:
Closing extra DB connections in the threaded functions

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -552,6 +552,9 @@ class FileHandler:
 
             logger.debug('Marking job id of %s', job_id)
             mark_job_status(job_id, "finished")
+            # only need to close it if we threaded it in the first place
+            if not self.isLocal:
+                GlobalDB.close()
             return {"message": "Success", "file_name": timestamped_name}
         except Exception as e:
             logger.exception('Exception caught => %s', e)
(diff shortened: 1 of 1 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -552,6 +552,9 @@ def load_d_file(self, url, upload_name, timestamped_name, job_id, is_local, resp
 
             logger.debug('Marking job id of %s', job_id)
             mark_job_status(job_id, "finished")
+            # only need to close it if we threaded it in the first place
+            if not self.isLocal:
```
nmonga91: You could put this in a finally block for the try/except and have a single point of closure instead of duplicating the check in 2 places. Also, I don't think we need to close only if it's non-local. If we close regardless of whether or not we threaded, any subsequent access to GlobalDB would just grab another session.

alburde1: The reason I didn't put it in a "finally" is because both the try and the except raise/return something, which means the code wouldn't get to the finally right? I can remove the "if" check though

nmonga91: The finally block will run regardless of the return statement or the exception in the preceding blocks. Here's the documentation that I'm referencing: https://docs.python.org/3/reference/compound_stmts.html#the-try-statement

alburde1: I have learned a lot, thank you

PR #669, opened 2017-04-20
Db 1695 totalfundingamount derivation

alburde1:
add total_funding_amount column to PublishedAwardFinancialAssistance table and create calculations to insert the value on publish of a FABS! submission.

Technical Release Notes:
- Updated PublishedAwardFinancialAssistance table, new alembic head


dataactcore/migrations/versions/403d00de3dce_.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
```
nmonga91: We should rename these merged files to be something like "403d00de3dce_merge_[rev1]_[rev2].py" so it eases traceability

PR #675, opened 2017-04-25
Removing job manager

nmonga91:
Moving job manager functionality for File E and F generation into the broker, which will now use threading to start the generation of the files.

Checklist:
- [x] Documentation updated
- [x] Config updated

Technical Release Notes:
- No longer using Celery & RabbitMQ for asynchronous job execution

### DIFF
--- a/dataactbroker/handlers/fileGenerationHandler.py
+++ b/dataactbroker/handlers/fileGenerationHandler.py
@@ -1,15 +1,10 @@
 from contextlib import contextmanager
 import logging
 
-from celery import Celery
-from celery.exceptions import MaxRetriesExceededError
 from flask import Flask
-import requests
 
-from dataactcore.config import CONFIG_DB, CONFIG_SERVICES, CONFIG_JOB_QUEUE
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.interfaces.function_bag import mark_job_status
-from dataactcore.logging import configure_logging
 from dataactcore.models.jobModels import Job
 from dataactcore.models.stagingModels import AwardFinancialAssistance, AwardProcurement
 from dataactcore.models.domainModels import ExecutiveCompensation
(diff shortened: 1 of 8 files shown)


nmonga91:
[run tests]

alburde1:
Not reviewing yet, just a question, if we no longer need celery or rabbit (do we? I haven't looked at the changes but it says "removing") wouldn't that be a technical change for the description of this PR? "No longer need to run celery" lol (pretty big one for me)

nmonga91:
This PR didn't actually remove Celery and RabbitMQ since that would have to be done manually and separately. But it'll still be good to put the release notes here as well. Flake issue resolved as well but I will need to do another update to remove any necessary documentation we have in the setup docs involving Celery & RabbitMQ.

alburde1:
Build still broken :)

So just for clarification, after this PR do we still need to be running celery and rabbit when we work or no?

nmonga91:
[run tests]

nmonga91:
Build issue was related to configs for unit tests. And yes, that's correct. Celery and rabbit won't be used going further in any way.

dataactbroker/handlers/fileGenerationHandler.py
```diff
@@ -123,8 +82,3 @@ def generate_e_file(task, submission_id, job_id, timestamped_name, upload_file_n
         session.commit()
 
         write_csv(timestamped_name, upload_file_name, is_local, fileE.Row._fields, rows)
-
-
-if __name__ in ['__main__', 'jobQueue']:
-    configure_logging()
-    celery_app.conf.update(BROKER_URL=broker_url('localhost'))
```
alburde1: Do we still have to configure_logging if it's `__main__` or is this never called individually so logging is already configured?

nmonga91: So since this won't be running as its own script anymore, we don't need to call `configure_logging` since it'll be setup prior the calling any function here, which happens in `app.py`.

PR #677, opened 2017-04-28
Db 1681 fabs validate zip codes

alburde1:
Validate whether zip codes provided for FABS! files are real zip codes or not.

Includes a script to read in USPS data from S3 (or local folders) to create a table of zips (it's SO BIG). The script will need to be run before the validation works 

Technical Release Notes:
- Created a new table and updated a couple old ones
- New SQL rules
- New script that needs to be run using `python readZips.py ` from the dataactcore/scripts folder (for local testing purposes, the content in the sample file SHOULD be sufficient)
- updated config example (local users will need to update their config)


dataactcore/scripts/readZips.py
```diff
@@ -0,0 +1,136 @@
...
+        base_path = os.path.join(CONFIG_BROKER["path"], "dataactvalidator", "config", CONFIG_BROKER["zip_folder"])
+        file_list = [f for f in os.listdir(base_path)]
+        for file in file_list:
+            parse_zip4_file(open(os.path.join(base_path, file)), sess)
+
+    logger.info("Zipcode script complete")
+
+
+if __name__ == '__main__':
+    with create_app().app_context():
```
nmonga91: Since this script will be run on it's own, do we need to call `configure_logging`?

alburde1: This is probably a valid concern, I'll go add that

nmonga91:
Looks good, merging

PR #678, opened 2017-05-01
Db 1689 Active CFDA Program

kkuratsu:
**Acceptance Criteria:**
- The CFDA program information is loaded in from the CFDA ftp site at ftp://ftp.cfda.gov/usaspending/
- The CFDA program information is updated in the broker weekly
- CFDA_Number is a required field for all submissions and cannot be blank.
- CFDA_Number must be in ##.### format, where # represents a number from 0 to 9. (DB-1697 CFDA validations are forwards compatible with CFDA numbers that are XX.XXX or XXX.XXXX)
- For new assistance awards (ActionType = A), the CFDA_Number must be active as of the ActionDate. This does not apply to correction records (those with CorrectionLateDeleteIndicator = C).
- For existing assistance awards (ActionType = B, C, or D), the CFDA_Number need not be active as of the ActionDate, though a warning will be produced in this case. This does not apply to correction records (those with CorrectionLateDeleteIndicator = C).

**Technical changes:**
- Load cfda program information from ftp. New script that needs to be run using `python load_cfda.py `from the dataactcore/scripts folder.
- Add cfda_program database migration (needs to run `alembic upgrade head`)
- Created a new table
- New SQL rules


dataactvalidator/scripts/load_cfda.py
```diff
@@ -0,0 +1,16 @@
...
+ftp.cwd('usaspending')
+
+data = []
+ftp.dir('-t', data.append)
+file_listing = data[0]
+file_parts = file_listing.split(' ')
+file_name = file_parts[-1]
+print('Loading ' + file_name)
+
+ftp.retrbinary('RETR ' + file_name, open('dataactvalidator/config/cfda_program.csv', 'wb').write)
```
nmonga91: Here, we may want to consider streaming directly to the database versus downloading the data, writing to the file, and then processing that new file. 

This could potentially take up a lot of storage as well so if we decide to stay with this current process, the file should be deleted once it's been loaded into the database.

nmonga91:
As discussed previously, CDFA_title and CDFA_number are FABS derivations so they should be incorporated into the derivation architecture put in place by @alburde1

dataactvalidator/config/sqlrules/d37_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM detached_award_financial_assistance AS dafa
+WHERE submission_id = {0}
+    AND dafa.action_type = 'A'
+    AND ((dafa.correction_late_delete_ind != 'C')
+        or (dafa.correction_late_delete_ind is null)
+    )
+    AND dafa.row_number IN (
+        SELECT DISTINCT dafa.row_number
+        FROM detached_award_financial_assistance AS dafa
+            JOIN cfda_program AS cfda
```
nmonga91: How large can the `cdfa_program` table get? The join here could be expensive based on the size of the tables being joined. Something to be aware of.

dataactvalidator/config/sqlrules/d37_detached_award_financial_assistance_2.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM detached_award_financial_assistance AS dafa
+WHERE submission_id = {0}
+    AND dafa.action_type IN ('B', 'C', 'D')
+    AND ((dafa.correction_late_delete_ind != 'C')
+        or (dafa.correction_late_delete_ind is null)
+    )
+    AND dafa.row_number IN (
+        SELECT DISTINCT dafa.row_number
+        FROM detached_award_financial_assistance AS dafa
+            JOIN cfda_program AS cfda
```
nmonga91: Same concern as above with the JOIN

dataactvalidator/scripts/load_cfda.py
kkuratsu: The file is actually save as  `cfda_program.csv` to /dataactvalidator/config directory. The new file will replace the old `cfda_program.csv`. We don't need to delete the old file since it saves as a replacement of the old file. I did it that way since we have all other data tables load to this directory prior to load to the database.  It's better to be fully consistent? 

Let me know which one you prefer.

dataactvalidator/config/sqlrules/d37_detached_award_financial_assistance_1.sql
kkuratsu: The cfda_program table is not big, it has 2908 lines (compare to program_activity which has 19931 lines). I did add  'SELECT DISTINCT dafa.row_number' as @alburde1 recommended for all `JOIN` to speed up the validation.

dataactvalidator/config/sqlrules/d37_detached_award_financial_assistance_2.sql
kkuratsu: The cfda_program table is not big, it has 2908 lines (compare to program_activity which has 19931 lines). I did add  'SELECT DISTINCT dafa.row_number' as @alburde1 recommended for all `JOIN` to speed up the validation.

dataactvalidator/scripts/load_cfda.py
nmonga91: The files are loaded there only for local data loads. The data loads pull files from S3 when running on Jenkins.

alburde1:
Just a quick note, it looks like CDFA_number IS NOT a derivation. It's required for all fields. CDFA_title is the part that's derived.

PR #679, opened 2017-05-01
Db mk 1806 file upload check

kimminahm:
Added routes to reset the submission state

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1113,6 +1113,28 @@ class FileHandler:
 
         return JsonResponse.create(StatusCode.OK, {"message": "Success"})
 
+    def fail_validation(submission):
+        # update all validation jobs to "ready"
+        sess = GlobalDB.db().session
+        initial_file_types = [FILE_TYPE_DICT['appropriations'], FILE_TYPE_DICT['program_activity'],
+                              FILE_TYPE_DICT['award_financial']]
+
+        jobs = sess.query(Job).filter(Job.submission_id == submission.submission_id).all()
+
+        # set all jobs to their initial status of "waiting"
+        for job in jobs:
+            job.job_status_id = JOB_STATUS_DICT['waiting']
+
+        # update upload jobs to "running", only for files A, B, and C
+        upload_jobs = [job for job in jobs if job.job_type_id in [JOB_TYPE_DICT['file_upload']] and
+                       job.file_type_id in initial_file_types]
+
+        for job in upload_jobs:
+            job.job_status_id = JOB_STATUS_DICT['failed']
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -1113,6 +1113,29 @@ def restart_validation(submission):
...
+                              FILE_TYPE_DICT['award_financial']]
+
+
+        jobs = sess.query(Job).filter(Job.submission_id == upload_id).all()
+
+        # set all jobs to their initial status of "waiting"
+        for job in jobs:
+            job.job_status_id = JOB_STATUS_DICT['waiting']
+
+        # update upload jobs to "running", only for files A, B, and C
```
alburde1: 1) This is setting it to "failed" not "running"
2) Why are we setting them to "failed"? I think we can delete if it's in "waiting" can't we? Is this something I need to see the front end to understand?
3) assuming we need to set them to "failed" why do we have to set the other jobs to "waiting" at all? If the uploads are still running then the rest shouldn't be running? If we're trying to invalidate purely for deletion purposes, couldn't we just filter all running jobs and set them to failed? Or just set all jobs to "failed" since they're about to vanish anyway?
4) why do we need to filter `initial_file_types` out at all?

kimminahm: updated to set all the jobs associated with the submission to be set to failure so the submission can be deleted.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1113,6 +1113,18 @@ def restart_validation(submission):
 
         return JsonResponse.create(StatusCode.OK, {"message": "Success"})
 
+    def fail_validation(self, upload_id):
+        # update all validation jobs to "ready"
```
alburde1: This comment is still a lie.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1113,6 +1113,18 @@ def restart_validation(submission):
...
+        sess = GlobalDB.db().session
+
+        jobs = sess.query(Job).filter(Job.submission_id == upload_id).all()
```
nmonga91: You can use `one_or_none()` here. If a user passes in an `upload_id` that doesn't correspond to any upload job in the database (since they can bypass the frontend by calling the API directly), an appropriate response and error code should be returned for that kind of error.

alburde1: @nmonga91 how is that going to get all the jobs that need to be set to failed though?

nmonga91: I clarified with Minahm. Upload ID = submission ID so this isn't applicable. Clearer naming was requested.

PR #680, opened 2017-05-02
db1813- Refactor CSV file handling from S3 to remove streaming

kkuratsu:
**Acceptance Criteria:**
switch reading from S3 from custom byte streaming logic to an existing library, like boto

**Technical Changes:**
- New csvReader file (Read entire file to temp file on local filesystem, then use local reader)


dataactvalidator/filestreaming/csvReader.py
```diff
@@ -67,12 +83,22 @@ def get_writer(bucket_name, filename, header, is_local, region=None):
...
         if region is None:
             region = CONFIG_BROKER["aws_region"]
         return CsvS3Writer(region, bucket_name, filename, header)
 
+    def _transfer_s3_file_to_local(self, bucket, filename):
+        (file, file_path) = tempfile.mkstemp()
+        file.close()
+
+        s3 = boto3.client('s3')
+        s3.download_file(bucket, filename, file_path)
```
nmonga91: Downloading the file locally and reading from there could be a potential risk with storage and memory. If this is the fastest route to take, cleanup of the downloaded files will need to happen.

Also, will this be less memory intensive as the byte-streaming that was happening previously? If we're reading in extremely large files, it may have an adverse impact. It should be tested with a realistic file set.

kkuratsu:
Note: Unlike TemporaryFile(), the user of mkstemp() is responsible for deleting the temporary file when done with it.

https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp

dataactvalidator/filestreaming/csvReader.py
```diff
@@ -67,12 +83,24 @@ def get_writer(bucket_name, filename, header, is_local, region=None):
...
             return CsvLocalWriter(filename, header)
         if region is None:
             region = CONFIG_BROKER["aws_region"]
         return CsvS3Writer(region, bucket_name, filename, header)
 
+    def _transfer_s3_file_to_local(self, bucket, filename):
+        (file, file_path) = tempfile.mkstemp()
+        file.close()
+        os.close(file_path)
+        os.remove(file_path)
```
nmonga91: If we're closing and removing the temp file before writing to it, would it still be used? I would think it would just create a new file with `file_path` when calling `open`. Wouldn't we want to delete this after we're done reading from it for the validations, etc.?

kkuratsu: mkstemp() deletes the temporary file when done with reading it. `_transfer_s3_file_to_local` is called on line 40.

dataactvalidator/filestreaming/csvReader.py
kkuratsu: The temp file is now removed after reading.

nmonga91:
Looks good, merging

PR #681, opened 2017-05-02
Bug fixes for zip reading parsing

alburde1:
Fixing a couple bugs:

- reading "all" files locally on a mac for the zip loader script would try to read hidden files as well (that's bad)
- padding congressional district no with leading 0s because it's either 2 letters or 2 numbers and the numbers get truncated sometimes.


nmonga91:
Looks good, merging

PR #686, opened 2017-05-05
Db 1846 display certified date

alburde1:
This PR also incorporates DB-1874.

- "Last Updated Date" for both sorting and display is based on a combination of the submission updated_at date and all related job updated_at dates
- We now pass the raw date including timestamp to the frontend so it can adjust for local time on last updated and certified dates.
- We also pass the last certified date

- [x] Must be merged before but very close to https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/543 (date displays will be off until frontend is merged)
- [x] Reviewed by backend

Technical Release Notes:
- Added an alembic upgrade (new view)

Side note: We may want to figure out how to put reflected views in classes at some point if we use a lot of them, but for now I think this is fine since it's used in one location.


dataactbroker/handlers/fileHandler.py
```diff
@@ -1343,6 +1344,10 @@ def list_submissions(page, limit, certified, sort='modified', order='desc'):
     """ List submission based on current page and amount to display. If provided, filter based on
     certification status """
     sess = GlobalDB.db().session
+    db = GlobalDB.db()
+    metadata = MetaData(bind=db.engine)
+
+    submission_updated_view = Table('submission_updated_at_view', metadata, autoload=True, info=dict(is_view=True))
```
nmonga91: When dealing with views in SQL Alchemy, can we query it the same way as the other models?

alburde1: Unfortunately no, that was part of the reason I had to figure this nonsense out. The problem is that, with the other ones, we basically create an object that automatically reflects the table and organizes the columns nicely when we create a model in the appropriate file. We can't make a model for a view, we have to just create it manually in the alembic migration, which means there's no quick-access method of doing it. We could turn this whole process into a class that we can call (which would still involve all this exact same stuff but in a different file/location), but I figured if it's just being done once for now we probably don't need to go that deep.

nmonga91: The latter part is what I was thinking as well. Moving this in as a separate file (alongside the models) but a file just for views that we're using. Each view would have it's own class, etc. It'll help with setting up the architecture for future enhancements since it's definitely a useful feature.

dataactcore/models/views.py
```diff
@@ -0,0 +1,16 @@
+from sqlalchemy import MetaData, Table
+from dataactcore.interfaces.db import GlobalDB
+
+
+class SubmissionUpdatedView:
```
nmonga91: General question: does it matter if we instantiate the view object versus making the `__init__` into another, static function to return the view? Wondering out of curiosity since some people in the python community say don't use static functions, some say do.

alburde1: Not that I'm aware of, I just grabbed other classes and copied them (the models I figured were very slightly different because they're used by alembic directly). Do you have a preference?

nmonga91: I personally don't. Just wasn't sure what the better route it. We can leave it the way it is for now and update if we come across something that clearly defines if one of those is best practice.

nmonga91:
Looks good, ready to merge.

PR #687, opened 2017-05-08
Db 1813 - Refactor csv file handling from S3 to remove streaming.

kkuratsu:
Refactor csv file handling from S3 to remove streaming.

**A/C:**
- Switch reading from S3 from custom byte streaming logic to an existing library, like boto


nmonga91:
Was this tested on sandbox? If so, were you able to see the file be created, and then deleted once it was done being used?

kkuratsu:
Yes, it being tested multiple times on sandbox. The file was created and then deleted once it done. The problem that it wasn't working on sandbox before is that we have to pass the `region` as well.

nmonga91:
Looks good, merging

PR #688, opened 2017-05-09
Db 1689 cfda program

kkuratsu:
**Acceptance Criteria:**

- The CFDA program information is loaded in from the CFDA ftp site at ftp://ftp.cfda.gov/usaspending/
- The CFDA program information is updated in the broker weekly
- CFDA_Number is a required field for all submissions and cannot be blank. (Error)
- CFDA_Number must be in ##.### format, where # represents a number from 0 to 9. (Error)
- For new assistance awards (ActionType = A), the CFDA_Number must be active as of the ActionDate. This does not apply to correction records (those with CorrectionLateDeleteIndicator = C). (Error)
- For existing assistance awards (ActionType = B, C, or D), the CFDA_Number need not be active as of the ActionDate, though a warning will be produced in this case. This does not apply to correction records (those with CorrectionLateDeleteIndicator = C). (Warning)

**Technical changes:**

- Load cfda program information from ftp. New script that needs to be run using `python datactvalidator/filestreaming/load_cfda.py` from the root folder.
- Add cfda_program database migration (needs to run alembic upgrade head)
- Created a new table
- New SQL rules


dataactvalidator/scripts/load_cfda.py
```diff
@@ -0,0 +1,16 @@
+from ftplib import FTP
+
+ftp = FTP('ftp.cfda.gov')      # connect to host, default port
+ftp.login()                    # user anonymous, passwd anonymous@
```
nmonga91: We don't specify that username or password anywhere for the FTP connection correct? 

Also, comments on the same line can get messy. It's best to place the comments on the preceding lines.

dataactvalidator/scripts/load_cfda.py
```diff
@@ -0,0 +1,16 @@
...
+ftp.cwd('usaspending')
+
+data = []
+ftp.dir('-t', data.append)
+file_listing = data[0]
```
nmonga91: Can you insert some comments here (and further down) indicating the importance of the certain indices that you're referencing? 

Also some comments around the different steps being taken to get the file in question would be helpful for someone coming into this file later on.

nmonga91:
One concern is that we're constantly updating the `cfda_program.csv` locally. To put that in context, when this is deployed on Jenkins, we'll spin up a Docker container (temporary workspace) and run the script. Once it's done, the container gets shut down and we lose the updated file. 

We need to also add it to the S3 bucket with the other domain value CSVs so that we have a up-to-date point of truth.

kkuratsu:
The comments is added to `load_cfda.py` on the preceding lines. The ftp://ftp.cfda.gov/usaspending/ is an anonymous FTP, the archive site allow general access. We don't need to specify username or password.

How should we proceed with adding this to S3 bucket? Should this be done after it got merged?

nmonga91:
Adding to S3 should be programmatic so we don't have to update it manually. It should be part of `load_cfda.py`

kkuratsu:
Transfering file from FTP to s3 is added to `load_cfda.py.` Tested uploading cfda file to `gtas-sf133-dev`. The file is uploaded to the bucket successfully.

dataactvalidator/scripts/load_cfda.py
```diff
@@ -0,0 +1,51 @@
...
+        s3.Bucket(CONFIG_BROKER['sf_133_bucket']).put_object(Key='cfda_program.csv', Body=data)
+
+        print('Loading file to S3 completed')
+    else:
+        # download file
+        ftp.retrbinary('RETR ' + file_name, open('dataactvalidator/config/cfda_program.csv', 'wb').write)
+        print('Loading file completed')
+
+    ftp.quit()
+load_cfda()
```
nmonga91: here when running the main function of the script, it should follow this example: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/dataactvalidator/scripts/loadTas.py#L124-L126

dataactvalidator/scripts/load_cfda.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # get the last string array (file name)
+    file_name = file_parts[-1]
+
+    print('Loading ' + file_name)
+
+    if CONFIG_BROKER["use_aws"]:
+        data = BytesIO()
+        # download file
+        ftp.retrbinary('RETR ' + file_name, data.write)
```
nmonga91: If it's not passing a path to a file it is supposed to write to, how does it handle the download?

nmonga91:
Were the SQL validations confirmed with Ross?

kkuratsu:
Just send another email to follow up with Ross.

dataactcore/models/domainModels.py
```diff
@@ -1,7 +1,8 @@
 from datetime import timedelta
 
 import sqlalchemy as sa
-from sqlalchemy import Column, Date, ForeignKey, Index, Integer, Numeric, Text, UniqueConstraint
+
```
nmonga91: Can we clean up the empty lines around the imports?

nmonga91:
If `dataactvalidator/config/cfda_program.csv` is present as an example (similar to program activity, object_class, etc), we should name it `example_cfda_program.csv` and may consider trimming it since it's about 21 mb at the moment.

nmonga91:
Looks good, will merge once checks are finished

PR #689, opened 2017-05-10
DB-1865: Hotfix empty d file dependency deletion

nmonga91:


dataactcore/interfaces/function_bag.py
```diff
@@ -296,7 +296,7 @@ def mark_job_status(job_id, status_name):
 
     # if status is changed to finished for the first time, check dependencies
     # and add to the job queue as necessary
-    if old_status != 'finished' and status_name == 'finished':
+    if old_status != 'finished' and status_name == 'finished' and not skip_check:
```
alburde1: Just making sure, this doesn't prevent cross validation from happening because one of its dependencies won't be "finished"

nmonga91: Cross-file will still get kicked off because the dependent job (the D file validation jobs) will be marked as finished which'll continue with the dependency check since `skip_check` by default is `False`. It'll still work normally (as it does now). 

The skip_check is primarily used for the D file upload jobs. If the API comes back with no data (empty D file), it'll skip the dependency checks for the upload job when marking it as `finished` so that it doesn't kick off the related validation job (since there's no data to validate).

alburde1: Right, but where do the validation jobs get marked as finished if they're never kicked off?

nmonga91: They get updated to `finished` here if there are no results: https://github.com/fedspendingtransparency/data-act-broker-backend/pull/689/files#diff-7507a7c0c5557b3a9aabb9c8c18c4e5bR494

And here if there is actual data: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/7e8c5df31f4c4a27a5e2785b27cf403b5ce797b4/dataactbroker/handlers/fileHandler.py#L572

PR #691, opened 2017-05-10
DB-1891 Create a route to check current page

kkuratsu:
A sub-task of DB-1858

**Acceptance Criteria:**
User gets directed to the latest step in the process when visiting any part of the submission, or already completed step of the submission

For example, cannot directly navigate to the D file generation page unless they have completed A-C file validation without critical errors

**Technical changes:**
Create a new route to check the current page

### DIFF
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -199,7 +199,6 @@ def add_file_routes(app, create_credentials, is_local, server_path):
             }
             return JsonResponse.create(StatusCode.OK, data)
 
-
     @app.route("/v1/generate_file/", methods=["POST"])
     @convert_to_submission_id
     @use_kwargs({'file_type': webargs_fields.String(
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -137,7 +137,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         sess = GlobalDB.db().session
 
         # check if submission_id exists in database
-        check_submission_id = sess.query(Job).filter(Job.submission_id == submission_id)
+        check_submission_id = sess.query(Submission).filter(Submission.submission_id == submission_id)
         if check_submission_id.count() < 1:
             data = {
                 "message": "A submission with the specified ID does not exist",
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -136,7 +136,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         sess = GlobalDB.db().session
         # /v1/reviewData/
         review_data = sess.query(Job).filter(Job.submission_id == submission_id,
-                                            Job.file_type_id.in_([6, 7]), Job.job_status_id == 4)
+                                             Job.file_type_id.in_([6, 7]), Job.job_status_id == 4)
         if review_data.count() > 0:
             data = {
                 "message": "The current progress of this submission id is on /v1/reviewData/ page.",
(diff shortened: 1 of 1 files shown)


dataactbroker/fileRoutes.py
```diff
@@ -130,6 +130,46 @@ def get_protected_files():
...
+            return "1"
+
+        # /v1/generateFiles/
+        generatefiles = sess.query(Job).filter(Job.submission_id == submission_id,
+                                               Job.file_type_id.in_([1, 2, 3]), Job.job_type_id == 2,
+                                               Job.number_of_errors == 0, Job.file_size.isnot(None))
+        if generatefiles.count() > 0:
+            return "2"
+
+        return "0"
```
nmonga91: All broker routes should return JSON responses as opposed to a string. For example: https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/dataactbroker/fileRoutes.py#L57-L61

Also, returning numbers isn't very descriptive. A string correlating to the page's name may be more descriptive/useful.

Additionally, when using numbers, if we were to add any pages in the process (might not be likely, but still possible), the route would require more updates to reorder what numbers should be returned versus just adding an additional check with a return value representing the name of the new page(s).

kkuratsu: @kimminahm Could you verify this and let me know if this is what you would like me to change to?

kimminahm: You can change it, just have it return {page: "value"}

kimminahm:
All working for me, @nmonga91 good if you are

dataactbroker/fileRoutes.py
```diff
@@ -130,6 +130,70 @@ def get_protected_files():
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
         return file_manager.get_protected_files()
 
+    @app.route("/v1/check_current_page/", methods=["GET"])
+    @convert_to_submission_id
```
nmonga91: Do we not want to check permissions for the submission in question? If a user with no permissions accesses this route, they would still know what phase the submission is in.

dataactbroker/fileRoutes.py
```diff
@@ -130,6 +130,70 @@ def get_protected_files():
...
+    def check_submission(submission_id):
+        sess = GlobalDB.db().session
+        # /v1/reviewData/
+        reviewdata = sess.query(Job).filter(Job.submission_id == submission_id,
```
nmonga91: we should be using underscoring as the naming convention for variables. So `reviewdata` should be `review_data`. It is more readable as opposed to being all one word.

dataactbroker/fileRoutes.py
kkuratsu: I'm not sure about this but currently, we are using `submission_id` which requires login. Do we want to allow no permission accesses?

dataactbroker/fileRoutes.py
```diff
@@ -130,6 +130,70 @@ def get_protected_files():
...
+                                                Job.number_of_errors == 0, Job.file_size.isnot(None))
+        if generate_files.count() > 0:
+            data = {
+                "message": "The current progress of this submission id is on /v1/generateFiles/ page.",
+                "step": "2"
+            }
+            return JsonResponse.create(StatusCode.OK, data)
+
+        data = {
+            "message": "The submission id is not found in the database.",
```
nmonga91: We should probably do this check first to avoid going through all of the other queries. Right now, this is the base case so it'll return by default if all the other queries return with 0 results, so it'll go through each query first then it'll see that a submission with that ID doesn't exist. 

Also, the message should be something like "A submission with the specified ID does not exist"

dataactbroker/fileRoutes.py
nmonga91: Requiring login doesn't ensure they have at least read only permissions. It essentially is just saying you can log into MAX and get to the broker successfully. I would suggest making this require _at least_ readonly permissions.

dataactbroker/fileRoutes.py
```diff
@@ -130,6 +130,76 @@ def get_protected_files():
...
         return file_manager.get_protected_files()
 
+    @app.route("/v1/check_current_page/", methods=["GET"])
+    @convert_to_submission_id
+    @requires_submission_perms('reader')
+    def check_submission(submission_id):
+        sess = GlobalDB.db().session
+
+        # check if submission_id exists in database
+        check_submission_id = sess.query(Job).filter(Job.submission_id == submission_id)
```
nmonga91: If we're checking for the existence of a submission, it should be checked against the `submission` table

nmonga91:
Looks good, merging once checks have completed

PR #694, opened 2017-05-11
Db 1850 D Download and messages

kimminahm:
Adding date to the response

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -785,7 +785,7 @@ class FileHandler:
             response_dict['status'] = 'invalid'
             response_dict['message'] = 'No generation job found with the specified ID'
             return JsonResponse.create(StatusCode.OK, response_dict)
-        elif upload_job.filename == null:
+        elif upload_job.filename is None:
             response_dict['status'] = 'invalid'
             response_dict['message'] = 'No generation job found with the specified ID'
             return JsonResponse.create(StatusCode.OK, response_dict)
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -785,6 +785,10 @@ def check_detached_generation(job_id):
             response_dict['status'] = 'invalid'
             response_dict['message'] = 'No generation job found with the specified ID'
             return JsonResponse.create(StatusCode.OK, response_dict)
+        elif upload_job.filename is None:
+            response_dict['status'] = 'invalid'
+            response_dict['message'] = 'No generation job found with the specified ID'
```
nmonga91: If the filename is `NONE`, we should have a different message no? Since that doesn't mean `No generation job found with the specified ID`

kimminahm: Changed

dataactbroker/handlers/fileHandler.py
```diff
@@ -782,9 +782,13 @@ def check_detached_generation(job_id):
                          'start': '', 'end': ''}
 
         if upload_job is None:
-            response_dict['status'] = 'invalid'
+            response_dict['status'] = 'Invalid'
             response_dict['message'] = 'No generation job found with the specified ID'
             return JsonResponse.create(StatusCode.OK, response_dict)
+        elif upload_job.filename is None:
+            response_dict['status'] = 'Invalid'
```
nmonga91: Is frontend doing a case-insensitive check on the status? Just asking since the it was capitalized and wanted to make sure there were no issues with that when using the response

nmonga91:
Looks good, merging after checks complete

PR #695, opened 2017-05-12
Db 1852 refactor error count calculations

nmonga91:
Resolves deadlocking issues that result from too many updates to the `Job` model

### DIFF
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -66,6 +66,8 @@ def populate_submission_error_info(submission_id):
     submission.number_of_warnings = sum_number_of_errors_for_job_list(submission_id, error_type='warning')
     sess.commit()
 
+    return submission
+
 
 def populate_job_error_info(job):
     """ Set number of errors and warnings for specified job. """
@@ -92,7 +94,6 @@ def sum_number_of_errors_for_job_list(submission_id, error_type='fatal'):
             error_sum += job.number_of_errors
         elif error_type == 'warning':
             error_sum += job.number_of_warnings
-    sess.commit()
     return error_sum
 
 
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -94,17 +94,6 @@ def sum_number_of_errors_for_job_list(submission_id, error_type='fatal'):
     return error_sum
 
 
-def check_number_of_errors_by_job_id(job_id, error_type='fatal'):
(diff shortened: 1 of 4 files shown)


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -309,7 +309,7 @@ def run_validation(self, job):
 
             error_list.write_all_row_errors(job_id)
             # Update error info for submission
-            populate_submission_error_info(submission_id)
```
alburde1: If we take this function out of this line, do we call it anywhere else other than in tests? If not, can we just cut the entire function? (submission error info is still populated with direct calls to `sum_number_of_errors_for_job_list` right?)

dataactcore/interfaces/function_bag.py
```diff
@@ -67,33 +67,35 @@ def populate_submission_error_info(submission_id):
...
     for job in jobs:
-        job_errors = check_number_of_errors_by_job_id(job.job_id, error_type)
         if error_type == 'fatal':
-            job.number_of_errors = job_errors
+            error_sum += job.number_of_errors
         elif error_type == 'warning':
-            job.number_of_warnings = job_errors
-        error_sum += job_errors
+            error_sum += job.number_of_warnings
     sess.commit()
```
nmonga91: yeah that can be removed

dataactvalidator/validation_handlers/validationManager.py
nmonga91: I reused it actually in `run_cross_validation` since it was actually duplicating the code

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -424,10 +424,7 @@ def run_cross_validation(self, job):
         error_list.write_all_row_errors(job_id)
         mark_job_status(job_id, "finished")
         logger.info('VALIDATOR_INFO: Completed run_cross_validation on submission_id: %s', submission_id)
-        submission = sess.query(Submission).filter_by(submission_id=submission_id).one()
-        # Update error info for submission
-        submission.number_of_errors = sum_number_of_errors_for_job_list(submission_id)
-        submission.number_of_warnings = sum_number_of_errors_for_job_list(submission_id, error_type="warning")
+        submission = populate_submission_error_info(job.submission_id)
```
alburde1: Why use `job.submission_id` when we have the `submission_id` variable already?

nmonga91: Didn't see that, good catch. I'll update it.

PR #696, opened 2017-05-15
db-1698-Human Readable Names for Agency Fields

kkuratsu:
**A/C:**

AwardingAgencyName, AwardingSubTierAgencyName, FundingAgencyName, and FundingSubTierAgencyName are populated based on their corresponding codes in the authoritative list of agencies at https://github.com/fedspendingtransparency/schema/blob/master/broker/authoritative_agency_list.xlsx


dataactvalidator/config/detachedAwardFieldsRaw.csv
```diff
@@ -6,8 +6,10 @@ AwardDescription,award_description,ASP: Conditional per Validation Rule,string,4
 AwardeeOrRecipientLegalEntityName,awardee_or_recipient_legal,ASP: Required,string,120,D22
```
alburde1: This should be in publishedAwardFieldsRaw.csv

dataactvalidator/config/detachedAwardFields.csv
```diff
@@ -6,8 +6,10 @@ awarddescription,award_description,FALSE,str,4000,D2,
 awardeeorrecipientlegalentityname,awardee_or_recipient_legal,TRUE,str,120,D22,
```
alburde1: This should be in publishedAwardFields.csv

dataactcore/migrations/versions/29600f7fe065_add_agency_fields_to_published.py
```diff
@@ -0,0 +1,49 @@
+"""Add agency fields to published
```
alburde1: Published what? Should clarify with full table name if possible.

alburde1: Published what? Should clarify with full table name if possible.

dataactcore/migrations/versions/29600f7fe065_add_agency_fields_to_published.py
```diff
@@ -0,0 +1,49 @@
...
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+def upgrade_data_broker():
+    op.add_column('detached_award_financial_assistance', sa.Column('awarding_agency_name', sa.Text(), nullable=True))
+    op.add_column('detached_award_financial_assistance', sa.Column('awarding_sub_tier_agency_n', sa.Text(), nullable=True))
+    op.add_column('detached_award_financial_assistance', sa.Column('funding_agency_name', sa.Text(), nullable=True))
+    op.add_column('detached_award_financial_assistance', sa.Column('funding_sub_tier_agency_na', sa.Text(), nullable=True))
```
alburde1: These shouldn't be adding to the detached one, we aren't asking them to include the title in their file, we're deriving it on submit

dataactcore/models/stagingModels.py
```diff
@@ -522,8 +522,10 @@ class DetachedAwardFinancialAssistance(Base):
     awardee_or_recipient_legal = Column(Text)
     awardee_or_recipient_uniqu = Column(Text)
     awarding_agency_code = Column(Text)
+    awarding_agency_name = Column(Text)
     awarding_office_code = Column(Text)
     awarding_sub_tier_agency_c = Column(Text)
+    awarding_sub_tier_agency_n = Column(Text)
```
alburde1: These shouldn't be adding to the detached one, we aren't asking them to include the title in their file, we're deriving it on submit

alburde1:
The build failed. I can't see why from my current Internet. I'm going to hold off reviewing until the build succeeds

dataactbroker/handlers/fileHandler.py
```diff
@@ -1573,8 +1573,60 @@ def map_generate_status(upload_job, validation_job=None):
...
+
+    sess = GlobalDB.db().session
+
     # deriving total_funding_amount
     federal_action_obligation = obj['federal_action_obligation'] or 0
     non_federal_funding_amount = obj['non_federal_funding_amount'] or 0
     obj['total_funding_amount'] = federal_action_obligation + non_federal_funding_amount
+
+    # deriving awarding agency name
+    try:
```
alburde1: Thinking about it while not focusing on sprint review, I'm wondering if we don't want to do a try/except here. It's generally bad practice to do a default "except" statement without specifying the error to begin with, but I think it might simply be cleaner/faster to use `one_or_none()` instead of `one()`. If it doesn't find the code, it'll just return `None`. This is relevant to all cases of try/except in this PR and also the one in the other PR.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1588,5 +1588,37 @@ def fabs_derivations(obj):
...
+    awarding_sub_tier_agency_name = sess.query(SubTierAgency).filter_by(
+        sub_tier_agency_code=obj['awarding_sub_tier_agency_c']).one_or_none()
+    if awarding_sub_tier_agency_name:
+        obj['awarding_sub_tier_agency_n'] = awarding_sub_tier_agency_name.sub_tier_agency_name
+    else:
+        logger.error('Awarding Sub-tier Agency Name not found for Awarding Sub-tier Agency Code %s',
+                     obj['awarding_sub_tier_agency_c'])
+
+    # deriving funding agency name
+    funding_agency_name = sess.query(CGAC).filter_by(cgac_code=obj['funding_agency_code']).one_or_none()
```
alburde1: `funding_agency_code` is optional so we probably want to put this in an if statement as well

dataactbroker/handlers/fileHandler.py
```diff
@@ -1588,5 +1588,37 @@ def fabs_derivations(obj):
...
+    if funding_agency_name:
+        obj['funding_agency_name'] = funding_agency_name.agency_name
+    else:
```
alburde1: As with above, we probably don't need this else statement because we do a validity check with rule D20 and the above `if` isn't needed either because of the above suggestion

dataactbroker/handlers/fileHandler.py
```diff
@@ -1588,5 +1588,37 @@ def fabs_derivations(obj):
...
+        logger.debug('Funding Agency Name not found for Funding Agency Code %s', obj['funding_agency_code'])
+
+    # deriving funding sub tier agency name
+    funding_sub_tier_agency_name = sess.query(SubTierAgency).filter_by(
```
alburde1: `funding_sub_tier_agency_co` is an optional field so we probably only want to search when it exists (wrap this in an if too)

dataactbroker/handlers/fileHandler.py
```diff
@@ -1588,5 +1588,37 @@ def fabs_derivations(obj):
...
+        sub_tier_agency_code=obj['funding_sub_tier_agency_co']).one_or_none()
+    if funding_sub_tier_agency_name:
+        obj['funding_sub_tier_agency_na'] = funding_sub_tier_agency_name.sub_tier_agency_name
+    else:
```
alburde1: D21 checks for validity here so we don't need the else if we wrap the entire code segment in an if statement. Same comment about the above "if" as before

PR #698, opened 2017-05-15
Db-1697 CFDA Title 

kkuratsu:
**Acceptance Criteria:**
CFDA_Title is populated based on the CFDA_Number field and the authoritative list of CFDA programs.

**Update:**
~~CFDA validations should fail for CFDA numbers that are XXX.XXXX~~ (Same as PR#688)


dataactcore/migrations/versions/5576159ed137_add_cfda_title_to_published.py
```diff
@@ -0,0 +1,37 @@
+"""Add cfda title to published
```
alburde1: Published what? Should clarify message with actual table name

dataactcore/migrations/versions/5576159ed137_add_cfda_title_to_published.py
```diff
@@ -0,0 +1,37 @@
...
+def upgrade(engine_name):
+    globals()["upgrade_%s" % engine_name]()
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+def upgrade_data_broker():
+    op.add_column('detached_award_financial_assistance', sa.Column('cfda_title', sa.Text(), nullable=True))
```
alburde1: It shouldn't be adding to the detached one, we aren't asking them to include the title in their file, we're deriving it on submit

dataactcore/models/stagingModels.py
```diff
@@ -528,6 +528,7 @@ class DetachedAwardFinancialAssistance(Base):
     business_funds_indicator = Column(Text)
     business_types = Column(Text)
     cfda_number = Column(Text)
+    cfda_title = Column(Text)
```
alburde1: It shouldn't be adding to the detached one, we aren't asking them to include the title in their file, we're deriving it on submit

dataactvalidator/config/detachedAwardFields.csv
```diff
@@ -12,6 +12,7 @@ awardmodificationamendmentnumber,award_modification_amendme,FALSE,str,25,D6,
 businessfundsindicator,business_funds_indicator,TRUE,str,3,D1,
 businesstypes,business_types,TRUE,str,3,D1,
 cfda_number,cfda_number,TRUE,str,7,D13,
+cfda_title,cfda_title,TRUE,str,50,,
```
alburde1: This line shouldn't be added to the detached config csv at all

dataactvalidator/config/detachedAwardFieldsRaw.csv
```diff
@@ -12,6 +12,7 @@ AwardModificationAmendmentNumber,award_modification_amendme,ASP: Optional,string
 BusinessFundsIndicator,business_funds_indicator,ASP: Required,string,3,D1
 BusinessTypes,business_types,ASP: Required,string,3,D1
 CFDA_Number,cfda_number,ASP: Required,string,7,D13
+CFDA_Title,cfda_title,ASP: Required,string,50,
```
alburde1: This line shouldn't be added to the detached config csv at all

tests/unit/dataactcore/factories/staging.py
```diff
@@ -239,6 +239,7 @@ class Meta:
     business_funds_indicator = fuzzy.FuzzyText()
     business_types = fuzzy.FuzzyText()
     cfda_number = fuzzy.FuzzyText()
+    cfda_title = fuzzy.FuzzyText()
```
alburde1: It shouldn't be adding to the detached one, we aren't asking them to include the title in their file, we're deriving it on submit

dataactbroker/handlers/fileHandler.py
```diff
@@ -1577,4 +1577,19 @@ def fabs_derivations(obj):
     federal_action_obligation = obj['federal_action_obligation'] or 0
     non_federal_funding_amount = obj['non_federal_funding_amount'] or 0
     obj['total_funding_amount'] = federal_action_obligation + non_federal_funding_amount
+
+    # deriving cfda_title from program_title in cfda_program table
+    sess = GlobalDB.db().session
```
alburde1: We should just put this at the top of the function because other derivations might need it and it's better practice to declare those kinds of variables at the top of the function.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1577,4 +1577,19 @@ def fabs_derivations(obj):
...
+
+    try:
+        cfda_title = sess.query(CFDAProgram).filter_by(program_number=obj['cfda_number']).one()
+    except:
+        print('CFDA Title not Found')
```
alburde1: This shouldn't be a print statement. You should use the logger.exception() or logger.debug() or logger.info() function (there's an example in this file somewhere) so it actually gets printed to our logs. Choose the one that you think is most relevant (probably exception but if you think differently then use a different one)

Also, maybe include which cfda_number it wasn't found for ("CFDA Title not found for CFDA Number X") (found shouldn't be capitalized)

PR #700, opened 2017-05-17
Hotfix - Add submission_id parameter

kkuratsu:


dataactbroker/fileRoutes.py
```diff
@@ -133,9 +133,12 @@ def get_protected_files():
     @app.route("/v1/check_current_page/", methods=["GET"])
     @convert_to_submission_id
     @requires_submission_perms('reader')
-    def check_submission(submission_id):
+    def check_current_page(submission_id):
+
         sess = GlobalDB.db().session
 
+        submission_id = Submission.submission_id
```
nmonga91: `Submission` is a reference to the model, not the object we want, so it'll fail when you try to do `Submission.submission_id`. The parameter above needs needs to be changed from `submission_id` to `submission`. Then going forward you should reference `submission` which will refer to the actual object.

kkuratsu:
We are passing `submission_id` not `submission`. I allow both passing submission as `object` or `int`.

PR #706, opened 2017-05-19
Db 1756 local fpds data

alburde1:
This PR is an intermediate PR per the request of @kaitlin so the script can be run on dev.

This script pulls FPDS data from the atom feed and maps it to a detached_award_procurement table where it can be accessed for generating D1 files and running validations later.

The date for import is FY 17.

It isn't optimized so the data load may take a while

Arguments:
- -a forces a clear of the DB and a load from FY 17 to now
- -l gets all data modified since the last_update date stored from previous data loads

Technical Release Notes:
- alembic upgrade to make 2 new tables and upgrade an existing one

- [ ] Signed off on by @klrBAH 
- [x] Reviewed by @nmonga91


dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1124 @@
...
+    return data
+
+
+def extract_text(data_val):
+    if type(data_val) is not str:
+        data_val = data_val['#text']
+    return data_val
+
+
+# Get values from the awardID level of the xml
```
nmonga91: We should convert these kinds of comments that describe what the function is doing into docstings

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1124 @@
...
+        except (KeyError, TypeError):
+            obj['referenced_idv_agency_iden'] = None
+
+        # not in IDV feed, just set it to None
+        obj['transaction_number'] = None
+
+    return obj
+
+
+def process_and_add(data, contract_type, sess, last_run=None):
```
nmonga91: Some additional comments here about what checking `last_run` means would be useful to folks looking at this for the first time

nmonga91:
Changes look good to me, ready to merge once checks pass

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+        except KeyError:
+            listed_data = []
+
+        for ld in listed_data:
+            data.append(ld)
+            i += 1
+
+        # Log which one we're on so we can keep track of how far we are (different numbers for different pulls)
+        if i % log_interval == 0:
+            logger.info("On line " + str(i))
```
klrBAH: This log output is the same as line 1062 (logging lines in get_delete_data). Can we specify which we're in to avoid confusion?

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+        last_run_date = last_run.update_date
+        params = 'LAST_MOD_DATE:[' + last_run_date.strftime('%Y/%m/%d') + ',' + yesterday.strftime('%Y/%m/%d') + '] '
+
+    # TODO remove this later, this is just for testing
+    # params += 'CONTRACTING_AGENCY_ID:1542 '
+    # params = 'VENDOR_ADDRESS_COUNTRY_CODE:"GBR"'
+    # params = 'PIID:"0046"+REF_IDV_PIID:"W56KGZ15A6000"'
+
+    i = 0
+    logger.info('Starting: ' + feed_url + params + 'CONTRACT_TYPE:"' + contract_type.upper() +
```
klrBAH: This log output looks to be about the same as the output on line 1043 (Start message for get_delete_data). Can we specify which we're in to avoid confusion?

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+            if i % 100000 == 0 and i != 0:
+                logger.info("Processing next 100,000 records")
+                process_and_add(data, contract_type, sess, last_run)
+                data = []
+
+        # if we got less than 10 records, we can stop calling the feed
+        if len(listed_data) < 10:
+            break
+
+    logger.info(i)
```
klrBAH: Can we specify that this is the total number of rows from the atom feed so we can parse/interpret if needed?

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+
+        # Every 100 lines, log which one we're on so we can keep track of how far we are
+        if i % 100 == 0:
+            logger.info("On line " + str(i))
+
+        # if we got less than 10 records, we can stop calling the feed
+        if len(listed_data) < 10:
+            break
+
+    logger.info(i)
```
klrBAH: Can we specify that this is the total number of rows from the delete feed so we can parse/interpret if needed?

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+        except KeyError:
+            listed_data = []
+
+        for ld in listed_data:
+            data.append(ld)
+            i += 1
+
+        # Every 100 lines, log which one we're on so we can keep track of how far we are
+        if i % 100 == 0:
+            logger.info("On line " + str(i))
```
klrBAH: See comment above on line 1014 about adding specificity as to which (get vs delete) the line number refers to.

dataactcore/scripts/pullFPDSData.py
```diff
@@ -0,0 +1,1158 @@
...
+
+def get_delete_data(contract_type, now, sess, last_run):
+    """ Get data from the delete feed """
+    data = []
+    yesterday = now - datetime.timedelta(days=1)
+    last_run_date = last_run.update_date
+    params = 'LAST_MOD_DATE:[' + last_run_date.strftime('%Y/%m/%d') + ',' + yesterday.strftime('%Y/%m/%d') + '] '
+
+    i = 0
+    logger.info('Starting: ' + delete_url + params + 'CONTRACT_TYPE:"' + contract_type.upper() + '"')
```
klrBAH: See comment above on line 994 about adding specificity as to which (get vs delete) the start message refers to.

dataactcore/scripts/pullFPDSData.py
alburde1: This one has `feed_url` and the other has `delete_url` which specify `FEEDNAME=PUBLIC` and `FEEDNAME=DELETED` respectively in the URL itself. Do we want more clarity than that or is that enough?

klrBAH: I think for clarity it would be good to have "Starting delete feed:" or "Starting get feed:" instead of just starting. That way, anyone who looks at the logs would understand what was happening without having to know the details about the feed URL.

PR #707, opened 2017-05-23
DB-1876 Add logs for each cross file SQL rule

kkuratsu:
**A/C:**
Add logs for each cross file SQL rule


dataactvalidator/validation_handlers/validator.py
```diff
@@ -162,6 +162,7 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict, first_file, sec
         failed_rows = conn.execute(
             rule.rule_sql.format(submission_id))
         if failed_rows.rowcount:
+            logger.info('VALIDATOR_INFO: Running query: %s on submission id %s', rule.query_name, submission_id)
```
nmonga91: This line will only get logged if there are rows that failed. This needs to be logged before line 162. 

We can also remove the "VALIDATOR_INFO" part of the string. That was there originally to differentiate the logs from the broker but now we can do that in Kibana. I would also consider this format of logging:
`logger.info('Running %s on {submission_id: %s, job_id: %s}', rule.query_name,submission_id, job_id)`

dataactvalidator/validation_handlers/validator.py
```diff
@@ -191,6 +192,9 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict, first_file, sec
                                 str(rule.rule_error_message), values, row['row_number'], str(rule.rule_label),
                                 rule.file_id, rule.target_file_id, rule.rule_severity_id])
 
+        logger.info('VALIDATOR_INFO: Completed SQL cross validation query %s on submission %s',
```
nmonga91: This log's format may need to be updated based on the results of the comment above.

nmonga91:
We also need some more logs, and may need to refactor existing logs, in `validationManager.py`

EricSchles:
[run tests]

kkuratsu:
@nmonga91 Refactored part of the `validationManager.py` to have the same format with the `validator.py`. Currently, the PR is asking for `Add logs for each cross file SQL rule`, so I will add more logs as needed for testing with `DB-1895 Investigate why SSA's submission fails to write warning file for large C file`.

PR #710, opened 2017-05-23
updating local config examples to current

EricSchles:


klrBAH:

PR #711, opened 2017-05-23
updating config example yaml, removing jobqueue db

EricSchles:


klrBAH:

PR #712, opened 2017-05-23
adding is_historical to PublishedAwardFinancialAssistance

EricSchles:


kkuratsu:
Need to add a column in `tests/unit/dataactcore/factories/staging.py`. https://github.com/fedspendingtransparency/data-act-broker-backend/blob/development/tests/unit/dataactcore/factories/staging.py#L273-L324

PR #715, opened 2017-05-26
DB MK 1901 Generate Detached Files

kimminahm:
Allows users to get a list of all CGACs

### DIFF
--- a/dataactbroker/README.md
+++ b/dataactbroker/README.md
@@ -841,6 +841,27 @@ Example output:
 }
 ```
 
+#### GET "/v1/list_sub_tier_agencies/"
+Gets all CGACS that the user has submit/certify permissions
+
+Example input:
+
+None
+
+Example output:
+
+```json
+{
+    "sub_tier_agency_list": [
+      {
+        "agency_name": "Sample Agency",
+        "agency_code": "000",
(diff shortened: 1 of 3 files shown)


dataactbroker/README.md
```diff
@@ -801,6 +801,46 @@ List submissions for all agencies for which the current user is a member of. Opt
 }
 ```
```
alburde1: If you're adding the `list_agencies` route as well as your own, it might make sense to add the `list_sub_tier_agencies` route at the same time.

kimminahm: added

dataactbroker/README.md
```diff
@@ -801,6 +801,67 @@ List submissions for all agencies for which the current user is a member of. Opt
...
+      {
+        "agency_name": "Sample Agency",
+        "cgac_code": "000"
+      }, ...
+    ]
+}
+```
+
+#### GET "/v1/list_sub_tier_agencies/"
+Gets all CGACS that the user has submit/certify permissions
```
alburde1: This isn't exactly accurate. I mean, it does do this, but it also gets all sub tier agencies listed under that CGAC

alburde1:
[run tests]

PR #719, opened 2017-06-01
Fix E/F page status

kkuratsu:


dataactbroker/fileRoutes.py
```diff
@@ -149,6 +149,16 @@ def check_current_page(submission):
             }
             return JsonResponse.create(StatusCode.OK, data)
 
+        # /v1/generateEF/
+        generate_ef = sess.query(Job).filter(Job.submission_id == submission_id, Job.file_type_id == 4,
```
alburde1: `Job.file_type_id == 4` <- that should be job_type_id I think. Check if it works with that change

PR #727, opened 2017-06-12
Db 1955 historical submission files

alburde1:
This PR updates the handling of certifying submissions, placing each certified submission in the certified bucket as before (with the warning files now) but under another folder that corresponds to the certify_history_id for the certification. It also creates a set of entries in the certified_files_history table that stores links to the uploaded and warning files.

It also includes 2 new routes. One to list all the certification history for a specified submission and one to get the signed s3 url for a specific certified file.

Technical Release Notes:
- Alembic upgrade to certify_history table and to create certified_files_history table


kkuratsu:
Looks good.

PR #731, opened 2017-06-16
Db 1867 fabs county state city code

alburde1:
- PrimaryPlaceOfPerformanceCode is a required field for all submissions and cannot be blank.
- 00***** is a valid value and indicates a multi-state project.
- XX***** validates where XX is the 2 character state code.
- XX**### validates for countywide projects, where ### represents the 3-digit county. All transactions reported on an aggregate basis MUST be in the countywide format.
- 00FORGN indicates that the place of performance is in a foreign country. (just allow it to pass, validation for foreign location is in another rule)

Newly added requirements before this PR is merged, make sure checklist is complete before merging:

- [ ] Reviewed by backend
- [x] https://github.com/fedspendingtransparency/data-act-broker-backend/pull/732 has been reviewed and merged
- [ ] initialize.py has been run after the above merge with the `-i` or `-l` flags (if this happens automatically that's fine too, just need to check)
- [x] Validation d40_detached_award_financial_assistance_1.sql updated by Alisa to use the new county table and compare county AND state codes

Technical Release Notes:
- New sql validations


tests/unit/dataactvalidator/test_d39_detached_award_financial_assistance_3.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_success(database):
+    """ PrimaryPlaceofPerformanceZIP+4 should not be provided for any format of PrimaryPlaceOfPerformanceCode
+        other than XX#####. """
+
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="NY12345",
+                                                          place_of_performance_zip4a="1234")
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="ny98765",
+                                                          place_of_performance_zip4a="4312")
+    det_award_3 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code=None,
```
michaeldhess: Shouldn't we want this to fail? The place_of_performance_code isn't in XX##### format but the place_of_performance_zip4a still exists, which goes against the description of the rule. If it's supposed to fail, maybe include NULL in that description of the rule?

tests/unit/dataactvalidator/test_d39_detached_award_financial_assistance_3.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_failure(database):
+    """ Test failure for PrimaryPlaceofPerformanceZIP+4 should not be provided for any format of
+        PrimaryPlaceOfPerformanceCode other than XX#####. """
+
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="00FORGN",
+                                                          place_of_performance_zip4a="1234")
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="00*****",
+                                                          place_of_performance_zip4a="4312")
```
michaeldhess: Add `ny**987` with a zip4a to the failures test

tests/unit/dataactvalidator/test_d39_detached_award_financial_assistance_3.py
alburde1: We probably actually want it to pass but for a different reason. It SHOULD fail if it's None but `place_of_performance_code` is a required field which means we'd have 2 errors showing up for the same problem if it was both missing and throwing an error for `None` as well. What's your opinion on how we should handle this conundrum?

PR #741, opened 2017-06-22
DB mk 2013 Check Current Page

kimminahm:
Fixes to make the check current page work correctly
Please confirm the prerequisites are correct!

### DIFF
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -170,8 +170,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         generate_ef = sess.query(Job).filter(Job.submission_id == submission_id, Job.job_type_id == 4,
                                              Job.number_of_errors == 0)
 
-
-        if review_data.count() == 2 and generate_ef.count() > 0: 
+        if review_data.count() == 2 and generate_ef.count() > 0:
             data = {
                 "message": "The current progress of this submission ID is on /v1/reviewData/ page.",
                 "step": "5"
@@ -187,8 +186,8 @@ def add_file_routes(app, create_credentials, is_local, server_path):
             return JsonResponse.create(StatusCode.OK, data)
 
         validate_cross_file = sess.query(Job).filter(Job.submission_id == submission_id,
-                                                 Job.file_type_id.in_([4, 5]), Job.job_type_id == 2,
-                                                 Job.number_of_errors == 0, Job.file_size.isnot(None))
+                                                     Job.file_type_id.in_([4, 5]), Job.job_type_id == 2,
+                                                     Job.number_of_errors == 0, Job.file_size.isnot(None))
 
         generate_files = sess.query(Job).filter(Job.submission_id == submission_id,
                                                 Job.file_type_id.in_([1, 2, 3]), Job.job_type_id == 2,
@@ -204,8 +203,6 @@ def add_file_routes(app, create_credentials, is_local, server_path):
             return JsonResponse.create(StatusCode.OK, data)
 
         # /v1/generateFiles/
-        
-
         if generate_files.count() == 3:
             data = {
                 "message": "The current progress of this submission ID is on /v1/generateFiles/ page.",
--- a/dataactbroker/fileRoutes.py
+++ b/dataactbroker/fileRoutes.py
@@ -165,7 +165,13 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         # Checks that both E and F files are finished
         review_data = sess.query(Job).filter(Job.submission_id == submission_id,
                                              Job.file_type_id.in_([6, 7]), Job.job_status_id == 4)
-        if review_data.count() == 2:
+
+        # Need to check that cross file is done as well
+        generate_ef = sess.query(Job).filter(Job.submission_id == submission_id, Job.job_type_id == 4,
+                                             Job.number_of_errors == 0)
+
+
+        if review_data.count() == 2 and generate_ef.count() > 0: 
             data = {
                 "message": "The current progress of this submission ID is on /v1/reviewData/ page.",
                 "step": "5"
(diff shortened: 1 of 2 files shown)


dataactbroker/fileRoutes.py
```diff
@@ -162,36 +162,52 @@ def check_current_page(submission):
         submission_id = submission.submission_id
 
         # /v1/reviewData/
+        # Checks that both E and F files are finished
         review_data = sess.query(Job).filter(Job.submission_id == submission_id,
                                              Job.file_type_id.in_([6, 7]), Job.job_status_id == 4)
```
alburde1: So what happens here if they re-upload A/B/C files and some validations fail? Or cross-file validations fail? We don't change the status of E/F generation if something changes I think. We need to also check that cross-file is done and there are no errors in it.

dataactbroker/fileRoutes.py
```diff
@@ -162,36 +162,52 @@ def check_current_page(submission):
...
             }
             return JsonResponse.create(StatusCode.OK, data)
 
         # /v1/generateEF/
+        # Checks that all uploads are successful and both D1 and D2 need to have 0 errors so this must == 2 not > 0
         generate_ef = sess.query(Job).filter(Job.submission_id == submission_id, Job.job_type_id == 4,
                                              Job.job_status_id == 4)
-        if generate_ef.count() > 0:
+        # Need to check that cross file is done as well
+        validate_cross_file = sess.query(Job).filter(Job.submission_id == submission_id,
```
alburde1: We don't need all this. `Job.job_type_id == 4` is the cross-file validation and `Job.job_status_id == 4` is checking if it finished. Cross-file will never be "finished" before D files are uploaded so you don't need that extra check. We DO probably want a number of errors check on it though.

dataactbroker/fileRoutes.py
```diff
@@ -162,36 +162,52 @@ def check_current_page(submission):
...
             data = {
                 "message": "The current progress of this submission ID is on /v1/validateCrossFile/ page.",
                 "step": "3"
             }
             return JsonResponse.create(StatusCode.OK, data)
 
+        # /v1/generateFiles/
+        generate_files = sess.query(Job).filter(Job.submission_id == submission_id,
+                                                Job.file_type_id.in_([1, 2, 3]), Job.job_type_id == 2,
+                                                Job.number_of_errors == 0, Job.file_size.isnot(None))
```
alburde1: I'm wondering if somewhere here we want to actually check for job_status being finished

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -7,8 +7,9 @@
 from dataactbroker import fileRoutes
 from dataactcore.models.lookups import PUBLISH_STATUS_DICT
 from tests.unit.dataactcore.factories.domain import CGACFactory
-from tests.unit.dataactcore.factories.job import SubmissionFactory
+from tests.unit.dataactcore.factories.job import (JobFactory, SubmissionFactory)
```
alburde1: Don't need `()` unless it's on multiple lines

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -60,3 +61,85 @@ def test_list_submissions(file_app, database, user_constants, job_constants):
...
     response = file_app.get("/v1/list_submissions/?certified=mixed")
     assert sub_ids(response) == {submissions[2].submission_id}
+
+
+def test_current_page(file_app, database, user_constants, job_constants, monkeypatch):
+    """Test the route to check what the current progress of the submission is at
+    the correct page
+    """
+
+    cgacs = [CGACFactory() for _ in range(1)]
```
alburde1: If you're only making one you don't really need to do an array. You can just declare a new cgacfactory and work with it without having to index every time.

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -60,3 +61,85 @@ def test_list_submissions(file_app, database, user_constants, job_constants):
...
+    job_c = JobFactory(submission_id=sub.submission_id, file_type=database.session.query(FileType)
+                       .filter_by(name='award_financial').one(), job_type=csv_validation, number_of_errors=0,
+                       file_size=123, job_status=finished_job)
+    job_d1 = JobFactory(submission_id=sub.submission_id, file_type=database.session.query(FileType)
+                        .filter_by(name='award_procurement').one(), job_type=csv_validation, number_of_errors=0,
+                        file_size=123, job_status=finished_job)
+    job_d2 = JobFactory(submission_id=sub.submission_id, file_type=database.session.query(FileType)
+                        .filter_by(name='award').one(), job_type=csv_validation, number_of_errors=0, file_size=123,
+                        job_status=finished_job)
+    job_e = JobFactory(submission_id=sub.submission_id, file_type=database.session.query(FileType)
```
alburde1: E and F don't have csv_validation jobs, only upload jobs

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -60,3 +61,85 @@ def test_list_submissions(file_app, database, user_constants, job_constants):
...
+    response = file_app.get("/v1/check_current_page/?submission_id=" + str(sub.submission_id))
+    response_json = json.loads(response.data.decode('UTF-8'))
+    assert response_json['step'] == '2'
+
+    job_c.number_of_errors = 6
+    database.session.commit()
+    # Fail C file upload
+    response = file_app.get("/v1/check_current_page/?submission_id=" + str(sub.submission_id))
+    response_json = json.loads(response.data.decode('UTF-8'))
+    assert response_json['step'] == '1'
```
alburde1: We need a test somewhere in here (perhaps at the end) where E/F have successfully generated but it still sends us back to an earlier point because something earlier failed (cross-file will not be status 4 if it or anything before it failed).

We also need a test where A/B/C were reuploaded and failed (cross-file will be 1 in this case) and D files were already generated. It needs to go back to the first step in this case

dataactbroker/fileRoutes.py
```diff
@@ -162,36 +162,57 @@ def check_current_page(submission):
...
-        if validate_cross_file.count() > 0:
+                                                 Job.file_type_id.in_([4, 5]), Job.job_type_id == 2,
+                                                 Job.number_of_errors == 0, Job.file_size.isnot(None))
+
+        generate_files = sess.query(Job).filter(Job.submission_id == submission_id,
+                                                Job.file_type_id.in_([1, 2, 3]), Job.job_type_id == 2,
+                                                Job.number_of_errors == 0, Job.file_size.isnot(None),
+                                                Job.job_status_id == 4)
+
+        # /v1/validateCrossFile/)
```
alburde1: Remove the `)`

dataactbroker/fileRoutes.py
```diff
@@ -162,36 +162,57 @@ def check_current_page(submission):
...
+        if validate_cross_file.count() == 2 and generate_files.count() == 3:
             data = {
                 "message": "The current progress of this submission ID is on /v1/validateCrossFile/ page.",
                 "step": "3"
             }
             return JsonResponse.create(StatusCode.OK, data)
 
+        # /v1/generateFiles/
+        
+
```
alburde1: Too many newlines, flake will be mad

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -60,3 +61,98 @@ def test_list_submissions(file_app, database, user_constants, job_constants):
...
     assert sub_ids(response) == {submissions[2].submission_id}
+
+
+def test_current_page(file_app, database, user_constants, job_constants, monkeypatch):
+    """Test the route to check what the current progress of the submission is at
+    the correct page
+    """
+
+    cgacs = CGACFactory()
+    print(cgacs)
```
alburde1: Extra print statement (also, just name it `cgac` instead of `cgacs` because it's just the one. (make sure this works if it isn't an array, if it needs to be one you can just wrap it in `[]` I think)

tests/unit/dataactbroker/test_fileRoutes.py
```diff
@@ -60,3 +61,98 @@ def test_list_submissions(file_app, database, user_constants, job_constants):
...
+    database.session.commit()
+    # E or F failed
+    response = file_app.get("/v1/check_current_page/?submission_id=" + str(sub.submission_id))
+    response_json = json.loads(response.data.decode('UTF-8'))
+    assert response_json['step'] == '4'
+
+    job_e.job_status_id = 4
+    job_cross_file.number_of_errors = 6
+    database.session.commit()
+    # Fail C file upload
```
alburde1: This comment is not correct (apply same feedback to several other tests below)

PR #753, opened 2017-06-27
Db 1683 exec comp loader (excluding Exec Comp Loader)

dpb-bah:
Acceptance Criteria:

DUNS required for all grants and cooperative agreements with ActionDate after October 1, 2010. For these records:

i. Must be a 9-digit DUNS number

ii. When ActionType = A, must be active in the System for Award Management (SAM) on the ActionDate of the award.
//Note: this should be a hard error

iii. When ActionType = B, C, or D, DUNS should exist in SAM but need not be active on the ActionDate (i.e., it may be expired).
//Note: two options: 1) make iii) a warning for everything; or 2) make it an error for financial mods of ActionType B and C. It would be a warning for ActionType D and for non-financial mods. Can probably stick with a warning for now for all these ActionTypes and adjust later.

Technical Specifications:
- Alembic Updates to add two columns to exec_compensation table
- 2 SQL rules


dataactcore/scripts/loadExecComp.py
```diff
@@ -0,0 +1,176 @@
...
+from dataactcore.logging import configure_logging
+from dataactvalidator.health_check import create_app
+from dataactvalidator.scripts.loaderUtils import insert_dataframe
+from dataactcore.config import CONFIG_BROKER
+
+
+logger = logging.getLogger(__name__)
+
+REMOTE_SAM_DIR = '/current/SAM/6_EXECCOMP'
+# sftp -i SAMDATAProd -P 22 samdataprod04@66.77.18.174
```
alburde1: Should this comment still be here? Seems like something we probably don't want just hanging around. If it's ok then I don't mind, just wanted to make sure we weren't putting important information in the code.

dataactcore/scripts/loadExecComp.py
```diff
@@ -0,0 +1,176 @@
...
+            if exec_title.lower() not in unaccepted_titles:
+                exec_comp_data["high_comp_officer{}_full_na".format(index)] = exec_name
+                exec_comp_data["high_comp_officer{}_amount".format(index)] = exec_comp
+
+    return exec_comp_data
+
+
+def get_config():
+    sam_config = CONFIG_BROKER.get('sam')
+    print(sam_config)
```
alburde1: do we want to be printing the sam_config? Is that public information?

dataactvalidator/config/sqlrules/sqlRules.csv
```diff
@@ -161,4 +161,6 @@ D42,"PrimaryPlaceOfPerformanceForeignLocationDescription must be blank for domes
 D43,"When PrimaryPlaceOfPerformanceCountryCode is not USA, PrimaryPlaceOfPerformanceCongressionalDistrict must be blank.","When PrimaryPlaceOfPerformanceCountryCode is not USA, PrimaryPlaceOfPerformanceCongressionalDistrict must be blank.",n,detached_award,fatal,,d43_detached_award_financial_assistance_1
 D43,"If no PrimaryPlaceOfPerformanceZIP+4 is provided, a PrimaryPlaceOfPerformanceCongressionalDistrict must be provided.","If no PrimaryPlaceOfPerformanceZIP+4 is provided, a PrimaryPlaceOfPerformanceCongressionalDistrict must be provided.",n,detached_award,fatal,,d43_detached_award_financial_assistance_2
 D43,"If PrimaryPlaceOfPerformanceCongressionalDistrict is provided it must be valid in the state indicated by PrimaryPlaceOfPerformanceCode.","If PrimaryPlaceOfPerformanceCongressionalDistrict is provided it must be valid in the state indicated by PrimaryPlaceOfPerformanceCode.",n,detached_award,fatal,,d43_detached_award_financial_assistance_3
+D44,"Invalid DUNS number","Invalid DUNS number",n,detached_award,warning,,d44_detached_award_financial_assistance
```
alburde1: If you have more than one part, they should just be `_1`, `_2`, etc. You shouldn't have any un-numbered parts. This applies for your file names and the test files as well.

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_pubished_date_success(database):
+    """ Test valid.
+        Must be a valid 9-digit DUNS number
+        AwardeeOrRecipientUniqueIdentifier is required for AssistanceType of 02, 03, 04, or 05 whose
+        ActionDate after October 1, 2010.
+    """
+
+    exec_comp_1 = ExecutiveCompensation(awardee_or_recipient_uniqu="111111111")
+    det_award_01 = DetachedAwardFinancialAssistanceFactory(assistance_type="02", action_date="10/02/2010",
```
alburde1: Action date has to be YYYYMMDD format (although that shouldn't mess anything up, but we should be consistent). What happens if you enter it in an invalid date format though? Like YYYYDDMM? Does this crash? Pass silently? That should be one of the checks (if action_date isn't a valid date it'd be best if it passed silently since we have other validations picking up that slack)

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance.py
```diff
@@ -0,0 +1,71 @@
...
+                                                           awardee_or_recipient_uniqu="111111111")
+    det_award_02 = DetachedAwardFinancialAssistanceFactory(assistance_type="03", action_date="10/02/2010",
+                                                           awardee_or_recipient_uniqu="111111111")
+    det_award_03 = DetachedAwardFinancialAssistanceFactory(assistance_type="04", action_date="10/02/2010",
+                                                           awardee_or_recipient_uniqu="111111111")
+    det_award_04 = DetachedAwardFinancialAssistanceFactory(assistance_type="05", action_date="10/02/2010",
+                                                           awardee_or_recipient_uniqu="111111111")
+    det_award_05 = DetachedAwardFinancialAssistanceFactory(assistance_type="02", action_date="09/01/2010",
```
alburde1: For these tests from before the date, it would be nice to check that DUNS being blank doesn't return a failure and that it being an invalid DUNS doesn't break it.

dataactvalidator/config/sqlrules/d44_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,46 @@
...
+        WHEN pg_temp.is_date(COALESCE(dafa.action_date, '0'))
+        THEN
+            CAST(dafa.action_date as DATE)
+    END) > CAST('10/01/2010' as DATE)
+    AND dafa.action_type = 'A'
+    AND dafa.row_number NOT IN (
+        SELECT DISTINCT sub_dafa.row_number
+        FROM detached_award_financial_assistance_d44_2_{0} as sub_dafa
+            JOIN executive_compensation AS exec_comp
+            ON (CAST(sub_dafa.awardee_or_recipient_uniqu as float) IS NOT DISTINCT FROM CAST(exec_comp.awardee_or_recipient_uniqu as float)
```
alburde1: `awardee_or_recipient_uniqu` is not a `numeric` in the table (either of them, although I assume we're controlling the `executive_compensation` table) so we can't cast them as float. The user may enter `AAAAAAAAA` which would be invalid but it would still get to the tests because we wouldn't be throwing the type error. This would crash the tests potentially and leave it hanging forever.

dataactvalidator/config/sqlrules/d44_detached_award_financial_assistance_1.sql
```diff
@@ -0,0 +1,46 @@
...
+            AND CAST(sub_dafa.action_date as Date) >= exec_comp.activation_date
```
alburde1: You need another CASE statement here because SQL doesn't do short-circuiting and action_date, again, may be something invalid which causes everything to implode (try putting in `YYYYMMDD` exactly in your tests and see what I mean)

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance.py
```diff
@@ -0,0 +1,71 @@
...
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(assistance_type="02", action_date="10/02/2010",
+                                                          awardee_or_recipient_uniqu=None)
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(assistance_type="03", action_date="10/02/2010",
+                                                          awardee_or_recipient_uniqu='')
+    det_award_3 = DetachedAwardFinancialAssistanceFactory(assistance_type="04", action_date="10/02/2010",
+                                                          awardee_or_recipient_uniqu=None)
+    det_award_4 = DetachedAwardFinancialAssistanceFactory(assistance_type="05", action_date="10/02/2010",
+                                                          awardee_or_recipient_uniqu='')
+    det_award_5 = DetachedAwardFinancialAssistanceFactory(assistance_type="05", action_date="10/02/2010",
+                                                          awardee_or_recipient_uniqu="111111112")
```
alburde1: Add at least one action_date that's just completely broken (all letters or something, just to make sure our casting doesn't break the validation completely)

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,52 @@
...
+        ActionDate of the award.
+        AwardeeOrRecipientUniqueIdentifier is required for AssistanceType of 02, 03, 04, or 05 whose
+        ActionDate after October 1, 2010.
+        When ActionType = B, C, or D, DUNS should exist in SAM but need not be active on the ActionDate
+        (i.e., it may be expired).
+    """
+
+    exec_comp_1 = ExecutiveCompensation(awardee_or_recipient_uniqu="111111111", activation_date="06/23/2017")
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(assistance_type="03", action_date="06/22/2017",
+                                                          awardee_or_recipient_uniqu="111111111", action_type="A")
```
alburde1: Add at least one action_date that's just completely broken (all letters or something, just to make sure our casting doesn't break the validation completely)

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,52 @@
...
+    exec_comp_1 = ExecutiveCompensation(awardee_or_recipient_uniqu="111111111", activation_date="06/21/2017")
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(awardee_or_recipient_uniqu="111111111", action_type="A",
+                                                          assistance_type="02", action_date="06/23/2017")
+    # inactive at action date but not A files
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(awardee_or_recipient_uniqu="111111111", action_type="B",
+                                                          assistance_type="03", action_date="06/21/2017")
+    det_award_3 = DetachedAwardFinancialAssistanceFactory(awardee_or_recipient_uniqu="111111111", action_type="C",
+                                                          assistance_type="04", action_date="06/21/2017")
+    det_award_4 = DetachedAwardFinancialAssistanceFactory(awardee_or_recipient_uniqu="111111111", action_type="D",
+                                                          assistance_type="05", action_date="06/21/2017")
```
alburde1: Put at least one with an invalid `awardee_or_recipient_uniqu` that is outside the range.

tests/unit/dataactvalidator/test_d44_detached_award_financial_assistance_1.py
```diff
@@ -0,0 +1,52 @@
...
+    """ Test valid. When ActionType = A, must be active in the System for Award Management (SAM) on the
+        ActionDate of the award.
+        AwardeeOrRecipientUniqueIdentifier is required for AssistanceType of 02, 03, 04, or 05 whose
+        ActionDate after October 1, 2010.
+        When ActionType = B, C, or D, DUNS should exist in SAM but need not be active on the ActionDate
+        (i.e., it may be expired).
+    """
+    exec_comp_1 = ExecutiveCompensation(awardee_or_recipient_uniqu="111111111", activation_date="06/21/2017")
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(awardee_or_recipient_uniqu="111111111", action_type="A",
+                                                          assistance_type="02", action_date="06/23/2017")
```
alburde1: Put at least one with an invalid awardee_or_recipient_uniqu that is outside the range.

dataactcore/scripts/loadExecComp.py
dpb-bah: It's not necessary as this info is stored in the config. Will remove.

dataactcore/scripts/loadExecComp.py
dpb-bah: That's also left from the original script. Removing.

dataactvalidator/config/sqlrules/sqlRules.csv
dpb-bah: Understood. Updating.

PR #763, opened 2017-07-05
Db 1927 fabs dashboard

dpb-bah:
Updating backend for DB-1927

- [x] Merged concurrently with [FrontEnd PR](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/630) 
- [x] Reviewed by backend

Acceptance Criteria:
1. Remove the date range request when a FABS submission is started (see image in JIRA)
2. Change "File D2: Financial Assistance" to "Financial Assistance Broker Submission (FABS) File" (see image in JIRA).
3. On the Submission Dashboard for FABS, replace the 'Reporting Period" section with "Action Date Range'. (see image in JIRA)
FABS would have “Earliest” and “Latest” here instead of "Start" and "End". This would be filled out based on the earliest and latest Action Dates in the specific submitted file.
4. This should also be reflected in the "Recent Activity" section on the front page for FABS, when that is designed.

Technical Specifications:
- Removing `NOT NULL` setting for `reporting_start_date` and `reporting_end_date` columns in the submissions table, requiring alembic updates


kimminahm:
Everything looks good to me but will have Alisa double check for me.

dataactbroker/fileRoutes.py
```diff
@@ -172,6 +174,17 @@ def check_current_page(submission):
 
         submission_id = submission.submission_id
 
+        # /v1/uploadDetachedFiles/
+        # DetachedFiles
+        detached_data = sess.query(Submission).filter(Submission.submission_id == submission_id,
```
alburde1: I'm not sure what the `Submission.d2_submission` does (is it working? Is it filtering as we expect? I don't know) but either way, this query isn't necessary. Because we have the entire submission in the `submission` variable we can' just check for `if submission.d2_submission` instead of `if detached_data.count() > 0:`, simplifying the check and removing an unnecessary DB call.

dataactbroker/handlers/fileHandler.py
```diff
@@ -711,42 +704,18 @@ def upload_detached_file(self, create_credentials):
                 existing_submission = None
                 existing_submission_obj = None
 
-            for request_field, submission_field in request_job_mapping.items():
```
alburde1: Why are we removing all this? It might be fine, just walk me through why that decision was made.

dataactcore/interfaces/function_bag.py
```diff
@@ -647,3 +647,21 @@ def get_last_validated_date(submission_id):
...
+    min_action_date = None
+    max_action_date = None
+    date_from_format = '%Y%m%d'
+    date_to_format = '%Y-%m-%d'
+    action_dates = sess.query(func.min(DetachedAwardFinancialAssistance.action_date).label("min_action_date"),
+                              func.max(DetachedAwardFinancialAssistance.action_date).label("max_action_date"))\
+        .filter(DetachedAwardFinancialAssistance.submission_id == submission_id)
+    res = action_dates.one()
+    if res.min_action_date:
+        min_action_date = datetime.strptime(res.min_action_date, date_from_format).date().strftime(date_to_format)
```
alburde1: This may be a dumb question, why are we doing `.date()` in the middle of this? Doesn't `strptime()` create the datetime object we can use `strftime()` on? I could be wrong, I don't use these too often.

Follow-up dumb question. If we only use this to set the reporting start and end dates later in the process, why do we have to format them at all? Does the given variant not work when inserting into a date?

dataactcore/interfaces/function_bag.py
```diff
@@ -647,3 +647,21 @@ def get_last_validated_date(submission_id):
...
+def get_action_dates(submission_id):
+    sess = GlobalDB.db().session
+
+    min_action_date = None
+    max_action_date = None
+    date_from_format = '%Y%m%d'
+    date_to_format = '%Y-%m-%d'
+    action_dates = sess.query(func.min(DetachedAwardFinancialAssistance.action_date).label("min_action_date"),
+                              func.max(DetachedAwardFinancialAssistance.action_date).label("max_action_date"))\
+        .filter(DetachedAwardFinancialAssistance.submission_id == submission_id)
```
alburde1: Do we want valid and invalid (has errors) rows to count towards the action dates or only valid rows? Might need to check with someone on this one.

dataactbroker/handlers/fileHandler.py
dpb-bah: Since for fabs submissions we're not getting reporting dates from the front end, this section of the code was removed which would handle the incoming reporting dates from the front end.

dpb-bah: The request_job_mapping dict was soley used for this purpose.

dataactcore/interfaces/function_bag.py
```diff
@@ -647,3 +647,24 @@ def get_last_validated_date(submission_id):
...
+        the earliest action date (str)
+        the latest action date (str)
+    """
+
+    sess = GlobalDB.db().session
+    action_dates = sess.query(func.min(DetachedAwardFinancialAssistance.action_date).label("min_action_date"),
+                              func.max(DetachedAwardFinancialAssistance.action_date).label("max_action_date"))\
+        .filter(DetachedAwardFinancialAssistance.submission_id == submission_id,
+                DetachedAwardFinancialAssistance.is_valid == valid)
+    res = action_dates.one()
```
alburde1: Could this just be done as part of the query? Just add `.one()` to the end of it and do `action_dates.` to get the max and min dates? Just seems like an extra step/variable we don't need. Also just checking, what happens if it doesn't have one result at least (all rows invalid). Maybe use `one_or_none()`?

dataactcore/interfaces/function_bag.py
```diff
@@ -647,3 +647,24 @@ def get_last_validated_date(submission_id):
...
+    Returns:
+        the earliest action date (str)
+        the latest action date (str)
+    """
+
+    sess = GlobalDB.db().session
+    action_dates = sess.query(func.min(DetachedAwardFinancialAssistance.action_date).label("min_action_date"),
+                              func.max(DetachedAwardFinancialAssistance.action_date).label("max_action_date"))\
+        .filter(DetachedAwardFinancialAssistance.submission_id == submission_id,
+                DetachedAwardFinancialAssistance.is_valid == valid)
```
alburde1: I don't think comparing true/false this way is accurate (or correct at least). You should be using `DetachedAwardFinancialAssistance.is_valid.is_(True)`. I also specify `True` because at no point will we be looking for `False` to my knowledge so this probably doesn't need to be a variable either.

dataactcore/interfaces/function_bag.py
```diff
@@ -647,3 +647,24 @@ def get_last_validated_date(submission_id):
...
+    res = action_dates.one()
+    return res.min_action_date, res.max_action_date
```
alburde1: What happens if there are no min/max action dates? Should there be some sort of check for that (return res.min_action_date if res else None or whatever the actual code would be?)

dataactcore/interfaces/function_bag.py
dpb-bah: I can't imagine when it'd return multiple results as it's just finding the min/max.

dataactcore/interfaces/function_bag.py
dpb-bah: When there are no min/max action dates. The columns still exist, but they're values are none.

PR #764, opened 2017-07-07
Db 2049 awarding office code length

alburde1:
Adding AwardingOfficeCode length check to rule D38 next to FundingOfficeCode length check

Technical Release Notes:
- Updated/new SQL rule


dpb-bah:
Good idea of connecting it to D38, merging

PR #765, opened 2017-07-07
Db 2051 ppop zip4 format check

alburde1:
Explicit PPoP Zip+4 format check (#####, #########, or #####-####). Improperly formatted zips now silently pass the state check so we can surface the error explicitly in a different rule.

Also updated one of the regex checks in the existing rule to use an optional modifier instead of doing 2 checks

Technical Release Notes:
- New/updated sql rule


dpb-bah:
Looks good, the regex check should help with optimization and the extra rule will be more explicit

PR #766, opened 2017-07-07
Db 2065 fabs max length error

alburde1:
Entering something that is too long for the field is now an error in FABS but still warning in normal submissions


dpb-bah:
Simple change, merging

PR #767, opened 2017-07-07
Db 2066 correction indicator allowed values

alburde1:
Adding rule d45: When provided, CorrectionLateDeleteIndicator has to be C, D, or L

Technical Release Notes:
- New sql rule


dpb-bah:
Looks simple enough, comments/messages are consistent, merging

PR #768, opened 2017-07-07
Db 2047 cfda number format

alburde1:
Stopping the validations from imploding when cfda_number is letters. The format check was correct before it just never got through validations if it wasn't all numbers.
Comparing using the existing string (it should be more accurate this way anyway)

Technical Release Notes:
- Updated sql rule


dpb-bah:
Makes sense, merging

PR #769, opened 2017-07-07
Db 2048 fail nonexistent cfda number

alburde1:
If a cfda number doesn't exist in our table, fail the row

Technial Release Notes:
- Updated sql rule


dpb-bah:
Looks good, merging

PR #770, opened 2017-07-07
Db 2068 updating wrong type error text

alburde1:
Changing text in "type_error" so it's more clear that other validations didn't run.

Technical Release Notes:
- Updating error text, will have to reinitialize your DB


dpb-bah:
Went through the code to see if there were any more instances of this and it seems all bases were covered, merging after build completes.

PR #774, opened 2017-07-10
Db mk 1944 gtas window restrictions

kimminahm:
Technical Notes:
upgrade alembic


dataactbroker/fileRoutes.py
```diff
@@ -75,6 +75,29 @@ def finalize_submission(upload_id):
...
+        window_open = False
+        data = {}
+
+        if gtas_window is not None:
+            message = 'GTAS Window is open'
+            data = gtas_window
+            window_open = True
+            data = {'start_date': str(gtas_window.start_date), 'end_date': str(gtas_window.end_date)}
+
+        return JsonResponse.create(StatusCode.OK, {"message": message, "open": window_open, "data": data})
```
alburde1: Do you need to return a message at all? You only really need to determine if it's open or not. The banner shouldn't display if the window isn't open so you can just manually set the message. Additionally, you don't need to return "open" because you can either return `None` or data for "data" and just compare that instead. Data will always be `None` when "open" is `False`.

dataactbroker/fileRoutes.py
```diff
@@ -376,6 +399,11 @@ def certify_submission(submission):
         if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
             return JsonResponse.error(ValueError("Submission has already been certified"), StatusCode.CLIENT_ERROR)
 
+        window = gtas_window()
```
alburde1: If you're going to be calling this function outside the route, you should make it a helper function that is called from inside the route and here. You shouldn't be calling route functions from other functions generally. Maybe name it `get_gtas_window` and call it to get the gtas_window for each of these. It can just return `one_or_none` from the query.

dataactcore/migrations/versions/bd821df63bb1_merging_alisas_and_my_branches.py
```diff
@@ -0,0 +1,36 @@
+"""merging alisas and my branches
```
alburde1: This message probably should say something like "merge 551a8 df2f5" because no one will care whose branches it was merging. That should also be the title of this file after the hash at the beginning.

dataactbroker/fileRoutes.py
```diff
@@ -376,6 +387,11 @@ def certify_submission(submission):
         if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
             return JsonResponse.error(ValueError("Submission has already been certified"), StatusCode.CLIENT_ERROR)
 
+        window = get_gtas_window()
+        if window.open:
```
alburde1: This won't work. There's no `.open` to get.

dataactbroker/fileRoutes.py
```diff
@@ -75,6 +75,17 @@ def finalize_submission(upload_id):
     def check_status(submission):
         return get_status(submission)
 
+    @app.route("/v1/gtas_window/", methods=["GET"])
+    def gtas_window():
+        gtas_window = get_gtas_window()
+
+        data = None
+
+        if gtas_window is not None:
```
alburde1: Would it work if you just did `if gtas_window`? This statement is equivalent to `if (variable != null)` which, as you know, is kind of redundant.

dataactbroker/fileRoutes.py
```diff
@@ -376,6 +387,11 @@ def certify_submission(submission):
         if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
             return JsonResponse.error(ValueError("Submission has already been certified"), StatusCode.CLIENT_ERROR)
 
+        window = get_gtas_window()
+        if window is None:
```
alburde1: Same as above, can this just be `if not window:`? If it can't that's fine, but it seems cleaner

dataactbroker/README.md
```diff
@@ -721,6 +721,34 @@ This route certifies the specified submission, if possible. If a submission has
 ```
 * `message` - A message indicating whether or not the action was successful. Any message other than "Success" indicates a failure.
 
+#### GET "/v1/gtas_window"
+
+This route checks if there is a gtas window currently open, and if it is returns the start and end date, else returns None
+
+##### Body (JSON)
```
alburde1: If there's no body, just put `None` because you aren't actually passing a body back with this. Similarly, remove the Body Description.

dataactbroker/README.md
```diff
@@ -721,6 +721,34 @@ This route certifies the specified submission, if possible. If a submission has
...
+{}
+```
+
+##### Body Description
+
+
+##### Response (JSON)
+
+```
+{}
```
alburde1: This is not the response. The response you show here is just an example so you don't have to show an "or". Also, the response is
```
{
  data: {
       start_date: '2012-05-17',
       end_date: '2012-06-17'
   }
}
```

dataactbroker/README.md
```diff
@@ -721,6 +721,34 @@ This route certifies the specified submission, if possible. If a submission has
...
+##### Body (JSON)
+
+```
+{}
+```
+
+##### Body Description
+
+
+##### Response (JSON)
```
alburde1: You should write what the response is. If it IS a gtas window, it returns a JSON containing a data object with the start and end dates. If it isn't, it returns a data element containing `null`

dataactbroker/fileRoutes.py
```diff
@@ -376,6 +387,11 @@ def certify_submission(submission):
         if submission.publish_status_id == PUBLISH_STATUS_DICT['published']:
             return JsonResponse.error(ValueError("Submission has already been certified"), StatusCode.CLIENT_ERROR)
 
+        window = get_gtas_window()
+        if not window:
```
alburde1: wait, this logic doesn't make sense. If it isn't the window it can't be certified? It should be `if window:`

dataactbroker/README.md
```diff
@@ -721,6 +721,34 @@ This route certifies the specified submission, if possible. If a submission has
 ```
 * `message` - A message indicating whether or not the action was successful. Any message other than "Success" indicates a failure.
 
+#### GET "/v1/gtas_window"
+
+This route checks if there is a gtas window currently open, and if it is returns the start and end date, else returns None
+
+##### Body 
+
+```
```
alburde1: This doesn't need to be in quotes, it's just "None". Remove the `Body Description` section still

PR #776, opened 2017-07-10
Db 2067 awarding agency code derived

alburde1:
Making AwardingAgencyCode a derived field instead of a required one. It has been removed from DetachedAwardFinancialAssistance since it will never be filled in there now.

Technical Release Notes:
- Alembic migration
- Removing a sql rule


dpb-bah:
Removed the right sections and derivation looks good

PR #777, opened 2017-07-10
Db 2071 fixing to char formatting

alburde1:
to_char is dumb
Also updated tests to use the actual type that program_number is for a more correct test

Technical Release Notes:
- Fixing a sql validation


dpb-bah:
to_char is dumb, merging

PR #784, opened 2017-07-14
Db 2085 renaming rule d45

alburde1:
Renaming rule D45 to D22 to remove gap in rule labels

Technical Release Notes:
- Updating sql rule


dpb-bah:
Consistent, merging

PR #787, opened 2017-07-14
Db 2054 d43 only applies domestically

alburde1:
+ Fixing what columns are printed for one part of the rule
+ Making zip+4 or CD required only when country code is USA


dpb-bah:
Looks good, can merge when allowed

PR #788, opened 2017-07-14
Adding SQL rule to check formatting for DUNS number

dpb-bah:
- [ ]  Needs to be retested after Duns is loaded via the SAM script.

Technical Specifications:
- Added SQL rule


dpb-bah:
No matter how I revert and merge locally, it thinks my updated d45 should become d22. Closing this and making a new PR.

PR #789, opened 2017-07-14
Splitting part of D41 in two with different error messages

dpb-bah:
Splitting the original part of D41 ("When provided, PrimaryPlaceofPerformanceZIP+4 must be in the state specified by PrimaryPlaceOfPerformanceCode.") into two separate error messages. One being where the first five digits are wrong and the latter being when only the last four digits are wrong.

Technical Specifications:
Adding part of a SQL rule


dataactvalidator/config/sqlrules/d41_detached_award_financial_assistance_3.sql
```diff
@@ -11,26 +11,13 @@ SELECT
...
-                    -- if any other format, pass (will check format in another rule)
-                    ELSE FALSE
-                    END)
-             END)
-        -- if not zip4, just let it pass
-        ELSE FALSE
-        END
\ No newline at end of file
+WHERE CASE WHEN (COALESCE(dafa.place_of_performance_zip4a, '') != ''
+                 AND (dafa.place_of_performance_zip4a ~ '^\d\d\d\d\d$'
```
alburde1: We still need 2 WHEN clauses here because the comparison for 5 digits vs 9 digits is different (what if someone enters `12345` and we compare 5-digit `12345` and 4-digit `2345` and there's no 4-digit `2345` but the 5-digit (which is all they entered) is still valid? This isn't being checked for in this rule right now.) If we're ONLY checking 5-digit here (which it seems like the intent is), then the wording is incorrect in the rule description.

dataactvalidator/config/sqlrules/d41_detached_award_financial_assistance_5.sql
```diff
@@ -0,0 +1,25 @@
...
+                            FROM detached_award_financial_assistance_d41_5_{0} AS sub_dafa
+                            JOIN zips
+                            ON UPPER(LEFT(sub_dafa.place_of_performance_code, 2)) = zips.state_abbreviation
+                            AND UPPER(LEFT(sub_dafa.place_of_performance_zip4a, 5)) = zips.zip5)
+    AND dafa.row_number NOT IN (SELECT DISTINCT sub_dafa.row_number
+                                FROM detached_award_financial_assistance_d41_5_{0} AS sub_dafa
+                                JOIN zips
+                                ON UPPER(LEFT(sub_dafa.place_of_performance_code, 2)) = zips.state_abbreviation
+                                   AND LEFT(sub_dafa.place_of_performance_zip4a, 5) = zips.zip5
+                                   AND SUBSTRING(REPLACE(sub_dafa.place_of_performance_zip4a, '-', ''), 6, 4) = zips.zip_last4)
```
alburde1: Thinking about it, since we've already made that limit of "it HAS to be in this format" we can probably just use RIGHT instead of SUBSTRING(REPLACE()) (we should double-check but I'm pretty sure we can)

dataactvalidator/config/sqlrules/d41_detached_award_financial_assistance_3.sql
dpb-bah: After reviewing the ticket with the ticket reporter, the language of the first ticket should instead say:
>  "The provided PrimaryPlaceofPerformanceZIP+4 must be in the state specified by the PrimaryPlaceOfPerformanceCode. In this specific submission row, the ZIP5 (and by extension the full ZIP+4) is not a valid ZIP code in the state in question."

tests/unit/dataactvalidator/test_d41_detached_award_financial_assistance_5.py
```diff
@@ -0,0 +1,78 @@
...
+    actual = set(query_columns(_FILE, database))
+    assert expected_subset == actual
+
+
+def test_success(database):
+    """ The provided PrimaryPlaceofPerformanceZIP+4 must be in the state specified by PrimaryPlaceOfPerformanceCode.
+        In this specific submission row, the first five digits are valid and located in the correct state, but the
+        last 4 are invalid."""
+
+    # XX00000 validates here because it passes as long as the zip is valid in that state, this is checked
```
alburde1: This comment is probably not needed. We aren't checking anything with this format.

tests/unit/dataactvalidator/test_d41_detached_award_financial_assistance_3.py
```diff
@@ -13,7 +13,8 @@ def test_column_headers(database):
 
 def test_success(database):
     """ When provided, PrimaryPlaceofPerformanceZIP+4 must be in the state specified by PrimaryPlaceOfPerformanceCode.
-        Ignore cases where zip4 isn't provided """
+        In this specific submission row, the ZIP5 (and by extension the full ZIP+4) is not a valid ZIP code in the
+        state in question."""
 
     # XX00000 validates here because it passes as long as the zip is valid in that state, this is checked
```
alburde1: This comment is probably not needed. We aren't checking anything with this format. (I know it was already here, but we might as well remove it)

tests/unit/dataactvalidator/test_d41_detached_award_financial_assistance_5.py
```diff
@@ -0,0 +1,78 @@
...
+    zips = Zips(zip5="12345", zip_last4="6789", state_abbreviation="NY")
+    # ignored because no zip4
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="NY*****",
+                                                          place_of_performance_zip4a="")
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="Ny**123",
+                                                          place_of_performance_zip4a=None)
+    # valid 5 digit zip
+    det_award_3 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="Ny**123",
+                                                          place_of_performance_zip4a="12345")
+    det_award_4 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="NY98765",
```
alburde1: Do we need 2 tests for a valid 5-digit zip if it's being ignored in this rule? You already have a 5-digit zip section and if invalid ones are being ignored we can pretty much assume valid ones are passing (although you could move one of these down there just in case)

tests/unit/dataactvalidator/test_d41_detached_award_financial_assistance_5.py
```diff
@@ -0,0 +1,78 @@
...
+    det_award_4 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="ny10986",
+                                                          place_of_performance_zip4a="123-456789")
+    errors = number_of_errors(_FILE, database, models=[det_award_1, det_award_2, det_award_3, det_award_4, zips])
+    assert errors == 0
+
+    # invalid 5 digit zip - this should pass but is handled in d41_3
+    det_award_1 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="ny10986",
+                                                          place_of_performance_zip4a="12346")
+    det_award_2 = DetachedAwardFinancialAssistanceFactory(place_of_performance_code="NA*****",
+                                                          place_of_performance_zip4a='12345')
```
alburde1: Adding to these tests, there should be a 9-digit here with an invalid first 5 to see if it's properly ignored.

PR #790, opened 2017-07-17
Db 1943 report by frec

dpb-bah:
- [ ] [Front-End PR](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/650) is merged concurrently 
- [x] Reviewed by Backend

Acceptance Criteria:
- Agencies on the approved list can submit by their FR entity code instead of by their AID code
- MAX groups are created and updated to work the same way for the FR entity reporters
- Validations that use the AID code of the submitting agency are updated

Technical Specifications:
- Alembic Migrations


dataactbroker/handlers/accountHandler.py
```diff
@@ -239,24 +243,42 @@ def email_users(self, system_email):
...
             logger.warning('Malformed permission: %s', perm)
             continue
 
-        cgac_code, perm_level = components
+        codes, perm_level = components
         perm_level = perm_level.lower()
-        if cgac_code not in available_codes or perm_level not in 'rws':
+        split_codes = codes.split('-FREC_')
+        frec_code = cgac_code = None
+        if len(split_codes) == 2:
```
alburde1: It would be nice to have some comments here so we know what's going on and why we're checking these lengths.

dataactbroker/handlers/fileHandler.py
```diff
@@ -1443,13 +1464,18 @@ def list_submissions(page, limit, certified, sort='modified', order='desc', d2_s
     if not options.get(sort):
         sort = 'modified'
 
-    model = options[sort]['model']
-    col = options[sort]['col']
+    order_by = getattr(options[sort]['model'], options[sort]['col'])
```
alburde1: Can we change this variable name? It seems like bad practice to have a variable named the same thing as a function we're using on it (farther down)

dataactbroker/handlers/fileHandler.py
```diff
@@ -1704,6 +1730,13 @@ def fabs_derivations(obj):
     else:
         logger.error("CFDA title not found for CFDA number %s", obj['cfda_number'])
 
+    # deriving awarding agency name
+    if 'awarding_agency_code' in obj.values() and obj['awarding_agency_code']:
```
alburde1: This is always going to be false, we can just remove the entire thing. Also, we just noticed that there's a print a few lines down from this. Can you just remove it while you're in there?

dataactcore/migrations/versions/4d8408c33fee_add_frec_to_user_model.py
```diff
@@ -0,0 +1,60 @@
...
+    op.create_primary_key('user_affiliation_pkey', 'user_affiliation', ['user_affiliation_id'])
+    op.alter_column('user_affiliation', 'cgac_id',
+               existing_type=sa.INTEGER(),
+               nullable=True)
+    ### end Alembic commands ###
+
+
+def downgrade_data_broker():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('user_affiliation', 'cgac_id',
```
alburde1: This is going to mess up during downgrade. We need to default it to something

dataactvalidator/scripts/loadFile.py
```diff
@@ -7,7 +7,8 @@
 from dataactcore.config import CONFIG_BROKER
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.logging import configure_logging
-from dataactcore.models.domainModels import CGAC, SubTierAgency, ObjectClass, ProgramActivity, CountryCode, CFDAProgram
+from dataactcore.models.domainModels import CGAC, SubTierAgency, ObjectClass, ProgramActivity, CountryCode, \
```
alburde1: We normally surround the imports with `()` when they go on multiple lines. We should stay consistent.

tests/unit/dataactbroker/test_domainRoutes.py
```diff
@@ -43,20 +45,23 @@ def test_list_agencies_superuser(database, monkeypatch, domain_app):
...
     sess.commit()
     monkeypatch.setattr(domainRoutes, 'g', Mock(user=user))
 
     result = domain_app.get('/v1/list_all_agencies/').data.decode('UTF-8')
-    result = json.loads(result)
-    result = {el['cgac_code'] for el in result['cgac_agency_list']}
-    assert result == {'0', '1', '2'}    # i.e. all of them
+    response = json.loads(result)
+    result = {el['cgac_code'] for el in response['agency_list']}
+    assert result == {'0', '1', '2', '1'}  # i.e. all of them
```
alburde1: Why is there an extra '1' at the end of this (same question for below)?

dataactbroker/fileRoutes.py
```diff
@@ -275,7 +275,11 @@ def generate_file(submission_id, file_type):
     def generate_detached_file(file_type, cgac_code, frec_code, start, end):
         """ Generate a file from external API, independent from a submission """
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
-        return file_manager.generate_detached_file(file_type, cgac_code, frec_code, start, end)
+        if cgac_code or frec_code:
```
alburde1: If you're returning an error you can check `if not cgac_code and not frec_code` and then return. Then you don't need an else. This isn't necessarily incorrect but we want to stay consistent.

dataactbroker/fileRoutes.py
```diff
@@ -384,11 +388,13 @@ def delete_submission(submission):
                  'reporting_fiscal_year': webargs_fields.String(required=True),
                  'reporting_fiscal_period': webargs_fields.String(required=True)})
     def check_year_and_quarter(cgac_code, frec_code, reporting_fiscal_year, reporting_fiscal_period):
-        """ Check if cgac code, year, and quarter already has a published submission """
-        sess = GlobalDB.db().session
-
-        return find_existing_submissions_in_period(sess, cgac_code, frec_code, reporting_fiscal_year,
-                                                   reporting_fiscal_period)
+        """ Check if cgac (or frec) code, year, and quarter already has a published submission """
+        if cgac_code or frec_code:
```
alburde1: If you're returning an error you can check if not cgac_code and not frec_code and then return. Then you don't need an else. This isn't necessarily incorrect but we want to stay consistent.

dataactbroker/handlers/accountHandler.py
```diff
@@ -258,20 +258,20 @@ def perms_to_affiliations(perms):
...
         split_codes = codes.split('-FREC_')
         frec_code = cgac_code = None
-        if len(split_codes) == 2:
+        if len(split_codes) == 2:  # permissions for FR entity code
             frec_code = split_codes[1]
             if frec_code not in available_frecs:
                 logger.warning('Malformed permission: %s', perm)
                 continue
-        else:
+        else:  # permissions for CGAC
```
alburde1: We try not to do comments in the same line as the code where possible. Move these above the `if` and `else`

PR #794, opened 2017-07-17
Revert "Db 2054 d43 only applies domestically"

klrBAH:
Reverts fedspendingtransparency/data-act-broker-backend#787


dpb-bah:
Reverting from this morning

PR #785, opened 2017-07-14
Db 2078 1933 remove unnecessary rules

alburde1:
Remove rules A25-A28 and A31 because they are unnecessary/redundant.

Technical Release Notes:
- Removed a few sql rules

PR #795, opened 2017-07-17
Revert "Db 2078 1933 remove unnecessary rules"

klrBAH:
Reverts fedspendingtransparency/data-act-broker-backend#785


dpb-bah:
Reverting from this morning

PR #796, opened 2017-07-17
Revert "Db 2085 renaming rule d45"

klrBAH:
Reverts fedspendingtransparency/data-act-broker-backend#784


dpb-bah:
Reverting from this morning

PR #797, opened 2017-07-17
Backend: Dev stable -> Staging

alburde1:


dpb-bah:
Same commits, merging

PR #803, opened 2017-07-18
Db 2053 legal entity congressional district d35

alburde1:
Fixes to rule D35. Removing rule parts about LegalEntityCongressionalDistrict, adding exceptions to part of a rule, removing legal_entity_congressional column from DetachedAwardFinancialAssistance table completely (we weren't filling it in anyway, it was just always an empty column, oops)

Technical Release Notes:
- Alembic migration
- SQL rule updates


dataactbroker/handlers/fileHandler.py
```diff
@@ -1743,5 +1743,18 @@ def fabs_derivations(obj):
...
+        # if we have a legal entity zip+4 provided
+        if obj['legal_entity_zip_last4']:
+            zip_data = sess.query(Zips).\
+                filter_by(zip5=obj['legal_entity_zip5'], zip_last4=obj['legal_entity_zip_last4']).first()
+        else:
+            zip_data = sess.query(Zips).filter_by(zip5=obj['legal_entity_zip5']).first()
+        # set zip_data to the congressional district if we got a result
+        if zip_data:
+            zip_data = zip_data.congressional_district_no
+        obj['legal_entity_congressional'] = zip_data
```
dpb-bah: Stupid question: if no result is found, is it okay if this gets set to `None`? Should there be a default value?

alburde1: We don't have a default value for the zip. I'm not sure what we'd default to if we even tried. It's also required for domestic recipients so it wouldn't get here if it was domestic and didn't have a zip. If it isn't domestic, we don't need a CD because there isn't one.

dpb-bah: Sounds good, merging

PR #827, opened 2017-07-31
Db 1875 legal entity derivations

alburde1:
Creating derivations for legal entity content
+ Different derivations if it’s record_type 1 or 2
+ base derivations on zip5 (and zip4 where available) for record type 2
+ base derivations on ppop code for record type 1

Technical Release Notes:
* New data in PublishedAwardFinancialAssistance table, migration


dataactbroker/handlers/fileHandler.py
```diff
@@ -1808,18 +1808,54 @@ def fabs_derivations(obj):
...
+        obj['legal_entity_county_name'] = county_info.county_name
+
+        # legal entity state data
+        state_info = sess.query(States).filter_by(state_code=zip_data.state_abbreviation).one()
+        obj['legal_entity_state_code'] = state_info.state_code
+        obj['legal_entity_state_name'] = state_info.state_name
+
+    # deriving legal entity stuff that's based on record type of 1 (ppop code must be in the format XX**### for these)
+    if obj['record_type'] == 1:
+        obj['legal_entity_city_name'] = None
```
dpb-bah: The ticket says legal_entity_city_name should be derived from LegalEntityZip5 when present and doesn't mention whether this is the case for RecordType=1 or 2. Is LegalEntityZip5 not available for when RecordType=1 and if it is available, should we set it just the same as if it was RecordType=1? The lack of mentioning the RecordType for this field in the ticket raises these questions.

PR #830, opened 2017-08-08
DB-2116 update FABS! historical loader

michaeldhess:
AC: 
- Update FABS! model to allow us to keep historical changes to FABS! records
- Update historical loader to keep all data, including inactive records
- Update FABS! derivations, submission process, etc to match the new model

Technical Release Notes:
- Alembic migration: updated PublishedAwardFinancialAssistance table


dpb-bah:
Tested submission process, upload detached files, and ran the script all locally.

PR #831, opened 2017-08-08
Adding the DUNS table and data loader.

klrBAH:
Adding the DUNS table and data loader.

Technical Release Notes:
- Added DUNS table, migration (plus migrations for dev merge)
- Added LoadDUNS.py to load DUNS files from sftp site or locally. Requires "sam_duns" section in Broker section of config.


dpb-bah:
The codeclimate issues are valid and noted, though the cyclomatic complexity issues are simply due to the benchmark checks.

dataactcore/migrations/versions/c048f1dcdfa2_.py
```diff
@@ -0,0 +1,36 @@
+"""empty message
```
alburde1: Need a meaningful message. We usually put `Merge 9649eb55fa0b and 967fbc28a6ef ` for example. This file should also be renamed with `merge_9649e_and_967fb` tacked onto the end of the existing name.

dataactcore/migrations/versions/9649eb55fa0b_add_duns_index.py
```diff
@@ -0,0 +1,36 @@
+"""add_duns_index
```
alburde1: Why is this underscored? `Add DUNS index` seems more human-readable

dataactcore/migrations/versions/ff4728a82180_add_duns_table.py
```diff
@@ -0,0 +1,51 @@
+"""add_duns_table
```
alburde1: Why is this underscored? `Add DUNS table` seems more human-readable

dataactcore/scripts/loadDUNS.py
```diff
@@ -0,0 +1,322 @@
...
+    daily = args.daily
+    benchmarks = args.benchmarks
+    update = args.update
+
+    with create_app().app_context():
+        configure_logging()
+        sess = GlobalDB.db().session
+
+        if monthly and daily:
+            print("For loading a single local file, you must provide either monthly or daily.")
```
alburde1: We don't want print statements. The logger should be sufficient. (apply to all below print statements)

dataactcore/migrations/versions/ff4728a82180_add_duns_table.py
```diff
@@ -33,7 +33,7 @@ def upgrade_data_broker():
     sa.Column('created_at', sa.DateTime(), nullable=True),
     sa.Column('updated_at', sa.DateTime(), nullable=True),
     sa.Column('duns_id', sa.Integer(), nullable=False),
-    sa.Column('awardee_or_recipient_uniqu', sa.Text(), nullable=True),
+    sa.Column('awardee_or_recipient_uniqu', sa.Text(), nullable=True, index=True),
```
alburde1: I think you still have to do the `op.create_index('ix_duns_awardee_or_recipient_uniqu', 'duns', ['awardee_or_recipient_uniqu'], unique=False)` thing. I'm not sure this works this way (we don't have it like this in any of our other migrations)

dataactcore/migrations/versions/c048f1dcdfa2_.py
```diff
@@ -1,14 +1,14 @@
-"""empty message
+"""merge ff4728a82180 and 967fbc28a6ef
```
alburde1: File needs renaming

PR #843, opened 2017-08-16
DB 2127 Fabs Publish information

kimminahm:
Adding publish information to the check_status route

### DIFF
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -658,9 +658,9 @@ def get_fabs_meta(submission_id):
 
     valid_rows = total_rows.filter(DetachedAwardFinancialAssistance.is_valid)
 
-    publish_date = sess.query(CertifyHistory).filter(CertifyHistory.submission_id == submission_id).first()
-    if publish_date:
-        publish_date = publish_date.created_at.strftime('%-I:%M%p %m/%d/%Y')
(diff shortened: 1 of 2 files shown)


dataactcore/interfaces/function_bag.py
```diff
@@ -649,6 +649,26 @@ def get_last_validated_date(submission_id):
...
+def get_fabs_meta(submission_id):
+
+    sess = GlobalDB.db().session
+
+    total_rows = sess.query(DetachedAwardFinancialAssistance).filter(
+                        DetachedAwardFinancialAssistance.submission_id == submission_id)
+
+    valid_rows = total_rows.filter(DetachedAwardFinancialAssistance.is_valid)
+
+    publish_date = sess.query(CertifyHistory).filter(CertifyHistory.submission_id == submission_id).first()
```
alburde1: I don't think we particularly want to remake the wheel. There's already a function called `get_lastest_certified_date` that you could just call here instead of duplicating code (this would also mean we wouldn't have to fix things in 2 locations if we updated anything about CertifyHistory). It also already returns the `created_at` part of the date

kimminahm: fixed. why is it called 'lastest'?

alburde1: Because someone (probably me) started typing "last" switched to "latest" have way and no one noticed the spelling error.

PR #858, opened 2017-08-28
2091 Triple Quote

kimminahm:
Remove triple quote when parsing file

### DIFF
--- a/dataactvalidator/filestreaming/fieldCleaner.py
+++ b/dataactvalidator/filestreaming/fieldCleaner.py
@@ -137,8 +137,9 @@ class FieldCleaner(StringCleaner):
                 # Remove extra whitespace
                 value = value.strip()
                 # If field is string and has triple quotes, remove. Single quote because csv converts triple to single
-                if field_type in ["STRING"] and value.startswith('"') and value.endswith('"'):
-                        value = value[1:-1]
+                if value.startswith('"') and value.endswith('"'):
+                    value = value[1:-1]
+                    value = value.strip()
                 if field_type in ["INT", "DECIMAL", "LONG"]:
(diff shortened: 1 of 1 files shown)


dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -136,6 +136,9 @@ def clean_row(cls, row, long_to_short_dict, fields):
             if value is not None:
                 # Remove extra whitespace
                 value = value.strip()
+                # If field is string and has triple quotes, remove. Single quote because csv converts triple to single
+                if field_type in ["STRING"] and value.startswith('"') and value.endswith('"'):
+                        value = value[1:-1]
```
alburde1: Just wondering, do we want to strip() again here?

Also, just making sure, does this work for any field? There's some chance that they'll wrap all fields in quotes and if it's an integer field we want to strip them anyway. There's no situation in which we wouldn't want to strip the surrounding quotes.

dataactvalidator/filestreaming/fieldCleaner.py
```diff
@@ -136,6 +136,10 @@ def clean_row(cls, row, long_to_short_dict, fields):
             if value is not None:
                 # Remove extra whitespace
                 value = value.strip()
+                # If field is string and has triple quotes, remove. Single quote because csv converts triple to single
+                if value.startswith('"') and value.endswith('"'):
+                    value = value[1:-1]
```
alburde1: could you just put `.strip()` at the end of this instead of making it a separate line? Or does that not work?

alburde1: Forgot to say it but comment is incorrect.

PR #859, opened 2017-08-29
HOTFIX: Update load_agencies function to fix SubTierAgency mapping

michaeldhess:


dataactvalidator/scripts/loadAgencies.py
```diff
@@ -139,7 +139,7 @@ def delete_missing_sub_tier_agencies(models, new_data):
 def update_sub_tier_agencies(models, new_data, cgac_dict, frec_dict):
     """Modify existing models or create new ones"""
     for _, row in new_data.iterrows():
-        if row['cgac_code'] not in cgac_dict or row['frec_code'] not in frec_dict:
+        if row['cgac_code'] not in cgac_dict:
```
alburde1: Why did we remove the frec check? Wouldn't we want to change it to an "and" instead of removing the check entirely?

dataactvalidator/scripts/loadAgencies.py
```diff
@@ -170,7 +170,7 @@ def load_sub_tier_agencies(file_name):
             data,
             SubTierAgency,
             {"cgac_agency_code": "cgac_code", "subtier_code": "sub_tier_agency_code", "priority": "priority",
-             "fpds_department_id": "frec_code", "subtier_name": "sub_tier_agency_name", "is_frec": "is_frec"},
+             "frec": "frec_code", "subtier_name": "sub_tier_agency_name", "is_frec": "is_frec"},
```
alburde1: Was this just incorrectly named or did something else change that didn't make it into this PR?

michaeldhess: Incorrectly mapped.

PR #863, opened 2017-08-31
Db 2188 fabs speedup

alburde1:
Major speedups for FABS31 rules. Smaller speedup for FABS39 (joining on a much smaller table for the same results)

Technical Release Notes:
- Updating SQL rules

### DIFF
--- a/dataactvalidator/config/sqlrules/fabs37_detached_award_financial_assistance_2.sql
+++ b/dataactvalidator/config/sqlrules/fabs37_detached_award_financial_assistance_2.sql
@@ -2,7 +2,7 @@
 -- This does not apply to correction records (those with CorrectionLateDeleteIndicator = C).
 -- Should not be active (action_date <= archived_date and when archived date exists)
 -- If the ActionDate is < published_date, should trigger a warning.
-WITH detached_award_financial_assistance_d37_2_{0} AS
+WITH detached_award_financial_assistance_fabs37_2_{0} AS
     (SELECT submission_id,
         row_number,
         cfda_number,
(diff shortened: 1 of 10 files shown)


dataactvalidator/config/sqlrules/fabs37_detached_award_financial_assistance_2.sql
```diff
@@ -17,16 +17,16 @@ SELECT
...
     AND ((dafa.correction_late_delete_ind != 'C')
         or (dafa.correction_late_delete_ind is null)
     )
     AND dafa.row_number IN (
         SELECT DISTINCT sub_dafa.row_number
-        FROM detached_award_financial_assistance_d37_2_{0} AS sub_dafa
+        FROM detached_award_financial_assistance_fabs37_2_{0} AS sub_dafa
             JOIN cfda_program AS cfda
-            ON (sub_dafa.cfda_number IS NOT DISTINCT FROM to_char(cfda.program_number, 'FM99D999')
+            ON (sub_dafa.cfda_number = to_char(cfda.program_number, 'FM00.000')
```
michaeldhess: Is there a reason that "FM99D999" was replaced with "FM00.000"? I can't find it in the acceptance criteria

alburde1: I had changed it in part 1 because of a bug where an entry of `12.340` would be read as `12.34` and that's not correct. The 9 indicates a digit and the 0 indicates a 0-filled value so it's ALWAYS that length. I realized while I was fixing this one for speed that I hadn't fixed the formatting. I don't remember why I changed D to . but they both mean a decimal point (I think the . is used more for 0-filled? There was research involved a month or two ago) so I figured we should be consistent.

PR #876, opened 2017-09-06
Db mk 2195 historical fabs derivations

kimminahm:
Adding relevant derivations to the historical fabs derivations


dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        return None
+    awarding_sub_tier = sess.query(SubTierAgency). \
+        filter_by(sub_tier_agency_code=row['agency_code']).one()
+    use_frec = awarding_sub_tier.is_frec
+    awarding_agency = awarding_sub_tier.frec if use_frec else awarding_sub_tier.cgac
+    return awarding_agency.frec_code if use_frec else awarding_agency.cgac_code
+
+
+def derive_awarding_agency_name(row, sess):
+    if not row['agency_code']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        return None
+    awarding_sub_tier = sess.query(SubTierAgency). \
+        filter_by(sub_tier_agency_code=row['agency_code']).one()
+    use_frec = awarding_sub_tier.is_frec
+    awarding_agency = awarding_sub_tier.frec if use_frec else awarding_sub_tier.cgac
+    return awarding_agency.agency_name
+
+
+def derive_awarding_sub_tier_agency_n(row, sess):
+    if not row['agency_code']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        return None
+    awarding_sub_tier = sess.query(SubTierAgency). \
+        filter_by(sub_tier_agency_code=row['agency_code']).one()
+    return awarding_sub_tier.sub_tier_agency_name
+
+
+def derive_place_of_perform_county_na(row, sess):
+    if not row['principal_place_zip']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        ppop_code = row['principal_place_code'].upper()
+        ppop_state = None
+        if ppop_code[:2] == '00':
+            return None
+        elif re.match('^[0-9]{2}', ppop_code[:2]):
```
alburde1: If we're only grabbing the first 2 characters, do we need the "start" character on this regex?

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+    zip_four = None
+    if len(row['principal_place_zip']) > 5:
+        zip_four = row['principal_place_zip'][-4:]
+    if zip_four:
+        zip_info = sess.query(Zips). \
+            filter_by(zip5=zip_five, zip_last4=zip_four).first()
+    else:
+        zip_info = sess.query(Zips). \
+            filter_by(zip5=zip_five).first()
+    county_info = sess.query(CountyCode). \
```
alburde1: Are we assuming this data and the zip data are in our DB? What if they didn't do a validation or something has changed since they validated and the zip or county isn't the same?

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        filter_by(county_number=zip_info.county_number, state_code=zip_info.state_abbreviation).first()
+    return county_info.county_name
+
+
+def derive_place_of_performance_city(row, sess):
+    if not row['principal_place_zip']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        ppop_code = row['principal_place_code'].upper()
+        ppop_state = None
+        if ppop_code[:2] == '00':
+            return None
+        elif re.match('^[0-9]{2}', ppop_code[:2]):
```
alburde1: If we're only grabbing the first 2 characters, do we need the "start" character on this regex?

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        elif re.match('^[A-Z]{2}\*\*\d{3}$', ppop_code[:2]):
+            ppop_state = sess.query(States).filter_by(state_code=ppop_code[:2]).one()
+        if re.match('\d{5}$', ppop_code) and not re.match('0{5}$', ppop_code):
+            # getting city and county name
+            city_code = ppop_code[-5:]
+            city_info = sess.query(CityCode).filter_by(city_code=city_code, state_code=ppop_state.state_code).first()
+            return city_info.feature_name
+        return None
+    zip_five = row['principal_place_zip'][:5]
+    city_info = sess.query(ZipCity).filter_by(zip_code=zip_five).one()
```
alburde1: Are we assuming this data and the zip data are in our DB? What if they didn't do a validation or something has changed since they validated and the zip or county isn't the same?

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+    return city_info.city_name
+
+
+def derive_legal_entity_state_name(row, sess):
+    if row['legal_entity_zip5']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+        zip_data = None
+        # if we have a legal entity zip+4 provided
+        if row['legal_entity_zip_last4']:
```
alburde1: We need to check for is_nan because we'll always have this column but it may be null (np.nan, not None)

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -177,6 +187,141 @@ def format_proper_casing(row, header):
...
+            zip_data = sess.query(Zips).\
+                filter_by(zip5=row['legal_entity_zip5'], zip_last4=row['legal_entity_zip_last4']).first()
+
+        # if legal_entity_zip_last4 returned no results (invalid combination), grab the first entry for this zip5
+        # for derivation purposes. This will exist because we wouldn't have gotten this far if it didn't,
```
alburde1: This comment is not accurate to this because this is historical data. We don't have this guarantee for this historical data because we didn't run any validations on this.

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -291,7 +423,7 @@ def set_active_rows(sess):
         "FROM published_award_financial_assistance GROUP BY afa_generated_unique) sub_pafa " +
         "ON pafa.modified_at = sub_pafa.modified_at AND " +
         "COALESCE(pafa.afa_generated_unique, '') = COALESCE(sub_pafa.afa_generated_unique, '') " +
-        "WHERE COALESCE(UPPER(pafa.correction_late_delete_ind), '') != 'D' AS selected " +
+        "WHERE COALESCE(UPPER(pafa.correction_late_delete_ind), '') != 'D') AS selected " +
```
alburde1: Why is there an extra parenthesis here? Wouldn't this break the SQL?

dataactcore/scripts/loadHistoricalFabs.py
```diff
@@ -324,6 +456,10 @@ def main():
     logger.info("Historical FABS script complete")
 
 
+def is_nan(num):
```
alburde1: What is this doing? Pandas has an `isnull()` function, not sure if that works for single values but you could take a look?

dataactcore/models/stagingModels.py
```diff
@@ -879,6 +880,7 @@ class PublishedAwardFinancialAssistance(Base):
     funding_agency_name = Column(Text)
     funding_agency_code = Column(Text, index=True)
     funding_office_code = Column(Text, index=True)
+    funding_office_name = Column(Text, index=True)
```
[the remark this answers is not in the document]
kimminahm: this one isnt. i think i misclicked the copy paste

dataactcore/models/stagingModels.py
```diff
@@ -887,6 +889,7 @@ class PublishedAwardFinancialAssistance(Base):
     legal_entity_address_line2 = Column(Text)
     legal_entity_address_line3 = Column(Text)
     legal_entity_city_name = Column(Text)
+    legal_entity_city_code = Column(Text, index=True)
```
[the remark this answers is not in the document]
kimminahm: i guess not. removing

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: pd.isnull(val) works

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: No, mike accidentally deleted it when copy pasting before so we had to add it back in to work

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: just the second part is invalid right?

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: no we dont

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: done

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: done

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: fixed

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: no

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: done

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: done

dataactcore/scripts/loadHistoricalFabs.py
kimminahm: done

PR #894, opened 2017-09-14
DB-2210 fix DUNS loader

michaeldhess:


dataactcore/scripts/loadDUNS.py
```diff
@@ -70,35 +69,35 @@ def activation_check(data, activated_models, benchmarks=False):
...
-        update_duns_start = time.time()
-    for _, row in new_data.iterrows():
-        awardee_or_recipient_uniqu = row['awardee_or_recipient_uniqu']
-        if awardee_or_recipient_uniqu not in models:
-            models[awardee_or_recipient_uniqu] = DUNS()
-        for field, value in row.items():
-            setattr(models[awardee_or_recipient_uniqu], field, value)
-    if benchmarks:
-        logger.info("Updating duns took {} seconds".format(time.time() - update_duns_start))
+# Removed this function when adding registration_date
```
alburde1: If we removed this function why are we keeping it?

michaeldhess: We're keeping it because they changed their minds about what should be pulled, so we decided to make sure we keep the previous changes in case this changes once more.

PR #940, opened 2017-09-22
Db 2234 fixing fabs derivations

alburde1:
Fixing existing and potential future issues with derivations and removing a few extraneous queries (hopefully that makes them run at least a little faster)

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -771,6 +771,7 @@ class FileHandler:
         """
         sess = GlobalDB.db().session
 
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -771,6 +771,7 @@ def check_detached_generation(job_id):
         """
         sess = GlobalDB.db().session
 
+        # TODO is this comment wrong or is the code wrong?
```
michaeldhess: Looks like the comment is wrong. This may also be changed in the D file generation so we'll check it out when we merge.

PR #942, opened 2017-09-25
DB-MK-2172 Funding Agency Derivation

kimminahm:
Technical Release Notes
- funding_agency_code is no longer a used field in FABS submissions
- rule FABS20 has been deleted


kimminahm:
Technical Release Notes
-funding_agency_code is no longer a used field in FABS submissions
-rule FABS20 has been deleted

tests/unit/dataactbroker/test_fabs_derivations.py
```diff
@@ -136,7 +135,7 @@ def test_funding_agency_name(database):
     assert obj['funding_agency_name'] is None
 
     # when funding_agency_code is provided
```
alburde1: This comment is no longer correct. This entire test is no longer needed actually because the test below it covers all those bases. I can't comment down there but `funding_agency_code ` and `funding_agency_name` should be tested down there too. It could probably be renamed to `test_funding_agency_derivations` or something

tests/unit/dataactbroker/test_fabs_derivations.py
```diff
@@ -255,7 +250,7 @@ def test_primary_place_country(database):
 
     obj = initialize_test_obj(primary_place_country='NK')
     obj = fabs_derivations(obj, database.session)
-    assert obj['place_of_perform_country_n'] is None
+    assert not obj['place_of_perform_country_n']
```
alburde1: These should stay `is None`, extend this comment to the rest of these situation

tests/unit/dataactbroker/test_fabs_derivations.py
```diff
@@ -294,7 +289,23 @@ def test_legal_country(database):
 
     obj = initialize_test_obj(legal_country='NK')
     obj = fabs_derivations(obj, database.session)
-    assert obj['legal_entity_country_name'] is None
+    assert not obj['legal_entity_country_name']
+
+
+def test_primary_place_county(database):
```
alburde1: We're already deriving/testing `place_of_perform_county_na` in a higher test (`test_ppop_derivations`) so this new test is unnecessary. You also removed all the test lines from that function for `place_of_perform_county_co` so they need to go back in.

PR #975, opened 2017-10-03
DB-MK-2172 Funding Agency Derivation

kimminahm:
Technical Release Notes
- funding_agency_code is no longer a used field in FABS submissions but is derived
- rules FABS20 has been removed


tests/unit/dataactbroker/test_fabs_derivations.py
```diff
@@ -136,7 +135,7 @@ def test_funding_agency_name(database):
     assert obj['funding_agency_name'] is None
 
     # when funding_agency_code is provided
-    obj = initialize_test_obj(fund_agency_code="000")
+    obj = initialize_test_obj(sub_fund_agency_code="1234")
```
alburde1: This test should just be removed completely because this was a test for funding_agency_code, which we removed. The test below it (which I can't comment on directly) should have the derivations for `funding_agency_name` and `funding_agency_code` added to its tests because that's where they're being derived now. The name of the below test should also be changed to reflect these changes (`funding_agency_derivations` for example)

PR #1003, opened 2017-10-10
DB-2112 File name to submission_list response

kimminahm:
Technical Release Notes:
- submission_list route now adds file names to the response


dataactbroker/handlers/fileHandler.py
```diff
@@ -1268,10 +1270,12 @@ def list_submissions(page, limit, certified, sort='modified', order='desc', d2_s
 
     query = query.order_by(sort_order)
 
-    query = query.limit(limit).offset(offset)
+    submission_list = serialize_submission(query)
```
michaeldhess: This means we're querying for **all** submissions every time, which is very inefficient. This will cause issues when we have hundreds, then thousands of submissions.

kimminahm:
[run tests]

PR #1013, opened 2017-10-25
DB-2295 cached D file generations

michaeldhess:
AC:
- D files generated with same agency code, date range, and on the same day are cached
- D2 files are uncached when changes are made to the contents through FABS
- Cached files are copied into their corresponding submission buckets in S3, when within a submission


dataactbroker/handlers/fileGenerationHandler.py
```diff
@@ -58,57 +77,90 @@ def generate_d_file(file_type, agency_code, start, end, job_id, file_name, uploa
...
+                copy_parent_file_request_data(sess, file_request.job, parent_req.job, file_type, is_local)
+        else:
+            # no cached file
+            # start D file generation
+            logger.debug('Starting file {} generation'.format(file_type))
+            file_utils = fileD1 if file_type == 'D1' else fileD2
+            full_file_path = "".join([CONFIG_BROKER['d_file_storage_path'], file_name])
+            page_idx = 0
 
-        try:
```
alburde1: Why did we remove the try/finally?

michaeldhess: Including that `try/finally` caused a bug that I found. If it threw an error during that period, it wouldn't reach the error handling in `job_context()`, which means the job wouldn't ever fail correctly and it would run forever.

PR #1015, opened 2017-10-27
Db 2288 speed up flex fields

alburde1:
Updating how we gather and write flex fields so it takes less time (new SQL query) and doesn't crash the entire application non-locally (gather them in chunks so we don't destroy the memory). Also updating rule B9 so it takes far less time. Writing error/warning files to S3 only after the entire file has been created temporarily in memory so we don't have an S3 connection open forever while the files validate.

- [x] Approved by @klrBAH 
- [ ] Reviewed by backend
- [x] Config repo updated

Technical Release Notes:
- Updates to rule B9
- Config updates
- Updates to how we handle file writing for error/warning files on SQL rules


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -509,30 +544,57 @@ def run_cross_validation(self, job):
...
-                writer.finish_batch()
-                warning_writer.finish_batch()
+
+            # get error file name/path
+            error_file_name = report_file_name(submission_id, False, first_file.name, second_file.name)
+            error_file_path = "".join([CONFIG_SERVICES['error_report_path'], error_file_name])
+            warning_file_name = report_file_name(submission_id, True, first_file.name, second_file.name)
+            warning_file_path = "".join([CONFIG_SERVICES['error_report_path'], warning_file_name])
+
+        # open error report and gather failed rules within it
```
michaeldhess: Tab this over

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -509,30 +544,57 @@ def run_cross_validation(self, job):
...
+            with open(error_file_path, 'w', newline='') as error_file,\
```
michaeldhess: This seems like duplicate code. Can we separate this out into its own function?

dataactvalidator/validation_handlers/validator.py
```diff
@@ -185,25 +192,61 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict, first_file, sec
...
-                failures.append([rule.file.name, target_file_type, full_column_string,
-                                str(rule.rule_error_message), values, row['row_number'], str(rule.rule_label),
-                                rule.file_id, rule.target_file_id, rule.rule_severity_id])
+            num_failed_rows = len(failed_rows)
+            slice_start = 0
+            slice_size = 10000
+            while slice_start <= num_failed_rows:
+                failed_row_subset = failed_rows[slice_start:slice_start+slice_size]
+                # finding out row numbers for logger
+                last_error_curr_slice = slice_start + slice_size
```
michaeldhess: Might as well move this up one and use it in the failed_row_subset calculation, to only do the addition once

dataactvalidator/validation_handlers/validationManager.py
alburde1: No, because they do very different things within the `with`

PR #1041, opened 2017-11-14
Db 2232 fabs flex fields all errors

alburde1:
Adding flex fields to all errors, not just SQL ones. Making the errors more uniform/grammatically correct and adding periods to the end of all of them.

Technical Release notes:
- Changes to error text


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -744,12 +745,28 @@ def write_errors(failures, job, short_colnames, writer, warning_writer, row_numb
...
+        # get flex fields
+        field_names = [field_name]
+        flex_list = []
+        # only add the value if there's something to add, otherwise our join will look bad
+        if failure.value:
+            flex_list = [field_name + ": " + failure.value]
+
+        # if there are any flex columns, append them to the field names and values
+        if flex_cols:
+            for flex_col in flex_cols:
```
michaeldhess: Can you make this before the previous `for` loop (line 733)? Your way creates this list for each `failure`, but it doesn't necessarily need to.

You should be able to loop through the `flex_cols` higher up to create an array of `flex_col.header`s and another array `flex_col.header: flex_col.cell`s, then just `extend` the `field_names` and `flex_list` variables within this `for` loop.

PR #1044, opened 2017-11-16
Db 1195 sql rule comments

alburde1:
Updating all SQL rules with comments and standardizing SQL style. Includes a style guide for SQL.

Technical Release Notes:
- Updates to all SQL rules (nothing should have changed in performance but they all should be reloaded anyway)


dataactvalidator/config/sqlrules/c21_award_financial.sql
```diff
@@ -113,8 +116,7 @@ FROM (
 -- The second subquery selects the sum of the corresponding 32 elements in File B
 -- Again, the sum is based on TAS, PAC, and Submission ID
 -- We do a FULL OUTER JOIN of this result, as we don't care if TAS/PAC combindations from File B aren't in File C
```
michaeldhess: "combindations"

PR #1062, opened 2017-12-08
Ds 1984 derive county info fpds loader

alburde1:
- [x] Cleared by @klrBAH 
- [x] Reviewed by backend

Deriving county code in nightly FPDS pull. Also fixing a little bit of logic and a logger printing error. Adding relevant columns.

Technical Release Notes:
- alembic migrations


dataactcore/scripts/pullFPDSData.py
```diff
@@ -59,6 +64,33 @@ def extract_text(data_val):
     return data_val
 
 
+def is_valid_zip(zip_code):
+    if re.match('^\d{5}(-?\d{4})?$', zip_code):
+        return True
+    return False
+
+
+def get_county_by_zip(sess, zip_code):
```
michaeldhess: `is_valid_zip` accounts for hyphenated zips (00000-0000), but it would never get past any of the `if` statements including `len(zip_code)` because they only allow a length of 9 or 5

alburde1: See? This is why I need extra sets of eyes. I changed how I handled the checking if it was valid part and didn't change the length checks.

PR #1065, opened 2017-12-13
DS 1977 Update Awarding/Funding agency for detached_award_procurement

kimminahm:


dataactbroker/scripts/update_detached_procurement.py
```diff
@@ -0,0 +1,54 @@
...
+from dataactcore.models.userModel import User # noqa
+from dataactvalidator.health_check import create_app
+
+logger = logging.getLogger(__name__)
+
+
+def update_table(sess, type):
+    logger.info('updating ' + type)
+    invalid = sess.execute("select * from detached_award_procurement where " + type + "_agency_code='999'")
+    invalid_count = len(invalid.fetchall())
```
alburde1: Do we really want to do a fetchall() here? I believe that actually grabs all the rows. Is there any need for that since we don't use them? It might make more sense to just run a count query, especially over selecting * from the FPDS table (which has WAY too many columns)

dataactbroker/scripts/update_detached_procurement.py
```diff
@@ -0,0 +1,54 @@
+import logging
+
+from dataactcore.logging import configure_logging
+from dataactcore.interfaces.db import GlobalDB
+from dataactcore.models.jobModels import Submission # noqa
+from dataactcore.models.userModel import User # noqa
```
alburde1: If we aren't importing any other models (now that we're just doing raw sql here), does it still throw errors without Submission and User? It might very well be that it does, but if it doesn't then we should take these imports out.

dataactbroker/scripts/update_detached_procurement.py
```diff
@@ -0,0 +1,54 @@
...
+        "from sub_tier_agency sub " +
+        "INNER JOIN cgac ON cgac.cgac_id = sub.cgac_id " +
+        "INNER JOIN frec ON frec.frec_id = sub.frec_id ) agency " +
+        "where detached_award_procurement." + type + "_agency_code = '999' " +
+        "and detached_award_procurement." + type + "_sub_tier_agency_c" + suffix +
+        " = agency.sub_tier_agency_code "
+    )
+    sess.commit()
+    invalid = sess.execute("select * from detached_award_procurement where " + type + "_agency_code='999'")
+    print_report(invalid_count, len(invalid.fetchall()), type)
```
alburde1: Same question here for the fetchall(), can we just run a count instead?

alburde1:
Another note, it might be nice to have some comments as to what each portion is doing. It's pretty self-explanatory (counting for logging, updating, etc) but if we come back to this in a year (or someone else does) it'll be nice to know how/why things are happening quickly.

dataactbroker/scripts/update_detached_procurement.py
kimminahm: Good point. seems like it is not necessary because we use raw sql instead of sqlalchemy there is no need for the models

dataactbroker/scripts/update_detached_procurement.py
kimminahm: changed it to `count(*)`

PR #1073, opened 2018-01-03
Ds 1978 1980 historical fpds fabs county update

alburde1:
Scripts for deriving location data for historical fabs and fpds data

Country code/name
State code/name
County code/name
Fixing congressional district problems


dataactcore/scripts/deriveHistoricalAssistanceLocation.py
```diff
@@ -0,0 +1,592 @@
...
+    # have to
+    if not row.legal_entity_county_code or not row.legal_entity_county_name:
+
+        zip_data = get_zip_data(sess, row.legal_entity_zip4)
+        if zip_data:
+            if not row.legal_entity_county_code:
+                row.legal_entity_county_code = zip_data.county_number
+
+            # if we got the zip data and have a state code to work with, if it's valid then grab the county name
+            if not row.legal_entity_county_name and row.legal_entity_state_code:
```
nmonga91: Here, we need the state code + county code to be able to derive the county name since county code isn't unique enough on its own, right?

dataactcore/scripts/deriveHistoricalAssistanceLocation.py
```diff
@@ -0,0 +1,592 @@
...
+                state = row.legal_entity_state_code.upper()
```
nmonga91: This is slightly nit-picky but if we're referring to the state code, it'd be more clear for the variable to be named `state_code` so the referenced value is 100% clear.

dataactcore/scripts/deriveHistoricalAssistanceLocation.py
```diff
@@ -0,0 +1,592 @@
...
+            and row.place_of_perform_county_co in county_by_code[state]:
+        row.place_of_perform_county_na = county_by_code[state][row.place_of_perform_county_co]
+
+
+def process_fpds_derivations(sess, country_list, state_by_code, state_code_by_fips, county_by_code, county_by_name,
+                             data):
+    """ Process derivations for FPDS location data """
+    for row in data:
+        # Don't update the updated_at timestamp
+        row.ignore_updated_at = True
```
nmonga91: Liked seeing this in there since we got it to work 💯

dataactcore/scripts/deriveHistoricalAssistanceLocation.py
alburde1: Yes, but if we got zip data we will always have county code. If it didn't exist before we set it above and if it did then it was already there. Because it's below the setting of the county code, it will never be blank.

PR #1090, opened 2018-01-19
Hotfix speed up location derivations script

alburde1:
Size saving and speeding up for the location derivations on FPDS data


alburde1:
[run tests]

PR #1113, opened 2018-02-05
Hotfix speed up location derivations script

alburde1:
- Create fixes to speed up the location derivation script.
- Create a new script that is pure SQL to derive the county data that wasn't properly derived by the script for whatever reason.


dataactcore/scripts/fixCountyData.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def update_fabs_le(sess):
+    logger.info("Starting FABS legal entity derivations, starting FABS legal entity 9-digit zips")
+
+    # FABS LE 9-digit
+    sess.execute(
+        """UPDATE published_award_financial_assistance AS pafa
+            SET legal_entity_county_code = zc.county_number,
+                legal_entity_county_name = CASE WHEN pafa.legal_entity_county_name IS NOT NULL
```
shinson: General Question: If we are only updating county names that are null, is there any checks if county name is incorrectly formatted?

PR #1115, opened 2018-02-06
Add index to created_at in FABS and updated_at in FPDS

michaeldhess:
FABS = `published_award_financial_assistance`
FPDS = `detached_award_procurement`


dataactcore/migrations/versions/7597deb348fb_fabs_created_at_and_fpds_updated_at_.py
```diff
@@ -0,0 +1,42 @@
+"""FABS created_at and FPDS updated_at indexes
+
```
nmonga91: Do we want to add one for `correction_late_delete_ind`?

michaeldhess: Already exists:```-- Index: public.ix_published_award_financial_assistance_correction_late_fa61

-- DROP INDEX public.ix_published_award_financial_assistance_correction_late_fa61;

CREATE INDEX ix_published_award_financial_assistance_correction_late_fa61
  ON public.published_award_financial_assistance
  USING btree
  (correction_late_delete_ind COLLATE pg_catalog."default");```

PR #1124, opened 2018-02-09
Hotfix historical duns

shinson:
Updating script to update DUNS table for historical agencies to make changes found through testing.

 [Story Link](https://federal-spending-transparency.atlassian.net/browse/DB-2235)

Performance notes on running script on test environment:
Time to update 10k rows to DUNS table: 1-5s 

The file contains over 4 million rows which equates to 409 chunks. It took 17 mins to finish on the test environment.

### DIFF
--- a/dataactbroker/scripts/update_historical_duns.py
+++ b/dataactbroker/scripts/update_historical_duns.py
@@ -56,7 +56,7 @@ def run_duns_batches(file, sess, block_size=10000):
     logger.info("Retrieved row count of {} in {} s".format(row_count, (datetime.now()-start).total_seconds()))
 
     duns_reader_obj = pd.read_csv(file, skipinitialspace=True, header=None,  encoding='latin1', quotechar='"',
-                          dtype=str, names=column_headers, iterator=True, chunksize=block_size, skiprows=1)
+                                  dtype=str, names=column_headers, iterator=True, chunksize=block_size, skiprows=1)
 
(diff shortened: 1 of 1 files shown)


dataactbroker/scripts/update_historical_duns.py
```diff
@@ -47,24 +47,19 @@ def clean_duns_csv_data(data):
...
                                 dtype=str, names=column_headers, skiprows=1))
     logger.info("Retrieved row count of {} in {} s".format(row_count, (datetime.now()-start).total_seconds()))
 
-    batches = row_count // block_size
+    duns_reader_obj = pd.read_csv(file, skipinitialspace=True, header=None,  encoding='latin1', quotechar='"',
+                          dtype=str, names=column_headers, iterator=True, chunksize=block_size, skiprows=1)
 
-    while batch <= batches:
-        logger.info("Begin updating duns batch {} ".format(batch))
+    for duns_df in duns_reader_obj:
```
nmonga91: For my understanding, `duns_reader_obj` is an iterator over the CSV file, where each `duns_df` is a chunk of data from the CSV where the chunk size = `block_size`. Is that a correct understanding?

shinson: Yes that's correct, Pandas returns a `TextFileReader object` to iterate, and each chuck is a dataframe.

PR #1153, opened 2018-03-12
Dev 436 fpds 1.5

alburde1:
Updating FPDS nightly loader and columns based on DEV-436 and DEV-438 (this was determined to be a non-action)

Technical Release Notes:
- Config CSVs updated
- Alembic migration


tests/integration/data/awardProcValid.csv
```diff
@@ -1,11 +1,11 @@
...
...n_owned_small_business,economically_disadvantaged_women_owned_small_business,joint_venture_women_owned_small_business,joint_venture_economically_disadvantaged_women_owned_small_business,veteran_owned_business,service_disabled_veteran_owned_business,contracts,grants,receives_contracts_and_grants,airport_authority,council_of_governments,housing_authorities_public_tribal,interstate_entity,planning_commission,port_authority,transit_authority,subchapter_scorporation,limited_liability_corporation,foreign_owned_and_located,american_indian_owned_business,alaskan_native_owned_corporation_or_firm,indian_tribe_federally_recognized,native_hawaiian_owned_business,tribally_owned_business,asian_pacific_american_owned_business,black_american_owned_business,hispanic_american_owned_business,native_american_owned_business,subcontinent_asian_asian_indian_american_owned_business,other_minority_owned_business,for_profit_organization,nonprofit_organization,other_not_for_profit_organization,us_local_government,referenced_idv_modification_number,undefinitized_action,domestic_or_foreign_entity,cage_code,inherently_government_func,organizational_type,number_of_employees,annual_revenue,total_dollars_obligated
```
shinson: I don't see `government_furnished_prope` as  a field in the csv

dataactcore/models/stagingModels.py
```diff
@@ -509,6 +509,12 @@ class AwardProcurement(Base):
     place_of_perform_county_na = Column(Text)
     base_exercised_options_val = Column(Text)
     base_and_all_options_value = Column(Text)
+    cage_code = Column(Text)
+    inherently_government_func = Column(Text)
+    organizational_type = Column(Text)
+    number_of_employees = Column(Text)
+    annual_revenue = Column(Text)
+    total_obligated_amount = Column(Text)
```
shinson: Does broker need to match the DAIMS schema v1.2 IDD. under their document this field has a Terse 30 Label `total_dollars_obligated`

Others that are different:
```
- inherently_governmental_funct
- materials_supplies_articles_e
- construction_wage_rate_reqs
```

alburde1: It was already not matching to begin with so we left it (it was only added to AwardProcurement, it was already in DetachedAwardProcurement). We also left `gfe_gfp` when we renamed the full label for it (after talking to Kristine) and every single description tag that we have doesn't end in `tag` like the IDD does.

Additionally, the terse labels go up to 30 characters, with a good portion being 27-29 (like all of the ones you listed). Our standards for column names is 26 characters for broker so a ton will never match no matter what we do. The users never see these columns because we map them to something completely different on our end before they're printed to a D1 file.

PR #1158, opened 2018-03-23
Dev 450 fabsreq fabstype labels

alburde1:
Adding FABSTYPE and FABSREQ as numbered error labels to the error report


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -314,7 +327,8 @@ def run_validation(self, job):
                                                              "_" + (record['fain'] or '-none-') + "_" + \
                                                              (record['uri'] or '-none-')
                         passed_validations, failures, valid = Validator.validate(record, csv_schema,
-                                                                                 file_type in ["detached_award"])
+                                                                                 file_type in ["detached_award"],
```
shinson: Can this be written as `file_type == "detached_award"`?

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -314,7 +327,8 @@ def run_validation(self, job):
                                                              "_" + (record['fain'] or '-none-') + "_" + \
                                                              (record['uri'] or '-none-')
                         passed_validations, failures, valid = Validator.validate(record, csv_schema,
```
shinson: Is there any way that Validator.validate() is called that isn't when `file_type` is not  `detached_award`?

alburde1: Yes, it's called for every DABS validation, which is why `required_list` and `type_list` are only populated if it's a FABS submission (`detached_award` type) and in the validate function itself it checks if they exist before altering the error report

PR #1159, opened 2018-03-23
Update dev 441

shinson:
[Bug Link](https://federal-spending-transparency.atlassian.net/browse/DEV-441)

Adding rows and updating loading process for File E and File F

File E: Add
1. AwardeeOrRecipientLegalEntityName

File F: Add
1. LegalEntityStateName
2. PrimaryPlaceOfPerformanceStateName

Technical Release Notes:
- Updated config CSVs


dataactcore/migrations/versions/c75d57250419_.py
```diff
@@ -0,0 +1,49 @@
+"""empty message
```
alburde1: There should always be a message in migrations so we know what's happening. The name of the migration file should be a portion of this message. Here it can be "Adding state name columns to all fsrs tables" or something

dataactcore/migrations/versions/c75d57250419_fileF_add_state_name.py
```diff
@@ -0,0 +1,49 @@
+"""Adding schema v1.2 F columns
```
alburde1: Generally, when adding these messages, we don't want to say "F columns" or something because that's not really indicative of the database. We want to stick to facts about the migration like "Adding state name columns to all fsrs tables" or something.

dataactcore/migrations/versions/ee7bff1d660c_fileE_add_legal_entity_name.py
```diff
@@ -0,0 +1,33 @@
+"""Adding column lega entity name schema v1.2 File E
```
alburde1: Same as the F file one, but this one can be more specific since it's just one column (somehow GitHub didn't keep this comment last time)

dataactbroker/fsrs.py
```diff
@@ -137,6 +149,14 @@ def flatten_soap_dict(simple_fields, address_fields, comma_field, soap_dict):
     for prefix in address_fields:
         for field in ('city', 'street', 'state', 'country', 'zip', 'district'):
             model_attrs[prefix + '_' + field] = soap_dict[prefix].get(field)
+
+            # Deriving state name since not provided by FSRS feed
+            if field == 'state':
+                # Only populate for USA locations
+                if soap_dict[prefix].get('country') and soap_dict[prefix]['country'].upper() == 'USA':
+                    model_attrs[prefix + '_' + field + '_name'] = g_state_by_code.get(model_attrs[prefix + '_state'])
```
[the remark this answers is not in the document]
shinson: Fixed

dataactbroker/fsrs.py
```diff
@@ -137,6 +149,14 @@ def flatten_soap_dict(simple_fields, address_fields, comma_field, soap_dict):
...
+                else:
+                    model_attrs[prefix + '_' + field + '_name'] = None
```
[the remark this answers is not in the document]
shinson: Fixed

dataactcore/migrations/versions/c75d57250419_fileF_add_state_name.py
shinson: Updated Description and file name

dataactcore/migrations/versions/ee7bff1d660c_fileE_add_legal_entity_name.py
shinson: Updated description and file name

dataactbroker/fsrs.py
```diff
@@ -31,6 +32,17 @@ def config_valid():
     return bool(proc_wsdl) and bool(grant_wsdl)
 
 
+def get_state_mappings(sess):
+    """ Creates dictionary that maps state code to state name, deletes global variable if already processed"""
```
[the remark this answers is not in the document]
shinson: `del` deletes any variable (local/global) from reference. calling states after the `del` throws an error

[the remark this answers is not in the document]
alburde1: Right, my point was that you aren't deleting the global variable and the comment says you are. You're only deleting the local variable after you're done working with it, the global remains.

PR #1164, opened 2018-03-29
Dev 427 ignore different ata

alburde1:
Updating rules C8, C11, and C23 to be skipped not when the row has an ATA (valid or not) at all, but if the ATA and AID do not match.

Technical Release Notes:
- Updated SQL rules


tests/unit/dataactvalidator/test_c23_award_financial_1.py
```diff
@@ -14,55 +14,86 @@ def test_column_headers(database):
 
 
 def test_success(database):
-    """ Test that a four digit object class with no flag is a success, and a three digit object class with
-        a flag is a success. Ignores rows with parent_award_id from AwardFinancialFactory and doesn't care about
-        parent_award_id in AwardProcurementFactory """
+    """ Test For each unique PIID in File C, the sum of each TransactionObligatedAmount should match (but with opposite
```
shinson: would be helpful to understand tests if comment included that Award Financial entries that have a parent id , will be ignored.

tests/unit/dataactvalidator/test_c8_award_financial.py
```diff
@@ -15,130 +14,127 @@ def test_column_headers(database):
 
 
 def test_equal_fain(database):
-    """Tests that File C (award financial) fain matches
-    File D2 (award financial assistance) fain."""
+    """ Tests that File C (award financial) fain matches File D2 (award financial assistance) fain. """
     tas = _TAS
     af = AwardFinancialFactory(tas=tas, fain='abc', uri=None, allocation_transfer_agency=None)
-    afa = AwardFinancialAssistanceFactory(tas=tas, submisson_id=af.submission_id, fain=af.fain, uri=None)
+    afa = AwardFinancialAssistanceFactory(tas=tas, submisson_id=af.submission_id, fain=af.fain, uri=None,
```
shinson: Not sure if there is a standard on this: but other tests in the review specifically for D1, would write out the identifier (in this case the fain field), so instead of `af.fain` it would be `abc`

tests/unit/dataactvalidator/test_c23_award_financial_1.py
shinson: Fixed

PR #1167, opened 2018-04-02
Dev 442 new d1 elements

alburde1:
Adding way too many columns
Adding description columns to D1 and fixing a few other columns. Adding
a couple more columns to the FPDS nightly load

Technical Release Notes:
- Alembic upgrades
- Config CSV updates


dataactcore/utils/fileD1.py
```diff
@@ -241,7 +294,7 @@ def query_data(session, agency_code, start, end, page_start, page_stop):
 
 
 def initial_query(session):
-    return session.query(
+    return session.query(*[
```
shinson: What's the added benefit for using this notation?

alburde1: There are more than 255 columns we're querying, which essentially means more than 255 parameters for this function, which Python does not allow (throws a syntax error). This way, it's one parameter which is getting unwrapped after it's passed in.

PR #1168, opened 2018-04-03
adding retries in get_data to log statements, expanding exception scope

willkjackson:
This PR adds ReadTimeoutError and ConnectionError to the FPDS retries. Currently we were only checking for ConnectionResetError. The error I've been seeing during the nightly job for this is requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')), which would not be caught (or retried) and therefore immediately fail in the current version of the script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # DATA Act Broker Backend
 
-The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform. 
+The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform.
 
 The website that runs on these services is here: [https://broker.usaspending.gov/](https://broker.usaspending.gov/ "DATA Act Broker website").
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # DATA Act Broker Backend
 
-The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform.
+The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform. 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # DATA Act Broker Backend
 
-The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform.
+The DATA Act broker backend is a collection of services that power the DATA Act's central data submission platform.
```
alburde1: Why was this added?

dataactcore/scripts/pullFPDSData.py
```diff
@@ -1340,12 +1340,26 @@ def get_delete_data(contract_type, now, sess, last_run, start_date=None, end_dat
     i = 0
     logger.info('Starting delete feed: %sCONTRACT_TYPE:"%s"', delete_url + params, contract_type.upper())
     while True:
-        resp = requests.get(delete_url + params + 'CONTRACT_TYPE:"' + contract_type.upper() + '"&start=' + str(i),
-                            timeout=60)
-        resp_data = xmltodict.parse(resp.text, process_namespaces=True,
-                                    namespaces={'http://www.fpdsng.com/FPDS': None,
-                                                'http://www.w3.org/2005/Atom': None,
-                                                'https://www.fpds.gov/FPDS': None})
+        exception_retries = -1
```
alburde1: Are we getting timeouts in the delete feed too?

willkjackson: Yeah, I got them while testing these changes.

PR #1182, opened 2018-04-11
Dev 578 adjust c9

alburde1:
Adjusting rule C9 to account for assistance type and changing which lines from D2 are ignored based on assistance type and amount

Technical Release Notes:
- Updated SQL rule

### DIFF
--- a/tests/unit/dataactvalidator/test_c9_award_financial.py
+++ b/tests/unit/dataactvalidator/test_c9_award_financial.py
@@ -120,11 +120,21 @@ def test_zero_federal_action_obligation_and_positive_original_loan_subsidy_cost(
 
     tas = _TAS
     afa = AwardFinancialAssistanceFactory(tas=tas, fain='abc', uri=None, federal_action_obligation=0,
-                                          original_loan_subsidy_cost='1')
+                                          original_loan_subsidy_cost='1', assistance_type='08')
+    afa_2 = AwardFinancialAssistanceFactory(tas=tas, fain='abc', uri=None, federal_action_obligation=2,
+                                            original_loan_subsidy_cost='1', assistance_type='07')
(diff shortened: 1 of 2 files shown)


tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -120,11 +120,21 @@ def test_zero_federal_action_obligation_and_positive_original_loan_subsidy_cost(
...
+    afa = AwardFinancialAssistanceFactory(tas=tas, fain='abc', uri=None, federal_action_obligation=0,
+                                          original_loan_subsidy_cost='1', assistance_type='07')
     af = AwardFinancialFactory(tas=tas, submisson_id=afa.submission_id, fain=None, uri=None)
 
     errors = number_of_errors(_FILE, database, models=[afa, af])
-    assert errors == 1
+    assert errors == 0
 
 
 def test_positive_federal_action_obligation_and_zero_original_loan_subsidy_cost(database):
```
shinson: This function's name makes it seem that you are only testing a positive federal action obligation and a 0 loan subsidy cost, but you are testing loan costs of negative and positive values.

tests/unit/dataactvalidator/test_c9_award_financial.py
```diff
@@ -120,11 +120,21 @@ def test_zero_federal_action_obligation_and_positive_original_loan_subsidy_cost(
 
     tas = _TAS
     afa = AwardFinancialAssistanceFactory(tas=tas, fain='abc', uri=None, federal_action_obligation=0,
-                                          original_loan_subsidy_cost='1')
+                                          original_loan_subsidy_cost='1', assistance_type='08')
+    afa_2 = AwardFinancialAssistanceFactory(tas=tas, fain='abc', uri=None, federal_action_obligation=2,
```
shinson: Function name makes it seem that it is only testing federal action obl with a value of 0, but here fed action obl has a value of 2.

PR #1185, opened 2018-04-11
Update dev 249 job type

shinson:
Removing `db_transfer` and `external_validation` from broker.
- Removed from job types lookup
- Added functionality to `initialize.py --db` to remove job types and related jobs and job dependencies
- Removed `external_validation` job creation from submission process 

[Story Link](https://federal-spending-transparency.atlassian.net/browse/DEV-249)

### DIFF
--- a/dataactbroker/handlers/submission_handler.py
+++ b/dataactbroker/handlers/submission_handler.py
@@ -93,7 +93,6 @@ def get_submission_status(submission, jobs):
         string containing the status of the submission"""
     status_names = JOB_STATUS_DICT.keys()
     statuses = {name: 0 for name in status_names}
-    skip_count = 0
 
     for job in jobs:
         job_status = job.job_status.name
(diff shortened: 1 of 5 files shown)


dataactbroker/handlers/submission_handler.py
```diff
@@ -96,11 +96,8 @@ def get_submission_status(submission, jobs):
     skip_count = 0
 
     for job in jobs:
-        if job.job_type.name not in ["external_validation", None]:
```
alburde1: Was there a reason we were checking for the `None`? Do we want to keep checking for it?

dataactcore/interfaces/function_bag.py
```diff
@@ -372,46 +371,19 @@ def create_jobs(upload_files, submission, existing_submission=False):
             jobs_required.append(validation_job_id)
         upload_dict[upload_file.file_type] = upload_job_id
 
-    # once single-file upload/validation jobs are created, create the cross-file
-    # validation job and dependencies
-    # todo: remove external validation jobs from the code-base--they aren't used
-    if existing_submission and not submission.d2_submission:
-        # find cross-file and external validation jobs and mark them as waiting
-        # (note: job_type of 'validation' is a cross-file job)
-        val_job = sess.query(Job).\
```
alburde1: This is still important, `validation` is cross-file

dataactcore/scripts/setupJobTrackerDB.py
```diff
@@ -10,6 +12,7 @@ def setup_job_tracker_db():
     with create_app().app_context():
         sess = GlobalDB.db().session
         insert_codes(sess)
+        delete_unused_job_types(sess)
```
alburde1: This is going to happen every time? Seems excessive

PR #1202, opened 2018-04-26
CFDA Loader Update from GSA S3 Bucket

dpb-bah:
[Story Link](https://federal-spending-transparency.atlassian.net/browse/DEV-823)

- `load_cfda` now downloads from GSA's S3 bucket
- Note: GSA's bucket is not publicly accessible so we have to guess the filename. To get by, we first check to see if the weekly file has been uploaded and if that's not found, it looks for the past four weeks.

### DIFF
--- a/dataactvalidator/scripts/load_cfda.py
+++ b/dataactvalidator/scripts/load_cfda.py
@@ -14,7 +14,7 @@ logger = logging.getLogger(__name__)
 
 CFDA_FILE_FORMAT = os.path.join('Assistance Listings', 'usaspendinggov', '%Y', '%m-%b',
                                 'AssistanceListings_USASpendingGov_PUBLIC_WEEKLY_%Y%m%d.csv')
-WEEKDAY_UPLOADED = 5  # Usually uploaded on Saturdays
+WEEKDAY_UPLOADED = 5  # datetime.weekday()'s integer representing the day it's usually uploaded (Saturday)
 DAYS_TO_SEARCH = 4 * 7  # 4 weeks
 LOCAL_CFDA_FILE = os.path.join('dataactvalidator', 'config', 'cfda_program.csv')
 LOCAL_CFDA_FILE2 = os.path.join('dataactvalidator', 'config', 'cfda_program2.csv')
--- a/dataactvalidator/scripts/load_cfda.py
+++ b/dataactvalidator/scripts/load_cfda.py
@@ -12,8 +12,7 @@ from botocore.exceptions import ClientError
 s3 = boto3.client('s3', )
 logger = logging.getLogger(__name__)
 
-CFDA_FILE_FORMAT = os.path.join('Assistance Listings', 'usaspendinggov', '%Y', '%m-%b',
-                                'AssistanceListings_USASpendingGov_PUBLIC_WEEKLY_%Y%m%d.csv')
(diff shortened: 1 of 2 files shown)


alburde1:
[run tests]

dataactvalidator/scripts/load_cfda.py
```diff
@@ -1,54 +1,89 @@
...
+
+s3 = boto3.client('s3', )
+logger = logging.getLogger(__name__)
+
+CFDA_FILE_FORMAT = os.path.join('Assistance Listings', 'usaspendinggov', '%Y', '%m-%b',
+                                'AssistanceListings_USASpendingGov_PUBLIC_WEEKLY_%Y%m%d.csv')
+WEEKDAY_UPLOADED = 5  # datetime.weekday()'s integer representing the day it's usually uploaded (Saturday)
+DAYS_TO_SEARCH = 4 * 7  # 4 weeks
+LOCAL_CFDA_FILE = os.path.join('dataactvalidator', 'config', 'cfda_program.csv')
+LOCAL_CFDA_FILE2 = os.path.join('dataactvalidator', 'config', 'cfda_program2.csv')
```
nmonga91: This doesn't seem to be used anywhere in the script, is it necessary to have here?

dataactvalidator/scripts/load_cfda.py
```diff
@@ -1,54 +1,89 @@
...
-
-    # output the directory contents
-    ftp.dir('-t', data.append)
-
-    # get the most recent updated cfda file
-    file_listing = data[0]
+    gsa_connection = boto3.resource('s3', region_name='us-east-1')
+    # disregard aws credentials for public file
+    gsa_connection.meta.client.meta.events.register('choose-signer.s3.*', disable_signing)
+    gsa_bucket = gsa_connection.Bucket('falextracts')
```
nmonga91: Should this live in the settings file? we may want to consider external settings/config files similar to broker so it won't require a code change/deploy to update external references.

dataactvalidator/scripts/load_cfda.py
```diff
@@ -1,54 +1,89 @@
...
+import logging
 
 import boto3
+from botocore.handlers import disable_signing
+from botocore.exceptions import ClientError
+
+s3 = boto3.client('s3', )
+logger = logging.getLogger(__name__)
+
+CFDA_FILE_FORMAT = os.path.join('Assistance Listings', 'usaspendinggov', '%Y', '%m-%b',
```
nmonga91: Same question as below for the `gsa_bucket`.

dataactvalidator/scripts/load_cfda.py
dpb-bah: Whoops, that's an artifact from testing. Removing.

dataactvalidator/scripts/load_cfda.py
dpb-bah: Makes sense. I can work with Ops to figure out the best way of adding these setting.

dataactvalidator/scripts/load_cfda.py
dpb-bah: Makes sense. I can work with Ops to figure out the best way of adding these setting.

PR #1221, opened 2018-05-04
Update dev 885 census districts

shinson:
Adding ability to include congressional districts from Census 2000.

- Added loading  of census congressional districts to initializing ZipCode data
- Updated rule 43.3 and added 43.4 to update PPoP congressional district validation
- Updated rule 44.3 to conform more to 43.4 and include districts from Census 2000

[Bug Link](https://federal-spending-transparency.atlassian.net/browse/DEV-885)

- [x] Backend Code Review
- [x] Work with devOps to add file to s3

Technical Release Notes:
- SQL Rule Validations Update
- Config CSV updates
- Alembic Migrations
- Update to initialize.py


dataactvalidator/config/sqlrules/fabs43_detached_award_financial_assistance_3.sql
```diff
@@ -1,38 +1,10 @@
...
-                )
-            )
-            ELSE FALSE
-        END;
+    row_number,
+    place_of_performance_congr,
+    record_type
+FROM detached_award_financial_assistance
+WHERE submission_id = {0}
+    AND COALESCE(place_of_performance_congr) <> ''
```
alburde1: I think you lost a piece here with the coalesce (should be `COALESCE(place_of_performance_congr, '')`

dataactvalidator/config/sqlrules/fabs43_detached_award_financial_assistance_4.sql
```diff
@@ -0,0 +1,38 @@
...
+FROM detached_award_financial_assistance_fabs43_4_{0} AS dafa
+WHERE CASE WHEN COALESCE(dafa.place_of_performance_congr, '') <> ''
+            THEN ((dafa.place_of_performance_congr <> '90'
+                    AND dafa.row_number NOT IN (
+                        SELECT DISTINCT sub_dafa.row_number
+                        FROM detached_award_financial_assistance_fabs43_4_{0} AS sub_dafa
+                        JOIN state_congressional AS sc_1
+                            ON UPPER(LEFT(sub_dafa.place_of_performance_code, 2)) = sc_1.state_code
+                            AND sub_dafa.place_of_performance_congr = sc_1.congressional_district_no
+                            AND (sc_1.census_year IS NULL
```
alburde1: We can probably do a coalesce here as well instead of a null check `COALESCE(sc_1.census_year, 2000) >= 2000`. I believe that's faster

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
```diff
@@ -8,9 +10,10 @@ FROM detached_award_financial_assistance AS dafa
...
-        WHERE z.zip5 = dafa.legal_entity_zip5
-            AND z.congressional_district_no = dafa.legal_entity_congressional
-    );
+    AND dafa.legal_entity_congressional
+        NOT IN (SELECT sc.congressional_district_no
+    	        FROM state_congressional AS sc,
+    	 	        zips as z
+    	        WHERE z.zip5 = dafa.legal_entity_zip5
+    	            AND z.state_abbreviation = sc.state_code
+    	            AND sc.census_year IS NULL OR sc.census_year >= 2000);
```
alburde1: Same as comment above in terms of COALESCE

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
```diff
@@ -8,9 +10,10 @@ FROM detached_award_financial_assistance AS dafa
...
     AND COALESCE(legal_entity_zip5, '') <> ''
     AND COALESCE(legal_entity_congressional, '') <> ''
-    AND NOT EXISTS (
-        SELECT 1
-        FROM zips AS z
-        WHERE z.zip5 = dafa.legal_entity_zip5
-            AND z.congressional_district_no = dafa.legal_entity_congressional
-    );
+    AND dafa.legal_entity_congressional
+        NOT IN (SELECT sc.congressional_district_no
```
alburde1: We don't want to do a `NOT IN` because that can REALLY slow things down when there are a lot of results in this SELECT (which may happen quite often). We want to stick to the NOT EXISTS method of doing this. It's not noticeable in a small submission or locally, but with a large submission and full zips table it can really take a performance hit

dataactvalidator/scripts/readZips.py
```diff
@@ -247,7 +271,10 @@ def read_zips():
             citystate_file = os.path.join(CONFIG_BROKER["path"], "dataactvalidator", "config", "ctystate.txt")
             parse_citystate_file(open(citystate_file), sess)
 
+            census_file = os.path.join(base_path, "census_congressional_districts.csv")
+
         update_state_congr_table(sess)
+        update_state_congr_table_census(census_file, sess)
```
alburde1: Does a second function make sense here or does it make more sense to make both of them part of the `update_state_congr_table` since that function name seems to encompass what is happening here as well.

tests/unit/dataactvalidator/test_fabs44_detached_award_financial_assistance_3.py
```diff
@@ -13,25 +13,35 @@ def test_column_headers(database):
...
+    zip1 = ZipsFactory(zip5="12345", congressional_district_no="01", state_abbreviation="AB")
+    zip2 = ZipsFactory(zip5="23456", congressional_district_no="01", state_abbreviation="CD")
+    sc1 = StateCongressionalFactory(state_code="AB", congressional_district_no="01", census_year=None)
+    sc2 = StateCongressionalFactory(state_code="CD", congressional_district_no="02", census_year=2000)
+
     det_award_1 = DetachedAwardFinancialAssistanceFactory(legal_entity_zip5="12345", legal_entity_congressional="")
     det_award_2 = DetachedAwardFinancialAssistanceFactory(legal_entity_zip5="", legal_entity_congressional="01")
     det_award_3 = DetachedAwardFinancialAssistanceFactory(legal_entity_zip5="", legal_entity_congressional="")
     det_award_4 = DetachedAwardFinancialAssistanceFactory(legal_entity_zip5="12345", legal_entity_congressional=None)
+
```
alburde1: Why is there a random newline here?

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
shinson: Updating to use NOT EXISTS.

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
shinson: Updating to coalesce to 2010 (since should be the default)

dataactvalidator/scripts/readZips.py
shinson: I kept it separate so that it would not get lost in `update_state_congr_table`, since the census info comes from a file and the while `update_state_congr_table` is based on querying the Zips table. Was your idea that the all the code

tests/unit/dataactvalidator/test_fabs44_detached_award_financial_assistance_3.py
shinson: fixed.

dataactvalidator/scripts/readZips.py
alburde1: Maybe we should rename the first function in that case to make it less generic. Something like `update_state_congr_table_current`? Since right now the name makes it seem like all the updates should happen inside it.

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
```diff
@@ -10,10 +10,11 @@ FROM detached_award_financial_assistance AS dafa
...
-        NOT IN (SELECT sc.congressional_district_no
-    	        FROM state_congressional AS sc,
-    	 	        zips as z
-    	        WHERE z.zip5 = dafa.legal_entity_zip5
-    	            AND z.state_abbreviation = sc.state_code
-    	            AND sc.census_year IS NULL OR sc.census_year >= 2000);
+    AND NOT EXISTS (
+        SELECT 1
+        FROM state_congressional AS sc,
+            zips as z
```
alburde1: I missed this last time, which was my bad. I think we should explicitly use `join` here since that's how all our other SQL is written and we want to stay standardized. We don't usually do implicit joins. Also, all operators should be capitalized per our SQL standards, so the `as` should be caps.

dataactvalidator/config/sqlrules/fabs44_detached_award_financial_assistance_3.sql
```diff
@@ -12,9 +12,9 @@ WHERE submission_id = {0}
...
-        FROM state_congressional AS sc,
-            zips as z
-        WHERE z.zip5 = dafa.legal_entity_zip5
-            AND z.state_abbreviation = sc.state_code
-            AND sc.congressional_district_no = dafa.legal_entity_congressional
-            AND COALESCE(sc.census_year, 2010) >= 2000);
+        FROM state_congressional AS sc
+            INNER JOIN zips AS z
+            ON z.state_abbreviation = sc.state_code
+                WHERE z.zip5 = dafa.legal_entity_zip5
```
alburde1: Indentation is off. The `WHERE` is in relation to the `FROM` so it should be in line with it (adjust lower lines to match this)

PR #1223, opened 2018-05-08
Extract cfda from domain dataload

sharb:
tested with http://ec2-96-127-83-21.us-gov-west-1.compute.amazonaws.com:8080/view/Dashboard%20-%20Data%20Pipelines/job/Broker-CFDA-Load/

### DIFF
--- a/dataactcore/scripts/initialize.py
+++ b/dataactcore/scripts/initialize.py
@@ -17,7 +17,7 @@ from dataactcore.scripts.setupAllDB import setup_all_db
 from dataactvalidator.health_check import create_app
 from dataactvalidator.filestreaming.schemaLoader import SchemaLoader
 from dataactvalidator.filestreaming.sqlLoader import SQLLoader
-from dataactvalidator.scripts.loadFile import load_domain_values
+from dataactvalidator.scripts.loadFile import load_domain_values, load_cfda_program
 from dataactvalidator.scripts.load_sf133 import load_all_sf133
 from dataactvalidator.scripts.loadTas import load_tas
 from dataactvalidator.scripts.loadLocationData import load_location_data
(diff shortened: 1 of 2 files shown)


dataactcore/scripts/initialize.py
```diff
@@ -120,6 +123,7 @@ def main():
     parser.add_argument('-c', '--load_agencies', help='Update agency data (CGACs, FRECs, SubTierAgencies)',
                         action='store_true')
     parser.add_argument('-t', '--update_tas', help='Update broker TAS list', action='store_true')
+    parser.add_argument('--cfda_load', help='Pull Catalog of Federal Domestic Assistance', action='store_true')
```
alburde1: where is the short version of the argument? Everything else has one. Perhaps `-cfda` is a valid flag since we already have a `-c`

dataactvalidator/scripts/loadFile.py
```diff
@@ -99,7 +99,16 @@ def load_country_codes(filename):
     logger.info('{} records inserted to {}'.format(num, table_name))
 
 
-def load_cfda_program(filename):
+def load_cfda_program(base_path):
```
alburde1: If you're going to be removing this from the entire loadFile process, it should not be left in this file. The more important question is: "what is the point of splitting it out?"

sharb: We want to display our clients a clear view of data loads, so we needed to separate out the functions into individual data loads that can be called separately by initialize.py. @kwhickey

dataactcore/scripts/initialize.py
```diff
@@ -153,6 +157,9 @@ def main():
     if args.update_domain:
         load_domain_value_files(validator_config_path)
 
+    if args.cfda_load:
```
alburde1: If we're pulling cfda out of `load_domain_value_files`, there needs to be a way for devs to continue loading all the domain values at once, even if it's an argument `-ad` that calls `load_domain_value_files` and then `load_cfda` for now. `-d` should eventually be that thing but if we want to separate for now that would be the best thing

PR #1235, opened 2018-05-17
Pytest postgres94 fix

sharb:
Fix testing error with postgres 9.4.7 also works with 9.5.x

### DIFF
--- a/tests/unit/dataactbroker/test_file_generation_handler.py
+++ b/tests/unit/dataactbroker/test_file_generation_handler.py
@@ -62,14 +62,6 @@ def test_generate_d1_file_query(monkeypatch, mock_broker_config_paths, database,
             expected1.append(str(dap_one.__dict__[value]))
             expected2.append(str(dap_two.__dict__[value]))
 
-    print(file_rows[1])
-    print("===============================================================")
-    print(expected1)
-    print("===============================================================")
-    print(file_rows[2])
-    print("===============================================================")
-    print(expected1)
-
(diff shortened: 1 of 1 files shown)


tests/unit/dataactbroker/test_file_generation_handler.py
```diff
@@ -62,8 +62,8 @@ def test_generate_d1_file_query(monkeypatch, mock_broker_config_paths, database,
             expected1.append(str(dap_one.__dict__[value]))
             expected2.append(str(dap_two.__dict__[value]))
 
-    assert file_rows[1] == expected1
-    assert file_rows[2] == expected2
+    assert file_rows[1] == expected1 or file_rows[2] == expected1
```
alburde1: Does using `in` not work? We discussed how that might be a better solution

tests/unit/dataactbroker/test_file_generation_handler.py
```diff
@@ -62,8 +62,7 @@ def test_generate_d1_file_query(monkeypatch, mock_broker_config_paths, database,
             expected1.append(str(dap_one.__dict__[value]))
             expected2.append(str(dap_two.__dict__[value]))
 
-    assert file_rows[1] == expected1
-    assert file_rows[2] == expected2
+    assert expected1 in file_rows and expected2 in file_rows
```
alburde1: if we're doing an and, this should remain 2 assert statements so we can see where it breaks (if it does) more clearly

PR #1236, opened 2018-05-17
Local docker setup

sharb:
spins up local postgres 9.4.7 container and broker backend python 3.5 container. Will do alembic upgrade upon broker container initialization (see docker documentation in contribution.mb).


dataactcore/config_example.yml
```diff
@@ -6,24 +6,34 @@
...
 
     # Specify the url where the front end of the application will be accessed.
-    # For a local installation this will most likely be localhost or the 
-    # location where the /public files are located. If a port is required, 
+    # For a local installation this will most likely be localhost or the
+    # location where the /public files are located. If a port is required,
     # it should be appended to the URL.
-    full_url: http://dataactbroker.domain.com
+    full_url: http://localhost:3000
+    # full_url: https://alpha-broker-sandbox.usaspending.gov/
```
alburde1: This doesn't have to be here, this is from my local config file because I don't feel like remembering it all the time.

dataactcore/config_example.yml
```diff
@@ -6,24 +6,34 @@
...
+
+    sqs_queue_name: sqa-quene-name
```
alburde1: This set of misspellings will haunt me forever if I don't comment on them

dataactcore/config_example.yml
```diff
@@ -33,63 +43,112 @@ broker:
     # File D1 API
     # Uses "{}" because it requires string formatting to insert query parameters
     # If the URL is provided, the file name provided will be used when uploading to S3
-    award_procurement_url: https://data-usas.uat.workplace.gov/api/contracts?CGAC={}&date_signed_beginning={}&date_signed_end={}&callback={}
+    award_procurement_url: https://award_procurement_url.gov
```
alburde1: There is a large set of things we don't use anymore, I think if we're already fixing the example config, we need to yank them out. Anything to do with smartronix is probably one of those things (but give it a look before tearing it all out). Some things we no longer use have already been torn out further in the file.

It sounds like that's out of scope for this but to me that's exactly the scope. Part of the decision was to make it so you could just drag and drop the contents here to your own config, I think that involves cleaning out unneeded content but @michaeldhess or @dpb-bah can correct me if I'm wrong on that.

dataactcore/config_example.yml
```diff
@@ -33,63 +43,112 @@ broker:
...
-    server_debug: false #deprecated
-    debug: false
+    server_debug: true
+    debug: true
 
     # URL/IP address that hosts the broker API
-    broker_api_host: http://dataactbroker-api.domain.com
-    broker_api_port: 443
+    # broker_api_host: alpha-broker-sandbox-api.usaspending.gov
+    # broker_api_port: 443
```
alburde1: Another case where we should remove the commented-out lines

dataactcore/config_example.yml
```diff
@@ -33,63 +43,112 @@ broker:
...
+    broker_api_host: 127.0.0.1
+    broker_api_port: 9999
 
     # URL/IP address that hosts the validator API
-    validator_host: http://dataactbroker-validator-api.domain.com
-    validator_port: 80
+    # validator_host: 0.0.0.0
```
alburde1: Here as well, remove comment

dataactcore/config_example.yml
```diff
@@ -33,63 +43,112 @@ broker:
...
+    validator_host: 127.0.0.1
+    validator_port: 8889
 
     # If you would like to restrict access from other origins, set the
     # allowed origins here. Otherwise, leave as '*'.
+    # cross_origin_url: ['https://alpha-broker-sandbox.usaspending.gov','http://localhost', 'http://127.0.0.1', 'http://sandbox-data-act-web-app.s3-website-us-gov-west-1.amazonaws.com']
```
alburde1: Once more, with feeling, remove comment

dataactcore/config_example.yml
```diff
@@ -33,63 +43,112 @@ broker:
...
+    # Connection info for fetching data related to File F
+    fsrs:
+        procurement_service:
+            wsdl: 'https://www.fsrs.gov/ws/ffata_data_api.php?wsdl'
+            username: 'ffataexport'
+            password: 'password'
+        grant_service:
+            wsdl: 'https://www.fsrs.gov/ws/ffata_grants_data_api.php?wsdl'
+            username: 'ffataexport'
+            password: 'password'
```
alburde1: For this section above here, do we want to keep the real data? @michaeldhess ?

dataactcore/config_example.yml
```diff
@@ -98,86 +157,82 @@ services:
...
+    # Logstash EC2 host. Ignored if use_logstash is false.
+    # NOTE: the dummy value below MUST be changed to the correct
+    # value if use_logstash is true.
+    logstash_host: logstash-host.compute-1.amazonaws.com
+
+    # Logstash EC2 port. 514 is the port typically used with logstash,
+    # so you can mostly likely keep this as is.
+    logstash_port: 514
+
+job-queue:
```
alburde1: I have a strange feeling we don't use this at all anymore (we don't have to run rabbit personally anymore, that was something we had to do before). Double-check, see if everything runs without it, if so gut it. Otherwise, I'd say at least remove the real URL and uncomment the port one.

dataactcore/local_secrets_example.yml
```diff
@@ -23,34 +23,24 @@ broker:
...
-    # Only required if planning to load CFDA file
-    cfda_file_path: sample-cfda-file-path
+    # S3 filenames for SF-133 file, only required if planning to load SF-133 table
+    sf_133_bucket: s3-bucket
 
     # Name of the AWS role you're using to upload broker files. Ignored if
     # use_aws is false. NOTE: the dummy value below MUST be changed to the
     # correct value if use_aws is true.
-    aws_role: arn:aws:iam::123456789012:role/roleName
+    aws_role: redacted
```
alburde1: aws_role is both here and in the sample config. Here it's redacted, there it isn't. We should pick one and stick to it.

doc/CONTRIBUTING.md
```diff
@@ -337,6 +337,34 @@ for more configuration details. Everything within `python_config` is imported
 via `dictConfig` (in addition to some standard settings defined in
 `dataactcore.logging`.
 
+### Setup with Docker
+
+Install docker in your local machine by selecting your OS and hitting install from this [link](https://docs.docker.com/docker-for-mac/install).
```
alburde1: Does docker still only work on Mac or is this a way people with Windows can also join us? If it's the latter, do we want to link differently? If it's the former, ignore this comment.

doc/CONTRIBUTING.md
```diff
@@ -337,6 +337,34 @@ for more configuration details. Everything within `python_config` is imported
...
+These are the files you need to add/change before you begin working in your container. `_example` or `example_` needs to be removed from these file names.
+```
+dataactcore/config_example.yml
+dataactcore/local_config_example.yml
+dataactcore/local_secrets_example.yml
+dataactvalidator/config/example_agency_codes_list.csv
+dataactvalidator/config/example_cars_tas.csv
+dataactvalidator/config/example_cgac.csv
+dataactvalidator/config/example_object_class.csv
+dataactvalidator/config/example_program_activity.csv
```
alburde1: All of these files could probably go in the general install section because that's actually something that has to be done no matter what method of installation you're using.

doc/CONTRIBUTING.md
```diff
@@ -337,6 +337,34 @@ for more configuration details. Everything within `python_config` is imported
 via `dictConfig` (in addition to some standard settings defined in
 `dataactcore.logging`.
 
+### Setup with Docker
+
+Install docker in your local machine by selecting your OS and hitting install from this [link](https://docs.docker.com/docker-for-mac/install).
+This [link](https://docs.docker.com/docker-for-mac/install/) will already include docker-compose.
+
+These are the files you need to add/change before you begin working in your container. `_example` or `example_` needs to be removed from these file names.
```
alburde1: We probably want to say you want to make copies and remove the text there since we don't want them removing the example files entirely if they're ever going to make a commit (or if this is a dev that's joining us)

doc/CONTRIBUTING.md
```diff
@@ -337,6 +337,34 @@ for more configuration details. Everything within `python_config` is imported
 via `dictConfig` (in addition to some standard settings defined in
 `dataactcore.logging`.
 
+### Setup with Docker
```
alburde1: I don't see any instructions in here on how to set up the docker without spinning up a new DB, just pointing to an existing one.

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
+version: '3'
+services:
+  dataact-postgres:
+    container_name: dataact-postgres
+    image: postgres:9.4.7
```
alburde1: I'm going to go back to our discussion and loop @michaeldhess and @dpb-bah for their opinions on this. I still think we should go straight to postgres 9.5 on this one. I realize we're running 9.4 on dev/staging/prod now but an upgrade is in the plans and we're going to have to update and re-test this as soon as that happens anyway. All the devs already code against 9.5 so I see no reason to downgrade at this point.

dataactcore/config_example.yml
michaeldhess: Hmmm, it's difficult to say. It does sound out of scope, but I can't think of any other ticket that we've had since I got here that made this many changes to the config files. Ideally things get removed as they become irrelevant, but seeing as these changes were made months ago I'd say a good Spring cleaning is in order.

dataactcore/config_example.yml
michaeldhess: I'm thinking no. It looks like the real pw isn't in there, but I'd remove the real username as well (although it's already exposed in the Git history).

docker-compose.yml
michaeldhess: I'm starting to think otherwise. Unless there's a defined story in an upcoming Sprint to upgrade the Broker DBs, we don't know when it'll actually happen. We've been saying "soon" for months now, so it feels safer to downgrade the local envs until it's something that is actually planned.

doc/CONTRIBUTING.md
sharb: If using docker, this should work in windows also. I updated the link to the generic install link.

dataactcore/config_example.yml
alburde1: I don't think this was ever resolved.

doc/CONTRIBUTING.md
```diff
@@ -176,13 +176,32 @@ Before running the broker, you'll need to provide a few configuration options. T
...
+```
+#!/bin/bash
 
-1. Open `config_example.yml` in a text editor and save it as `config.yml`.
-2. Update the values in `config.yml` as appropriate for your installation and save. In many cases the default values will work just fine. The most important config values to change will be in `local_config.yml` and `local_secrets.yml`
-3. Open `local_config_example.yml` in a text editor and save it as `local_config.yml`.
-4. Update the values in `local_config.yml` as appropriate for your installation and save.
-5. Open `local_secrets_example.yml` in a text editor and save it as `local_secrets.yml`.
-6. Update the values in `local_secrets.yml` as appropriate for your installation and save.
+mv dataactcore/config_example.yml dataactcore/config.yml
```
alburde1: All of these mvs are bad. We discussed how the example files should not be altered or deleted in any way by the user unless they are actively trying to update something about configs. They should be cps

doc/CONTRIBUTING.md
```diff
@@ -337,6 +356,49 @@ for more configuration details. Everything within `python_config` is imported
...
+
+```
+#!/bin/bash
+
+mv dataactcore/config_example.yml dataactcore/config.yml
+mv dataactcore/local_config_example.yml dataactcore/local_config.yml
+mv dataactcore/local_secrets_example.yml dataactcore/local_secrets.yml
+
+
+mv dataactvalidator/config/example_agency_codes_list.csv dataactvalidator/config/agency_codes_list.csv
```
alburde1: Same comment about mvs

doc/CONTRIBUTING.md
alburde1: Bumping this comment

doc/CONTRIBUTING.md
alburde1: If these files are already in the general section, we don't also need them here. Just refer to the earlier list.

doc/CONTRIBUTING.md
```diff
@@ -337,6 +356,49 @@ for more configuration details. Everything within `python_config` is imported
...
+`docker-compose up -d`  This command will spin up the postgres container `dataact-postgres` and build your backend container `dataact-broker`. This will take much longer the first time because it's building the image and installing the requirements.
+
+Wait about 20 seconds for everything to come up (first time setup can take up to 8 minutes) then login/ssh to the backend container with this command:
+`docker exec -it dataact-broker /bin/bash`
+This will take you to the workspace directory within the dataact-backend container that will have your backend repository mounted so changes in that repository will also be changed within the container. This means developers can change the files locally and it will reflect what if they want to run them within the container.
+
+If you want to use postgres on your local machine, change the config to point to your host IP.
+
+At this point you are then ready to do `python dataactcore/initialize.py -i` and run your app.
+
```
shinson: When in the container, I run `python datactbroker/app.py`, it logs the app is running on `http://127.0.0.1:9999`. I am unable to get a response when make GET requests to endpoints like `http://127.0.0.1:9999/v1/current_user/`, it cannot find it. Not sure if  exposing the ports for the data broker api app (9999) and the data validator app (8889) is what is needed, or additional commands to run the apps through the container.  It would be good to add it to the documentation some how.

doc/CONTRIBUTING.md
sharb: I couldn't get my local postgres to work in the first place so I can't test how this would work. You should be able to change the config to point to your IP address but again I haven't been able to test it @alburde1 I have this line in the documentation if anyone wants to try it out "If you want to use postgres on your local machine, change the config to point to your host IP."

doc/CONTRIBUTING.md
sharb: 9999 is mapped to port 9000 to the local machine. So using 9000 instead of 9999 should work. Although I haven't tested this yet.

dataactcore/config.py
```diff
@@ -108,7 +108,7 @@
     'http://|:(.*)', '', CONFIG_SERVICES['validator_host'])
 # if hosts in config file are set to 0.0.0.0, override to
 # 127.0.0.1 for cross-platform compatibility
-if CONFIG_SERVICES['broker_api_host'] == '0.0.0.0':
+if env is not "local" and CONFIG_SERVICES['broker_api_host'] == '0.0.0.0':
```
sharb: This check is needed so that config.py will only change the IP of broker to 127.0.0.1 (which doesn't work within docker containers) if it's not running locally and when it's set to '0.0.0.0' .  The main purpose of this is not to change the IP (that is set to 0.0.0.0 in configs) when working locally so that docker can correctly expose ports.

doc/CONTRIBUTING.md
sharb: I did some research and added more documentation of using host postgres with containers

doc/CONTRIBUTING.md
```diff
@@ -337,6 +356,36 @@ for more configuration details. Everything within `python_config` is imported
...
 
+### Setup with Docker
+
+Install docker in your local machine by selecting your OS and hitting install from this [link](https://docs.docker.com/install/) (this installation includes `docker-compose` as well).
+
+Next step is to refer to the `Create Broker Config Files` section of this documentation to copy and rename config files, if you choose to use the default configs.
+
+After you successfully installed Docker, make sure the docker daemon is running on your local machine by running `docker version` and make sure you have your configs set up. Run the following command in the root level of this backend repository:
+
+- `docker-compose up -d`  This command will spin up the postgres container `dataact-postgres`, build your backend image `broker-backend` and run the two service `dataact-broker` and `dataact-validator`. This will take longer the first time because it's building the image and installs the requirements.
```
shinson: I would also add that not running it in the background `-d` will out put the logs from the broker and validator app.

doc/CONTRIBUTING.md
shinson: I was able to connect to my local database using docker.for.mac.host.internal since I'm running on docker 17.12.0.

dataactcore/config_example.yml
```diff
@@ -33,60 +42,91 @@ broker:
...
 
 services:
     # Set to true to turn on tracing rest errors.
-    rest_trace: false #deprecated
+    rest_trace: true
 
     # Set to true to turn on server debugging.
-    server_debug: false #deprecated
-    debug: false
+    server_debug: true
```
alburde1: Are these no longer deprecated?

dataactcore/config_example.yml
```diff
@@ -98,86 +138,82 @@ services:
...
-    # We import award/subaward records from FSRS
-    fsrs:
-        procurement_service:
-            wsdl: ''  # e.g. https://example.com/?wsdl'
-            username: ''
-            password: ''
-        grant_service:
-            wsdl: ''  # e.g. https://example.com/?wsdl'
-            username: ''
-            password: ''
```
alburde1: Where did the fsrs config go? We definitely use it in scripts, why is it gone?

dataactcore/config_example.yml
```diff
@@ -98,86 +138,82 @@ services:
...
+    # Logstash EC2 host. Ignored if use_logstash is false.
+    # NOTE: the dummy value below MUST be changed to the correct
+    # value if use_logstash is true.
+    logstash_host: logstash-host.compute-1.amazonaws.com
+
+    # Logstash EC2 port. 514 is the port typically used with logstash,
+    # so you can mostly likely keep this as is.
+    logstash_port: port
+
+job-queue:
```
alburde1: I still think we need to look into whether this is still required or not. I'm pretty sure we got rid of RabbitMQ entirely

dataactcore/local_secrets_example.yml
```diff
@@ -23,34 +23,24 @@ broker:
...
-    aws_bucket: sample-aws-bucket-name
-    
-    # S3 filenames for SF-133 file, only required if planning to load SF-133 table
-    sf_133_bucket: sample-sf133-bucket-name
-
-    # Region for pulling CFDA file, only required if planning to load CFDA file
-    cfda_region: us-east-1
-
-    # Bucket for pulling CFDA file, only required if planning to load CFDA file
-    cfda_bucket: sample-cfda-bucket-name
```
alburde1: Why were these removed? These are used

docker-compose.yml
alburde1: Looping back around to this, it sounds like we're upgrading to postgres 10 this sprint, do we want this to reflect that so we don't have to go back?

dataactcore/config_example.yml
sharb: I am not sure, should I remove them?

dataactcore/config_example.yml
sharb: I will add this back in

dataactcore/local_secrets_example.yml
sharb: Will add those back in

dataactcore/config_example.yml
sharb: Should I remove the whole "job-queue" block? Or just the rabbitMQ username and password?

docker-compose.yml
sharb: I will leave this version until the upgrade process begins, you can always change this value locally if you need to test it with different versions.

alburde1: Either this needs to be updated here before this is merged, or updating this file has to be a subtask on whatever ticket upgrades the DB versions. This can't be left at 9.4 for local dev when our production DBs are 10. This will also involve testing to make sure it doesn't break anything with the Docker.

sharb: the postgres upgrade hasn't begun yet so our production DBs are still 9.4.7, but i'll defer to @willkjackson if this should be postgres 10 or create subtask when the postgres upgrades begins.

willkjackson: This is subtask on the operations ticket to upgrade DB versions. We can drop this part of the review.

dataactcore/local_secrets_example.yml
sharb: realized this cfda_region and cfda_bucket was already in example_config.

PR #1248, opened 2018-06-04
Dev 1075 submission metadata endpoint

alburde1:
Create a new endpoint to retrieve submission metadata

- [x] Reviewed by backend
- [ ] Merged concurrently with https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/880

### DIFF
--- a/dataactbroker/handlers/submission_handler.py
+++ b/dataactbroker/handlers/submission_handler.py
@@ -124,6 +124,7 @@ def get_submission_metadata(submission):
         'reporting_period': reporting_date(submission),
         'publish_status': submission.publish_status.name,
         'quarterly_submission': submission.is_quarter_format,
+        'fabs_submission': submission.d2_submission,
         'fabs_meta': fabs_meta
     }
 
--- a/dataactbroker/handlers/submission_handler.py
+++ b/dataactbroker/handlers/submission_handler.py
@@ -5,9 +5,11 @@ from flask import g
 from sqlalchemy import func, or_, desc
 
 from dataactcore.interfaces.db import GlobalDB
-from dataactcore.interfaces.function_bag import sum_number_of_errors_for_job_list
+from dataactcore.interfaces.function_bag import (sum_number_of_errors_for_job_list, get_last_validated_date,
+                                                 get_fabs_meta)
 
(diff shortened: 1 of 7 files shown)


alburde1:
[run tests]

alburde1:
[run tests]

tests/unit/dataactbroker/test_submission_handler.py
```diff
@@ -0,0 +1,130 @@
...
+    sess = database.session
+
+    now = datetime.datetime.utcnow()
+    now_plus_10 = now + datetime.timedelta(minutes=10)
+    start_date = datetime.date(2000, 1, 1)
+    cgac = CGACFactory(cgac_code='001', agency_name='CGAC Agency')
+    frec_cgac = CGACFactory(cgac_code='999', agency_name='FREC CGAC')
+    frec = FRECFactory(frec_code='0001', agency_name='FREC Agency', cgac=frec_cgac)
+    # Quarterly DABS cgac corrected
+    sub_1 = SubmissionFactory(submission_id='1', created_at=now, updated_at=now_plus_10, cgac_code=cgac.cgac_code,
```
shinson: Should we separate each type of submission into its own test?

dataactbroker/handlers/submission_handler.py
```diff
@@ -85,6 +87,65 @@ def get_submission_stats(submission_id):
...
+            submission: submission to retrieve metadata for
+
+        Returns:
+            object containing metadata for the submission
+    """
+    sess = GlobalDB.db().session
+
+    # Determine the agency name
+    cgac = sess.query(CGAC).filter_by(cgac_code=submission.cgac_code).one_or_none()
+    frec = sess.query(FREC).filter_by(frec_code=submission.frec_code).one_or_none()
```
shinson: Do we want to call frec when we don't have to? Instead of assigning it to a variable we could call it within the conditional.

tests/unit/dataactbroker/test_submission_handler.py
alburde1: I was partially keeping them in the same test so I could show we were getting the right one, but we can break them up

PR #1253, opened 2018-06-06
extracted country codes from domain

sharb:
created country codes flag and deleted the temporary domain flag.

### DIFF
--- a/dataactvalidator/scripts/load_country_codes.py
+++ b/dataactvalidator/scripts/load_country_codes.py
@@ -14,50 +14,42 @@ from dataactvalidator.scripts.loaderUtils import clean_data, insert_dataframe
 logger = logging.getLogger(__name__)
 
 
-def load_country_codes(filename):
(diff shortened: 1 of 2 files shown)


dataactvalidator/scripts/load_country_codes.py
```diff
@@ -14,50 +14,42 @@
...
-        s3bucket = s3connection.lookup(CONFIG_BROKER['sf_133_bucket'])
-        country_codes_file = s3bucket.get_key("country_codes.csv").generate_url(expires_in=600)
-    else:
-        country_codes_file = os.path.join(base_path, "country_codes.csv")
-
-    logger.info('Loading country codes file: ' + "country_codes.csv")
-    load_country_codes(country_codes_file)
-
-
 if __name__ == '__main__':
```
shinson: Do not need this since configure logging is called in `initialize.py`

PR #1260, opened 2018-06-11
DEV-1067: DABS service account permissions

michaeldhess:
Updated `perms_to_affiliations` to accept service level accounts (`A`) and give them DABS `w` permissions and the new FABS `e` permissions (which are currently unused).

- [ ] Reviewed by backend

**Link to JIRA Ticket:**
[DEV-1067](https://federal-spending-transparency.atlassian.net/browse/DEV-1067)

**Technical Release Notes:**
Row added to the user permissions table (`PermissionType`): `editfabs`

### DIFF
--- a/dataactcore/models/lookups.py
+++ b/dataactcore/models/lookups.py
@@ -87,9 +87,8 @@ PERMISSION_TYPES = [
     LookupType(1, 'reader', 'This user is allowed to view any submission for their agency'),
     LookupType(2, 'writer', 'This user is allowed to create and edit any submission for their agency'),
     LookupType(3, 'submitter', 'This user is allowed to certify and submit any submission for their agency'),
-    # Placeholder 4: website_admin
-    LookupType(5, 'fabs', 'This user is allowed to publish any FABS data for their agency'),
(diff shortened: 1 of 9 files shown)


dataactcore/models/lookups.py
```diff
@@ -88,11 +88,14 @@
     LookupType(2, 'writer', 'This user is allowed to create and edit any submission for their agency'),
     LookupType(3, 'submitter', 'This user is allowed to certify and submit any submission for their agency'),
     # Placeholder 4: website_admin
-    LookupType(5, 'fabs', 'This user is allowed to create and publish any FABS data for their agency')
+    LookupType(5, 'fabs', 'This user is allowed to publish any FABS data for their agency'),
+    LookupType(6, 'editfabs', 'This user is allowed to create and edit any FABS data for their agency')
```
alburde1: I know it says 4 is a placeholder, but website admin is a flag on the user now. Also, we probably want to make the edit a lower number than the "everything" because that's the way we've been doing it in the past. We can just slide `editfabs` into the 4 slot I think.

PR #1263, opened 2018-06-13
DEV-1068: FABS service account permissions

michaeldhess:
Updated all FABS permission checks to also accept `editfabs` permissions, excluding publishing. Updated the permissions checks so when we check for DABS (`reader`<`writer`<`submitter`) permissions we ignore FABS (`reader`<`editfabs`<`fabs`).

- [ ] Reviewed by backend

**Link to JIRA Ticket:**
[DEV-1068](https://federal-spending-transparency.atlassian.net/browse/DEV-1068)

### DIFF
--- a/dataactbroker/file_routes.py
+++ b/dataactbroker/file_routes.py
@@ -26,7 +26,6 @@ def add_file_routes(app, create_credentials, is_local, server_path):
     # Keys for the post route will correspond to the four types of files
     @app.route("/v1/submit_files/", methods=["POST"])
     @requires_login
-    @requires_submission_perms('writer')
     def submit_files():
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
         return file_manager.validate_submit_files(create_credentials)
--- a/dataactbroker/file_routes.py
+++ b/dataactbroker/file_routes.py
@@ -225,7 +225,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
     @requires_submission_perms('writer', check_fabs='editfabs')
     def delete_submission(submission):
         """ Deletes all data associated with the specified submission
-            NOTE: THERE IS NO WAY TO UNDO THIS 
+            NOTE: THERE IS NO WAY TO UNDO THIS
         """
         return delete_all_submission_data(submission)
 
--- a/dataactbroker/file_routes.py
+++ b/dataactbroker/file_routes.py
@@ -11,7 +11,7 @@ from dataactbroker.handlers.submission_handler import (
     get_revalidation_threshold)
 
 from dataactbroker.decorators import convert_to_submission_id
-from dataactbroker.permissions import requires_login, requires_submission_perms
+from dataactbroker.permissions import requires_login, requires_submission_perms, current_user_can_on_submission
 
 from dataactcore.interfaces.function_bag import get_fabs_meta
 
@@ -179,6 +179,7 @@ def add_file_routes(app, create_credentials, is_local, server_path):
         return JsonResponse.create(StatusCode.OK, get_fabs_meta(submission.submission_id))
 
     @app.route("/v1/upload_detached_file/", methods=["POST"])
+    @requires_submission_perms('editfabs')
     def upload_detached_file():
         file_manager = FileHandler(request, is_local=is_local, server_path=server_path)
         return file_manager.upload_detached_file(create_credentials)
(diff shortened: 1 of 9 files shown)


dataactbroker/file_routes.py
```diff
@@ -179,16 +177,17 @@ def get_fabs_metadata(submission):
         return JsonResponse.create(StatusCode.OK, get_fabs_meta(submission.submission_id))
 
     @app.route("/v1/upload_detached_file/", methods=["POST"])
+    @requires_login
```
alburde1: Should this be just `requires login` or should we be checking for at least `editfabs` permissions?

dataactbroker/permissions.py
```diff
@@ -16,7 +17,16 @@
 
 
 def requires_login(func):
-    """Decorator requiring that _a_ user be logged in (i.e. that we're not using an anonymous session)"""
+    """ Decorator requiring that a user be logged in (i.e. that we're not using an anonymous session)
+
+        Args:
```
alburde1: I think you wrote this in the wrong place, none of these is an arg here and the return is incorrect.

dataactbroker/permissions.py
```diff
@@ -39,71 +49,111 @@ def inner(*args, **kwargs):
...
+            Boolean result on whether the user has permissions greater than or equal to perm
+    """
     is_owner = hasattr(g, 'user') and submission.user_id == g.user.user_id
     return (is_owner and check_owner) or current_user_can(perm, submission.cgac_code, submission.frec_code)
 
 
-def requires_submission_perms(perm, check_owner=True):
-    """Decorator that checks the current user's permissions and validates that the submission exists. It expects a
-    submission_id parameter and will return a submission object"""
+def requires_submission_perms(perm, check_owner=True, check_fabs=False):
```
alburde1: If check_fabs is going to be a string in the end, maybe the default should be `''` or `None` just to keep it consistent.

tests/unit/dataactbroker/test_permissions.py
```diff
@@ -4,75 +4,299 @@
...
+    assert not permissions.current_user_can('submitter', None, user_frec.frec_code)
+
+    # right agency, right permission
+    assert permissions.current_user_can('reader', None, user_frec.frec_code)
+    assert permissions.current_user_can('editfabs', None, user_frec.frec_code)
+    assert permissions.current_user_can('fabs', None, user_frec.frec_code)
+
+    # wrong agency, but superuser
+    user_fabs.website_admin = True
+    assert permissions.current_user_can('fabs', None, other_frec.frec_code)
```
alburde1: Putting this here because there isn't anywhere to just put it for the overall file. Can we get a test for someone with multiple permissions? Both dabs and fabs affiliations and make sure one doesn't overwrite the other in any way?

dataactbroker/file_routes.py
michaeldhess: Using `requires_submission_perms` means that a submission already exists to check against, so I added `current_user_can` to `submit_files` and `upload_detached_file`.

dataactbroker/file_routes.py
```diff
@@ -221,10 +222,10 @@ def post_submission_report_url(submission, warning, file_type, cross_type):
 
     @app.route("/v1/delete_submission/", methods=['POST'])
     @convert_to_submission_id
-    @requires_submission_perms('writer')
+    @requires_submission_perms('writer', check_fabs='editfabs')
     def delete_submission(submission):
         """ Deletes all data associated with the specified submission
-        NOTE: THERE IS NO WAY TO UNDO THIS """
+            NOTE: THERE IS NO WAY TO UNDO THIS """
```
alburde1: Just noticed, the `"""` should be on a new line (we're being picky now)

dataactbroker/permissions.py
```diff
@@ -39,71 +55,111 @@ def inner(*args, **kwargs):
...
+            cgac_code: 3-digit numerical string identifying a CGAC agency
+            frec_code: 4-digit numerical string identifying a FREC agency
+
+        Returns:
+            Boolean result on whether the user has permissions greater than or equal to permission
+    """
+    # If the user is not logged in, or the user is a website admin, there is no reason to check their permissions
+    if not hasattr(g, 'user'):
+        return False
+    elif g.user.website_admin:
```
alburde1: should this be an `elif` or just an `if` since we're returning above?

PR #1274, opened 2018-06-24
Feature dev 1078

IsaacRay:
**High level description:**
Redesign of program_activity_loader file to work properly with current version of file from OMB. Includes functioning unit tests.

**Technical Release Notes:**
Program activity loader stores date of last upload, and checks against it before running each time. Migration required to change ExternalDataLoadDate to datetime. Upgrade from boto to boto3. Enforces required fields in file. Threshold of 1% of rows can be ignored due to missing data before build is failed. Build also fails on blank files.

**Link to JIRA Ticket:**
[DEV-1078](https://federal-spending-transparency.atlassian.net/browse/DEV-1078)

**The following are ALL required for the PR to be merged:**
- [X] Unit & integration tests updated with relevant test cases


dataactcore/models/domainModels.py
```diff
@@ -2,7 +2,8 @@
 
 import sqlalchemy as sa
 
-from sqlalchemy import Column, Date, ForeignKey, Index, Integer, Numeric, Text, Float, UniqueConstraint, Boolean
+from sqlalchemy import Column, Date, DateTime, ForeignKey, Index, Integer, Numeric
+from sqlalchemy import Text, Float, UniqueConstraint, Boolean
```
michaeldhess: Why were these split up? Shouldn't they all be in the same line if they're from the same place?

dataactcore/utils/failureThresholdException.py
```diff
@@ -0,0 +1,8 @@
+
```
michaeldhess: Change this filename to use under_scores instead of camelCase.

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -71,6 +86,29 @@ def clean_data(data, model, field_map, field_options):
...
+        cleaned = data.dropna(subset=required_values)
+        dropped = data[np.invert(data.index.isin(cleaned.index))]
+        # log every dropped row
+        for index, row in dropped.iterrows():
+            logger.info("Dropped row due to faulty data: fyq:{}--agency:{}--alloc:{}--account:{}".format(
+                row["fiscal_year_quarter"],
+                row['agency_id'], row['allocation_transfer_id'], row['account_number']) +
+                "--pa_code:{}--pa_name:{}".format(row['program_activity_code'], row['program_activity_name']))
+
+        if (dropped.shape[0]/cleaned.shape[0]) > FAILURE_THRESHOLD_PERCENTAGE:
```
michaeldhess: We want row counts here, correct? I think we should be using `len(dropped.index)` instead of `dropped.shape[0]`. It might be technically doing the same thing, but comparing lengths should be faster than an entire column (which is what I think that's doing).

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -84,8 +122,10 @@ def clean_data(data, model, field_map, field_options):
     # add created_at and updated_at columns
     now = datetime.utcnow()
     data = data.assign(created_at=now, updated_at=now)
-
-    return data
+    if return_dropped_count:
+        return dropped.shape[0], data
```
michaeldhess: Once again I would use `len(dropped.index)` here.

michaeldhess:
[run tests]

michaeldhess:
[run tests]

michaeldhess:
[run tests]

alburde1:
[run tests]

alburde1:
[run tests]

IsaacRay:
[run tests]

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -71,6 +86,29 @@ def clean_data(data, model, field_map, field_options):
...
+        cleaned = data.dropna(subset=required_values)
+        dropped = data[np.invert(data.index.isin(cleaned.index))]
+        # log every dropped row
+        for index, row in dropped.iterrows():
+            logger.info("Dropped row due to faulty data: fyq:{}--agency:{}--alloc:{}--account:{}".format(
+                row["fiscal_year_quarter"],
+                row['agency_id'], row['allocation_transfer_id'], row['account_number']) +
+                "--pa_code:{}--pa_name:{}".format(row['program_activity_code'], row['program_activity_name']))
+
+        if (len(dropped.index)/len(cleaned.index)) > FAILURE_THRESHOLD_PERCENTAGE:
```
alburde1: question, shouldn't this be `len(dropped.index)/(len(data.index))` if you want it to be an actual percentage?

sharb:

tests/unit/dataactvalidator/test_load_program_activity.py
```diff
@@ -1,19 +1,24 @@
...
 
     load_program_activity.get_program_activity_file('some path')
 
-    boto.s3.connect_to_region().lookup().get_key.assert_called_with('program_activity.csv')
+    boto3.resource('s3').Object(
+            load_program_activity.PA_BUCKET, load_program_activity.PA_SUB_KEY+load_program_activity.PA_FILE_NAME
+            ).get.assert_called_with(load_program_activity.PA_SUB_KEY+load_program_activity.PA_FILE_NAME)
 
 
 def test_get_program_activity_file_local(monkeypatch):
```
michaeldhess: This test is failing.

nmonga91: This entire test file is passing locally. Looking into reproducing the failure.

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
+
+
+def get_stored_pa_last_upload():
+    """ Gets last recorded timestamp from last time file was processed.
+
+        Returns:
+            Upload date of most recent file we have recorded (Datetime object)
+    """
+    sess = GlobalDB.db().session
+    last_stored_obj = sess.query(ExternalDataLoadDate).join(ExternalDataType).filter(
```
alburde1: Do we need to do this join? We have a constant in `lookups.py` called `EXTERNAL_DATA_TYPE` that we can use to get the ID, which is what's stored in `ExternalDataLoadDate` so there's really no need to make a join.

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
+
+
+def set_stored_pa_last_upload(load_datetime):
+    """ Set upload date of most recent file we have recorded (Datetime object)
+
+        Args:
+            Datetime object representing the timestamp associated with the current file
+    """
+    sess = GlobalDB.db().session
+    last_stored_obj = sess.query(ExternalDataLoadDate).join(ExternalDataType).filter(
```
alburde1: Same question as above for the join.

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
+        ExternalDataType.name == "program_activity_upload"
+        ).one_or_none()
+    if not last_stored_obj:
+        data_type = sess.query(ExternalDataType).filter(
+            ExternalDataType.name == "program_activity_upload"
+            ).one_or_none()
+        last_stored_obj = ExternalDataLoadDate(external_data_type_id=data_type.external_data_type_id,
```
alburde1: Same as above question, do we need to get that object just to get an ID or can we just use the lookup?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
     """ Load program activity lookup table.
 
         Args:
             base_path: directory of domain config files
     """
+    last_upload = get_date_of_current_pa_upload(base_path)
+    if not (last_upload > get_stored_pa_last_upload()):
+        return
+    else:
+        set_stored_pa_last_upload(last_upload)
```
alburde1: A) Do we need an `else` here since the return would get us out of this function entirely?
B) Do we want to set it now when the process might still fail or do we want to set it at the end when the process is successfully completed?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
-    logger.info('Loading program activity: program_activity.csv')
+    logger.info('Loading program activity: ' + PA_FILE_NAME)
 
     with create_app().app_context():
         sess = GlobalDB.db().session
+        try:
+            data = pd.read_csv(program_activity_file, dtype=str)
+        except pd.io.common.EmptyDataError as e:
+            log_blank_file()
+            sys.exit(4)  # exit code chosen arbitrarily, to indicate distinct failure states. Approved by Ops.
```
alburde1: Do we want to clean this comment up a bit? This looks more like a comment to ops than one that would help developers. Maybe just `exit code to indicate distinct failure states`? We have this in several places below as well, do those need comments too?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
+                {"fyq": "fiscal_year_quarter", "agency_id": "agency_id", "allocation_id": "allocation_transfer_id",
+                 "account": "account_number", "pa_code": "program_activity_code", "pa_name": "program_activity_name"},
+                {"program_activity_code": {"pad_to_length": 4}, "agency_id": {"pad_to_length": 3},
+                 "allocation_transfer_id": {"pad_to_length": 3, "keep_null": True},
+                 "account_number": {"pad_to_length": 4}},
+                ["agency_id", "program_activity_code", "account_number", "program_activity_name"],
+                True
+            )
+        except FailureThresholdExceededException as e:
+            if e.count == 0:
```
alburde1: Can this happen considering the above exception? According to the pandas documentation, the above exception is raised when `Exception that is thrown in pd.read_csv (by both the C and Python engines) when empty data or header is encountered.` so I don't think we'd ever get a count of 0? If it can happen, then does the above exception ever happen?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
+                log_blank_file()
+                sys.exit(4)
+            else:
+                count_str = "Application tried to drop {} rows ".format(e.count)
```
alburde1: Small thing but while we're here, random space at the end and no period?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -20,54 +29,179 @@ def get_program_activity_file(base_path):
...
         # because we're only loading a subset of program activity info,
         # there will be duplicate records in the dataframe. this is ok,
-        # but need to de-duped before the db load.
-        data.drop_duplicates(inplace=True)
+        # but need to de-duped before the db load. We also need to log them.
+
+        deduped = data.drop_duplicates()
+        dropped_dupes = data[np.invert(data.index.isin(deduped.index))]
+        logger.info("Dropped {} duplicate rows.".format(len(dropped_dupes.index)))
+        data = deduped
```
alburde1: For this process, if the only reason we're not dropping inplace is because we want a count, couldn't we just get the original length, then drop dupes and do the original length minus this one? Then we wouldn't have to create two new objects, just a single count. Probably also faster just counting than doing inversions, etc.

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -38,11 +44,11 @@ def trim_item(item):
...
 
-def clean_data(data, model, field_map, field_options):
+def clean_data(data, model, field_map, field_options, required_values=[], return_dropped_count=False):
+
     """ Cleans up a dataframe that contains domain values.
 
-    Parameters:
-    ----------
+    Args:
         data : dataframe of domain values
```
alburde1: There's a space after `data` here. Depending on the editor it might not recognize that as one of the args. Also, we're missing `required_values` and `return_dropped_count` in this list.

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -52,9 +58,18 @@ def clean_data(data, model, field_map, field_options):
             "keep_null" when set to true, empty fields will not be padded
             "skip_duplicate" which ignores subsequent lines that repeat values
             "strip_commas" which removes commas
+    Returns:
```
alburde1: There should be a newline above this to conform to our standards.

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -71,6 +86,29 @@ def clean_data(data, model, field_map, field_options):
...
+        # check the columns that must have a valid value, and if they have white space,
+        # replace with NaN so that dropna finds them.
+        for value in required_values:
+            data[value].replace('', np.nan, inplace=True)
+        # drop any rows that are missing required data
+        cleaned = data.dropna(subset=required_values)
+        dropped = data[np.invert(data.index.isin(cleaned.index))]
+        # log every dropped row
+        for index, row in dropped.iterrows():
+            logger.info("Dropped row due to faulty data: fyq:{}--agency:{}--alloc:{}--account:{}".format(
```
alburde1: Why are we doing 2 formats here? Can't we just do it all in one big one?

dataactvalidator/scripts/loaderUtils.py
```diff
@@ -84,8 +122,10 @@ def clean_data(data, model, field_map, field_options):
     # add created_at and updated_at columns
     now = datetime.utcnow()
     data = data.assign(created_at=now, updated_at=now)
-
-    return data
+    if return_dropped_count:
+        return len(dropped.index), data
+    else:
```
alburde1: Do we need this `else`? It'll never get here if it passes the `if` check

tests/unit/dataactvalidator/test_load_program_activity.py
```diff
@@ -22,17 +27,51 @@ def test_get_program_activity_file_local(monkeypatch):
 
     pa_file = load_program_activity.get_program_activity_file('local_path')
 
-    assert pa_file == 'local_path/program_activity.csv'
+    assert pa_file == 'local_path/DATA Act Program Activity List for Treas.csv'
+
+
+def test_set_get_pa_last_upload_existing(monkeypatch, database):
```
alburde1: This should probably include a test for the epoch before we set the date the first time. Also, there is no comment for this, there should be a comment for this test explaining what it does (we do not have to include the args, but we should at least detail what the test is)

tests/unit/dataactvalidator/test_load_program_activity.py
```diff
@@ -22,17 +27,51 @@ def test_get_program_activity_file_local(monkeypatch):
...
+    sess.commit()
+
+    load_program_activity.set_stored_pa_last_upload(datetime.datetime(2017, 12, 31, 0, 0, 0))
 
+    stored_date = load_program_activity.get_stored_pa_last_upload()
+    expected_date = datetime.datetime(2017, 12, 31, 0, 0, 0)
+    assert stored_date == expected_date
 
+    # repeat this, because the first time, there is no stored object, but now test with one that already exists.
+
```
alburde1: Extra newline (small, but again, while we're here)

tests/unit/dataactvalidator/test_load_program_activity.py
```diff
@@ -22,17 +27,51 @@ def test_get_program_activity_file_local(monkeypatch):
...
+@patch('dataactvalidator.scripts.load_program_activity.get_stored_pa_last_upload')
+@patch('dataactvalidator.scripts.load_program_activity.get_date_of_current_pa_upload')
 @patch('dataactvalidator.scripts.load_program_activity.get_program_activity_file')
-def test_load_program_activity_data(mocked_get_pa_file, database, monkeypatch):
+def test_load_program_activity_data(mocked_get_pa_file,
+                                    mocked_get_current_date,
+                                    mocked_get_stored_date,
+                                    mocked_set_stored_date,
+                                    database,
+                                    monkeypatch):
```
alburde1: Why so many newlines?

tests/unit/dataactvalidator/test_load_program_activity.py
nmonga91: added

dataactvalidator/scripts/load_program_activity.py
nmonga91: we should be setting this at the end, only when it succeeds. Will refactor this.

PR #1276, opened 2018-06-28
Travis Ci Broker

sharb:
Had to change the image to alpine because default postgres image is based off debian which fails with `pytest`. Also added `depends_on` flag to docker-compose, this doesn't change the way local docker environment is setup, it just makes sure postgres spins up with broker.


sharb:

sharb:

PR #1287, opened 2018-07-09
DEV-1195: Fix submit_files permissions check 

michaeldhess:
**High level description:**
Updated the permissions check for the `submit_files` to function similarly to the rest of the checks.

**Technical details:**
Removed the non-functioning `current_user_can` check within the route and the permissions check within the `submit` function. Added a wrapper to check for permissions based on the agency associated with the request's existing Submission or the `cgac_code`/`frec_code` within the request itself.

**Link to JIRA Ticket:**
[DEV-1195](https://federal-spending-transparency.atlassian.net/browse/DEV-1195)

**The following are ALL required for the PR to be merged:**
- Unit & integration tests updated with relevant test cases N/A

### DIFF
--- a/dataactbroker/permissions.py
+++ b/dataactbroker/permissions.py
@@ -1,6 +1,8 @@
 from functools import wraps
-
 from flask import g
+
+from webargs import fields as webargs_fields
+from webargs.flaskparser import parser as webargs_parser
 
 from dataactcore.interfaces.db import GlobalDB
 from dataactcore.models.jobModels import Submission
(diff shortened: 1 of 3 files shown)


dataactbroker/permissions.py
```diff
@@ -146,6 +148,52 @@ def wrapped(submission_id, *args, **kwargs):
...
+        @requires_login
+        @wraps(fn)
+        def wrapped(*args, **kwargs):
+            sess = GlobalDB.db().session
+            req_args = webargs_parser.parse({
+                'existing_submission_id': webargs_fields.Int(missing=None),
+                'cgac_code': webargs_fields.String(missing=None),
+                'frec_code': webargs_fields.String(missing=None)
+            })
+
```
alburde1: Would it make sense to check here if all three are null to start? If so, we probably want to return a different error message, like "Must provide either a valid cgac/frec code or an existing submission ID." Or is that too much detail?

michaeldhess: If you think it'd be more clear, I can do that. But it wouldn't affect the speed or anything because nothing really happens between line 174 and 187.

alburde1: If you think this is clear enough about what the problem is then this is fine, I was just wondering if we should cover that base or not.

PR #1300, opened 2018-07-13
fix program activity file locally

IsaacRay:
**High level description:**
Hotfix to deal with local issues with python initialize -i caused by sys.exit used in program activity loader. Also replaces the local file with a current version of the PA file from OMB.

**Technical details:**
File is a direct replacement. To fix the initialize -i local issue, we check if local and don't do a sys.exit if we are.

**Link to JIRA Ticket:**
N/A

- [ N/A] Unit & integration tests updated with relevant test cases
- [ N/A] Frontend impact assessment completed


tests/unit/dataactvalidator/test_load_program_activity.py
```diff
@@ -99,7 +99,7 @@ def test_load_program_activity_data(mocked_get_pa_file, mocked_get_current_date,
 def test_load_program_activity_data_only_header(mocked_get_pa_file, mocked_get_current_date, mocked_get_stored_date,
                                                 mocked_set_stored_date, monkeypatch):
     """ Test actually loading the program activity data """
-    monkeypatch.setattr(load_program_activity, 'CONFIG_BROKER', {'use_aws': False})
+    monkeypatch.setattr(load_program_activity, 'CONFIG_BROKER', {'use_aws': False, "local": False})
```
alburde1: Just since everything else is single quotes, can this also be single quotes?

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -120,13 +120,14 @@ def load_program_activity_data(base_path):
             data = pd.read_csv(program_activity_file, dtype=str)
         except pd.io.common.EmptyDataError as e:
             log_blank_file()
-            sys.exit(4)  # exit code chosen arbitrarily, to indicate distinct failure states
-
+            if not exit_if_nonlocal(4):  # exit code chosen arbitrarily, to indicate distinct failure states
```
alburde1: We don't really need to make these `if` statements. Just run `exit_if_nonlocal` and then return after it, since it won't ever get to the return if it's not local

dataactvalidator/scripts/load_program_activity.py
```diff
@@ -197,3 +201,10 @@ def lowercase_or_notify(x):
 def log_blank_file():
     """ Helper function for specific reused log message """
     logger.error("File was blank! Not loaded, routine aborted.")
+
+
+def exit_if_nonlocal(exit_code):
+    if not CONFIG_BROKER['local']:
+        sys.exit(exit_code)
+    else:
+        return True
```
alburde1: Returning true would mean that none of the returns for checking `if not exit_if_nonlocal` would ever happen. Is that the intent? If not, then this isn't even needed. Just exit without checking for the else

PR #1304, opened 2018-07-17
DEV-1250: Inbound API for FABS

IsaacRay:
**High level description:**
Allows the submission of FABS file via API request

**Technical details:**
Request is a POST, with content type of multipart/form-data. Params such as agency_code and existing_submission can be submitted as strings, while the fabs variable can now be submitted as a file binary. A curl request would look like this:

curl -i -X POST /
    -H "x-session-id: abcdefg-1234567-hijklmno-89101112"  
    -H "Content-Type: multipart/form-data" 
    -F 'agency_code=2000'
    -F "fabs=@/Users/isaacray/Desktop/fabs.csv" 
/v1/upload_detached_file/
 
**Link to JIRA Ticket:**
[DEV-1250](https://federal-spending-transparency.atlassian.net/browse/DEV-1250)

- [X] Unit & integration tests updated with relevant test cases


dataactbroker/handlers/fileHandler.py
```diff
@@ -954,18 +970,21 @@ def build_file_map(self, request_params, file_type_list, response_dict, upload_f
         """
         for file_type in file_type_list:
             # if file_type not included in request, skip it, checks for validity are done before calling this
-            if not request_params.get(file_type):
+            if not file_dict.get(file_type):
                 continue
-
-            file_name = request_params.get(file_type)
+            file = file_dict.get(file_type)
```
michaeldhess: Can we use a more descriptive name here than`file`?

IsaacRay: So, I struggled a bit with this because at this point in the code, that variable "file" may refer to either a) a string which is the filename, or b) the actual file binary... its not determined yet. The next few lines that follow this parse out whether its a string or a file. So I picked "file" because it kinda covers both, but I guess it could be something more specific... like "file_reference"? open to any other thoughts though...

michaeldhess: I think `file_reference` works. I just try to avoid `file` because it was a function in earlier versions of python.

APISubmissionGuide.md
```diff
@@ -142,29 +142,30 @@
...
+        - `success`: boolean indicating whether the file went through
         - `submission_id`: int, ID of the submission that was created
-- Step 2: Upload FABS file (Process to be defined)
-- Step 3: Call `/v1/finalize_job/` (POST) for the FABS file
-    - Payload:
-        - `upload_id`: int, ID of the file upload received from the `submit_files` response
-    - Response:
-        - `success`: boolean, successful or failed progression of the jobs
+
+    - Example cUrl request:
```
michaeldhess: `curl`

tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -404,6 +404,25 @@ def test_submission_report_url_s3(monkeypatch):
...
+    upload_files = []
+    response_dict = {}
+    file_type_list = ["fabs"]
+    file_dict = {"fabs": "fabs_file.csv"}
+
+    def side_effect(filename):
+        return "123_"+filename
+    s3_url_handler = Mock()
+    s3_url_handler.get_timestamped_filename = Mock(side_effect=side_effect)
+    monkeypatch.setattr(fileHandler, 'S3Handler', s3_url_handler)
```
michaeldhess: Instead of this (and the previous 4 lines) you should be able to just do: ```monkeypatch.setattr(s3Handler, 'get_timestamped_filename', Mock(return_value="123+"+filename))```.

APISubmissionGuide.md
```diff
@@ -95,32 +95,19 @@
 ## FABS Submission Process
 
 ### Upload FABS File
-- Step 1: Call `/v1/upload_detached_file/` (POST) (Note, this is NOT the file upload)
-    - Header:
-        - `X-Session-ID`: string, session token id
+- Call `/v1/upload_detached_file/`
     - Payload:
```
michaeldhess: All of the technical documentation should be in `dataactbroker/README.md`, with just references in `APISubmissionGuide.md`.

APISubmissionGuide.md
```diff
@@ -95,32 +95,19 @@
...
-- Step 3: Call `/v1/finalize_job/` (POST) for the FABS file
-    - Payload:
-        - `upload_id`: int, ID of the file upload received from the `submit_files` response
-    - Response:
-        - `success`: boolean, successful or failed progression of the jobs
+- For details on its use, click [here](./dataactbroker/README.md#post-v1upload_detached_file)
 
 ### Validate FABS File
-- Validations are automatically started by `finalize_job`
+- Validations are automatically started once the upload completes.
```
michaeldhess: This only happens with backend file upload. We should keep the previous documentation in there as well. Something like "Validations are automatically started:" with two bullets "by `finalized_job` when using the front end" and "once the upload completes when using the API".

This is right for your changes, but we don't know how long it'll be until the frontend starts using your changes.

dataactbroker/README.md
```diff
@@ -294,6 +294,31 @@ Example output:
 }
 ```
 
+### POST "/v1/upload_detached_file/"
```
michaeldhess: We want `####` here (4 #'s).

dataactbroker/permissions.py
```diff
@@ -210,14 +212,21 @@ def inner(fn):
...
-            req_args = webargs_parser.parse({
-                'agency_code': webargs_fields.String(missing=None),
-                'existing_submission_id': webargs_fields.String(missing=None)
-            })
+            try:
+                req_args = {
+                    'agency_code': RequestDictionary.derive(request).get('agency_code', None),
+                    'existing_submission_id': RequestDictionary.derive(request).get('existing_submission_id', None)
+                    }
+            except:
```
michaeldhess: What Exception are we picking up here? We should be specific.

tests/unit/dataactbroker/test_file_handler.py
IsaacRay: I don't think I can (just tried)... I believe return_value doesn't have access to the calling args, so I get a NameError on 'filename'.

IsaacRay: Oh, I might be able to do a lambda though, lemme try that

dataactbroker/handlers/fileHandler.py
```diff
@@ -77,6 +78,7 @@ class FileHandler:
     FILE_TYPES = ["appropriations", "award_financial", "program_activity"]
     EXTERNAL_FILE_TYPES = ["award", "award_procurement", "executive_compensation", "sub_award"]
     VALIDATOR_RESPONSE_FILE = "validatorResponse"
+    UPLOAD_FOLDER = '/data-act/backend/tmp'
```
michaeldhess: We should be using env variables here, because this path is not used for every frontend dev env.

PR #1326, opened 2018-07-26
Initialize local postgres container

sharb:
1. Made local password consistent in the configs
2. Added data load and user creation part of `docker-compose up` process 
3. Documentation updated to reflect those changes


sharb:
@michaeldhess these commands are run on every `docker-compose up`
```
/bin/sh -c " \
            cd ./dataactcore
            sleep 9s
            alembic upgrade head
            cd scripts
            python initialize.py -i
            python initialize.py -a
            python ../../dataactbroker/app.py
          "
```

michaeldhess:
`python initialize.py -i` and `-a` initialize a DB and load in the reference data. Since we're linking to a local DB which should only be created once, it doesn't make sense for the dev to have to wait (sometimes up to 15 min, depending on the dev's local files) for these scripts to finish before they can start working. 

Also, according to @willkjackson, each container should not need the `sleep` command and should instead just be dependent on the previous container.

PR #1333, opened 2018-08-06
Hotfix: Fix monthly submission creation

michaeldhess:
**High level description:**
Fix bug that stopped users from creating monthly submissions

**Technical details:**
Multipart form-data casts the boolean to a string. We need to parse the string to understand if it's a truthy value. Added tests.

- [ ] Unit & integration tests updated with relevant test cases

### DIFF
--- a/dataactcore/utils/requestDictionary.py
+++ b/dataactcore/utils/requestDictionary.py
@@ -46,7 +46,7 @@ class RequestDictionary:
                 request_data = request.form.to_dict()
                 request_data['_files'] = request.files
                 for key, value in request_data.items():
-                    if value in ('null', 'None'):
+                    if str(value).upper() in ('NULL', 'NONE'):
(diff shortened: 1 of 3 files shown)


dataactcore/utils/requestDictionary.py
```diff
@@ -46,9 +46,8 @@ def derive(request, optional_request=False):
                 request_data = request.form.to_dict()
                 request_data['_files'] = request.files
                 for key, value in request_data.items():
-                    if value == 'null':
+                    if value in ('null', 'None'):
```
dpb-bah: Could we do a `value.lower().strip() in ('null', 'none')` to account for bad values ('NULL', 'none')

dpb-bah: Could we do a case.insensitive check? `value.lower() in ('null', 'none')`

PR #1336, opened 2018-08-07
DEV-1359: submission dashboard last modified date filter

alburde1:
**High level description:**
Adding a last modified date filter to `list_submissions` endpoint

**Technical details:**
Adding ability to filter by a last modified date range. **looking for advice** would this be better as a `last_modified_range` array with 2 elements? Opinions much appreciated.

**Link to JIRA Ticket:**
[DEV-1359](https://federal-spending-transparency.atlassian.net/browse/DEV-1359)

### DIFF
--- a/dataactbroker/README.md
+++ b/dataactbroker/README.md
@@ -933,8 +933,10 @@ This endpoint lists submissions for all agencies for which the current user is a
     "d2_submission": False,
     "filters": {
         "submission_ids": [123, 456],
-        "start_date": "01/01/2018",
-        "end_date": "01/10/2018"
+        "last_modified_range": {
+            "start_date": "01/01/2018",
+            "end_date": "01/10/2018"
+        }
     }
 }
 ```
@@ -959,8 +961,9 @@ This endpoint lists submissions for all agencies for which the current user is a
 - `d2_submission` - **optional** - a boolean indicating if the submissions listed should be FABS or DABS (True for FABS). Defaults to `False` if not provided.
 - `filters` - **optional** - an object containing additional filters to narrow the results returned by the endpoint. Possible filters are:
     - `submission_ids` - an array of integers or strings that limits the submission IDs returned to only the values listed in the array.
-    - `start_date` - **required if `end_date` is included** - a string indicating the start date for the last modified date range (inclusive) (MM/DD/YYYY)
-    - `end_date` - **required if `start_date` is included** - a string indicating the end date for the last modified date range (inclusive) (MM/DD/YYYY)
+    - `last_modified_range` - an object containing a start and end date for the last modified date range. Both must be provided if this filter is used.
+        - `start_date` - a string indicating the start date for the last modified date range (inclusive) (MM/DD/YYYY)
+        - `end_date` - a string indicating the end date for the last modified date range (inclusive) (MM/DD/YYYY)
 
 ##### Response (JSON)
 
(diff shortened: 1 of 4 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -1506,13 +1506,29 @@ def add_list_submission_filters(query, filters):
...
+        if 'start_date' not in filters or 'end_date' not in filters:
+            raise ResponseException("If either start_date or end_date is provided in filters, both must be provided",
+                                    StatusCode.CLIENT_ERROR)
+        start_date = filters['start_date']
+        end_date = filters['end_date']
+        # Start and end dates must be in the format MM/DD/YYYY and be
+        if not (StringCleaner.is_date(start_date) and StringCleaner.is_date(end_date)):
+            raise ResponseException("Start or end date cannot be parsed into a date of format MM/DD/YYYY",
+                                    StatusCode.CLIENT_ERROR)
+        # Need to add a day to the end date to make it actually inclusive
```
michaeldhess: Why can't we do a `<=` with the `end_date` provided?

dataactbroker/handlers/fileHandler.py
```diff
@@ -1506,13 +1506,29 @@ def add_list_submission_filters(query, filters):
...
+    # Checking for submission ID filter
     if 'submission_ids' in filters:
         sub_list = filters['submission_ids']
         if sub_list and isinstance(sub_list, list):
             sub_list = [int(sub_id) for sub_id in sub_list]
             query = query.filter(Submission.submission_id.in_(sub_list))
         elif sub_list:
             raise ResponseException("submission_ids filter must be null or an array", StatusCode.CLIENT_ERROR)
+    # Date range filter
+    if 'start_date' in filters or 'end_date' in filters:
```
michaeldhess: Shouldn't we also check that the `end_date` is later than the `start_date`?

dataactbroker/README.md
```diff
@@ -932,7 +932,9 @@ This endpoint lists submissions for all agencies for which the current user is a
     "order": "desc",
     "d2_submission": False,
     "filters": {
-        "submission_ids": [123, 456]
+        "submission_ids": [123, 456],
+        "start_date": "01/01/2018",
```
michaeldhess: To answer your question, I think these start/end dates might get in the way if we end up wanting to be able to filter by two different types of date ranges. I also think an array with two values wouldn't be the best way to do it either. This is what I recommend:
```
"filters": {
        "submission_ids": [123, 456],
        "last_modified_range": {
            "start_date": "01/01/2018",
            "end_date": "01/10/2018"
        }
}
```

dataactbroker/handlers/fileHandler.py
alburde1: Because technically it's `date 0:00:00` so it checks the same day/time twice. If it was the end of the day that'd be fine, but essentially by adding a day we're checking `< next day 0:00:00`

dataactbroker/handlers/fileHandler.py
alburde1: I don't think we do that with other date ranges on the backend. They give us a bad date range, they get no results. Do we want to alter how we do that here?

dataactbroker/handlers/fileHandler.py
michaeldhess: Fair enough, adding a day is simpler than adding `23:59`.

dataactbroker/handlers/fileHandler.py
michaeldhess: I don't think that's within the scope of this ticket, but if we don't do it elsewhere we should. I can create a JIRA ticket.

PR #1360, opened 2018-08-23
DEV-1354: Handle upload Job errors

michaeldhess:
**High level description:**
Currently, if something goes wrong during an upload, the errors are raised to the user but nothing happens to the upload Job itself.

**Technical details:**
Handle the errors by marking the upload Job as "failed" and setting it's `error_message` to the raised Exception.

**Link to JIRA Ticket:**
[DEV-1354](https://federal-spending-transparency.atlassian.net/browse/DEV-1354)

- Merged concurrently with Frontend N/A
- Unit & integration tests updated with relevant test cases N/A


dataactbroker/handlers/fileHandler.py
```diff
@@ -266,15 +267,34 @@ def upload(file_ref, file_type, app, current_user):
...
+            # handle a missing JSON response
+            if json_response is None:
+                json_response = JsonResponse.error(Exception("Failed to catch exception"), StatusCode.INTERNAL_ERROR)
+
+            # handle errors within upload jobs
+            if json_response.status_code != StatusCode.OK and submission:
+                jobs = sess.query(Job).filter(Job.submission_id == submission.submission_id,
+                                              Job.job_type_id == JOB_TYPE_DICT['file_upload'],
+                                              Job.job_status_id == JOB_STATUS_DICT['running'],
+                                              Job.file_type_id.in_([FILE_TYPE_DICT_LETTER_ID['A'],
```
alburde1: Do we always want to set all 3 to failed? What if we're only reuploading one file? We don't want to fail the other 3 jobs do we?

dataactbroker/handlers/fileHandler.py
```diff
@@ -636,17 +657,31 @@ def upload_fabs_file(self, fabs):
...
+        finally:
+            # handle a missing JSON response
+            if json_response is None:
+                json_response = JsonResponse.error(Exception("Failed to catch exception"), StatusCode.INTERNAL_ERROR)
+
+            if json_response.status_code != StatusCode.OK and submission:
+                jobs = sess.query(Job).filter(Job.submission_id == submission.submission_id,
+                                              Job.job_type_id == JOB_TYPE_DICT['file_upload'],
+                                              Job.job_status_id == JOB_STATUS_DICT['running'],
+                                              Job.file_type_id == FILE_TYPE_DICT_LETTER_ID['FABS']).all()
```
alburde1: Are we ever going to have more than one upload job for a given FABS submission? Do we need the `.all()` and the loop? If we feel that it isn't too bad to do this we can leave it, just seems odd.

dataactbroker/handlers/fileHandler.py
michaeldhess: This should only set `running` Jobs to `failed`.

alburde1: I can read... I swear

PR #1387, opened 2018-09-20
DEV-1585 : Fixing validation error handling

IsaacRay:
**High level description:**
Better error handling for broker validation

**Technical details:**
We should be printing a traceback for exceptions so that they are easier to locate. Does that here.

**Link to JIRA Ticket:**
[DEV-1585](https://federal-spending-transparency.atlassian.net/browse/DEV-1585)


dataactvalidator/app.py
```diff
@@ -102,7 +103,7 @@ def run_app():
 
             except ResponseException as e:
                 # Handle exceptions explicitly raised during validation.
-                logger.error(str(e))
+                logger.error(traceback.print_exc())
```
alburde1: Are we sure this works? How does it know what it's printing?

IsaacRay: It prints the exception that was caught in the current thread. I've used it pretty extensively, and also, I had the Inbound API user who was running broker locally make this change in his local file in order to figure out what his issue was.

PR #1392, opened 2018-09-21
DEV-1532: Populate Missing TAS data

dpb-bah:
**High level description:**
Updated TAS script to populate TAS records with missing fields to eventually be used on the website. 

**Technical details:**
- Modified TAS script to work with local `car_tas.csv`s and only populate fields with missing data.

**Link to JIRA Ticket:**
[DEV-1532](https://federal-spending-transparency.atlassian.net/browse/DEV-1532)

- [x] Data validation completed: Tested locally and on broker sandbox, was able to fill in all missing records
- [x] Appropriate Operations ticket(s) created: [OPS-356](https://federal-spending-transparency.atlassian.net/browse/OPS-356)
- [x] Frontend impact assessment completed (N/A)


dataactvalidator/scripts/loadTas.py
```diff
@@ -131,6 +143,25 @@ def existing_id(row, existing):
...
 
 
+def get_parser():
+    """ Generates list of command-line arguments
+
+        Returns:
+            argument parser to be used for commandline
+    """
+    tas_parser = argparse.ArgumentParser(description='Import data from the cars_tas.csv')
+    tas_parser.add_argument('--tas_file', '-f', type=str, default=None, help='Path to specifc cars_tas file to use')
```
alburde1: Do we have to manually enter every file? It might make more sense to use boto to gather the historical files one at a time in a loop until there are no more rows with empty `budget_function_code` columns OR we run out of historical files.

dataactvalidator/scripts/loadTas.py
```diff
@@ -74,23 +74,40 @@ def update_tas_lookups(csv_path):
...
+    if only_fill_in:
+        # Pull in the empty records
+        empty_records = sess.query(TASLookup).filter_by(budget_function_code=None).values('account_num')
+        empty_account_nums = [int(empty_record[0]) for empty_record in empty_records]
+        # find which incoming records can fill in the empty records
+        relevant_old_data = old_data[old_data['account_num'].isin(empty_account_nums)]
+        # Fill them in. If budget_function_code is empty, the following columns have also been empty.
+        fill_in_cols = ['account_title', 'budget_bureau_code', 'budget_bureau_name', 'budget_function_code',
+                        'budget_function_title', 'budget_subfunction_code', 'budget_subfunction_title',
+                        'reporting_agency_aid', 'reporting_agency_name']
```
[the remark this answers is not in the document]
dpb-bah: The choice of columns was based all the columns that were completely missing from dev/stg/prod. Can confirm:

```
select * from tas_lookup
where (
	account_title is null and 
	budget_bureau_code is null and
	budget_bureau_name is null and
	budget_function_title is null and
	budget_subfunction_code is null and
	budget_subfunction_title is null and
	reporting_agency_aid is null and
	reporting_agency_name is null)
	and budget_function_code is not null;
```
returns 0 results on dev.  Since the issue arose from not backfilling data when new columns were added, it's safe to say these columns represent the problem.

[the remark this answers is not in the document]
dpb-bah: If we wanted a more dynamic solution we could generate this list by figuring out which columns are completely null for future solutions

dataactvalidator/scripts/loadTas.py
```diff
@@ -131,6 +143,25 @@ def existing_id(row, existing):
...
 
+def get_parser():
+    """ Generates list of command-line arguments
+
+        Returns:
+            argument parser to be used for commandline
+    """
+    tas_parser = argparse.ArgumentParser(description='Import data from the cars_tas.csv')
+    tas_parser.add_argument('--tas_file', '-f', type=str, default=None, help='Path to specifc cars_tas file to use')
+    tas_parser.add_argument('--only_fill_in', '-o', action='store_true', help='Only fill in records with missing data')
```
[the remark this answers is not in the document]
dpb-bah: The `store_true` action defaults the value to `False` if not provided.

dataactvalidator/scripts/loadTas.py
dpb-bah: Fixed to work with the versions of the file in reverse chronological order.

dataactvalidator/scripts/loadTas.py
```diff
@@ -131,6 +143,25 @@ def existing_id(row, existing):
...
+
 if __name__ == '__main__':
+    parser = get_parser()
+    args = parser.parse_args()
     configure_logging()
-    load_tas()
+
+    tas_file = args.tas_file
+    only_fill_in = args.only_fill_in
+    if tas_file and not os.path.exists(tas_file):
```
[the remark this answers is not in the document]
dpb-bah: Modified since to pull from S3 instead of locally

dataactvalidator/scripts/loadTas.py
```diff
@@ -159,30 +214,38 @@ def existing_id(row, existing):
         Args:
             row: row to check in
             existing: Dict[account_num, List[TASLookup]]
+        Returns:
+            account number for existing row
     """
     for potential_match in existing[row['account_num']]:
```
alburde1: Dumb question, why is this a `for`? As soon as you return it's over? In fact, do we use this function anywhere at all? Can we just get rid of it?

dataactvalidator/scripts/loadTas.py
```diff
@@ -74,44 +109,97 @@ def update_tas_lookups(csv_path):
...
 
     sess.commit()
-    logger.info('%s records in CSV, %s existing', len(data.index), sum(data['existing_id'].notnull()))
 
 
-def load_tas(tas_file=None):
-    """Load TAS file into broker database. """
-    # read TAS file to dataframe, to make sure all is well
-    # with the file before firing up a db transaction
+def load_tas(tas_file=None, backfill_historic=False):
```
alburde1: At no point do we use the `tas_file` argument, there isn't even code for it to be used. In fact, there isn't even an `else` to handle what to do if we have a `tas_file`. Do we really need to keep it or any logic pertaining to it?

dataactvalidator/scripts/loadTas.py
```diff
@@ -74,44 +109,97 @@ def update_tas_lookups(csv_path):
...
     if not tas_file:
         if CONFIG_BROKER["use_aws"]:
-            s3connection = boto.s3.connect_to_region(CONFIG_BROKER['aws_region'])
-            s3bucket = s3connection.lookup(CONFIG_BROKER['sf_133_bucket'])
-            tas_file = s3bucket.get_key("cars_tas.csv").generate_url(expires_in=600)
+            # Storing version dictionaries in the list to prevent getting all the links at once and possibly work with
+            # expired AWS links
+            s3connection = boto3.client('s3', region_name=CONFIG_BROKER['aws_region'])
+            # list_object_versions returns the versions in reverse chronological order
+            tas_files = s3connection.list_object_versions(Bucket=CONFIG_BROKER['sf_133_bucket'],
```
alburde1: Do we need to get the versions if we aren't backfilling? Could this part be inside the `if` so we can just grab the latest and not worry about versions if they aren't needed?

dataactvalidator/scripts/loadTas.py
```diff
@@ -126,11 +214,38 @@ def existing_id(row, existing):
...
+
+        Returns:
+            argument parser to be used for commandline
+    """
+    tas_parser = argparse.ArgumentParser(description='Import data from the cars_tas.csv')
+    tas_parser.add_argument('--backfill_historic', '-b', action='store_true', help='Backfill tas with historical data')
+    return tas_parser
+
 if __name__ == '__main__':
+    parser = get_parser()
```
alburde1: For consistency, in Broker we generally define the arg parser in the main function (whether it's called `main()` or something else), rather than creating a `get_parser` function. This is usually done outside this `if __name__` section and after logging is configured.

Not something that absolutely has to be changed, just something to remember for next time.

dataactvalidator/scripts/loadTas.py
dpb-bah: I was keeping it in on the off chance we provide a local argument. It can be removed for now.

dataactvalidator/scripts/loadTas.py
dpb-bah: It's used by pandas to figure out which rows are updates vs additions: line 209. It basically goes over the list of account numbers found in the database and sees if the incoming rows's account numbers are in said list of account numbers.

dataactvalidator/scripts/loadTas.py
dpb-bah: Understood. Was doing it for code simplicity but I agree it can be taxing in the future.

dataactvalidator/scripts/loadTas.py
dpb-bah: Sounds good.

PR #1449, opened 2018-11-19
DEV-1696: Fixing docs for /v1/login

IsaacRay:
**High level description:**
Making it clear that /v1/login/ is not a production endpoint

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-1696](https://federal-spending-transparency.atlassian.net/browse/DEV-1696)

- [NA] Backend review completed


IsaacRay:
We don't make any reference to /v1/login/ at all in the submission guide, and it is my opinion that the less we mention this endpoint the better, so I think we should not add anything to the guide.

PR #1476, opened 2018-12-13
DEV-1378: Update list_submissions param names

alburde1:
**High level description:**
Updating parameters sent to list_submissions

**Technical details:**
Update `d2_submission` to `fabs` in the `list_submissions` route to better align with the rest of our code.

**Link to JIRA Ticket:**
[DEV-1378](https://federal-spending-transparency.atlassian.net/browse/DEV-1378)

- [ ] Merged concurrently with [Frontend#981](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/981)


tests/unit/dataactbroker/test_file_handler.py
```diff
@@ -21,7 +21,7 @@
 
 
 def list_submissions_result(d2_submission=False):
```
michaeldhess: Shouldn't we change this instance of `d2_submission` as well?

alburde1:
I think that we should hold off on d2_submission changes in terms of the model itself because that is a lot more things changed and might affect website as well. That should probably be its own ticket, which I will check for the existence of and create if it doesn't exist.

PR #1493, opened 2019-01-31
DEV-2089: file E errors

alburde1:
**High level description:**
Communicate more clearly to the user what the problem with their E file was when possible.

**Technical details:**
Catch errors like broken SAM and invalid credentials and return them in a more readable format so the user is less confused about what exactly is going wrong.

Also discovered that the error message wasn't getting set right and never actually updated properly or finished out the process after that. Fixed that bug.

**Link to JIRA Ticket:**
[DEV-2089](https://federal-spending-transparency.atlassian.net/browse/DEV-2089)


dataactcore/utils/fileE.py
```diff
@@ -102,10 +151,17 @@ def suds_to_row(suds_obj):
...
-    numbers."""
+    """ Soup-to-nuts creates a list of Row tuples from a set of DUNS numbers.
+
+        Args:
+            duns_list: A list of DUNS to search for
+    """
     if config_valid():
-        client = Client(CONFIG_BROKER['sam']['wsdl'])
+        try:
+            client = Client(CONFIG_BROKER['sam']['wsdl'])
```
dpb-bah: This line is copied in `dataactcore/utils/parentDuns.py`. Could you include this check in there too, or maybe refactor it out into a helper function which both can use?

PR #1510, opened 2019-02-07
DEV-1157: USPS loader updates

alburde1:
**High level description:**
Update USPS loader so it doesn't interrupt submissions like it currently does.

**Technical details:**
Load zips in a staging table and hot swap the tables in a transaction to prevent half-updated data from being left in the zips table or from data being missing if a validation is started while the USPS loader is running.

**Link to JIRA Ticket:**
[DEV-1157](https://federal-spending-transparency.atlassian.net/browse/DEV-1157)


dataactvalidator/scripts/read_zips.py
```diff
@@ -61,30 +102,49 @@ def update_state_congr_table_census(census_file, sess):
...
-            insert_statement = insert(Zips).values(**new_zip).\
-                on_conflict_do_nothing(index_elements=[Zips.zip5, Zips.zip_last4])
-            sess.execute(insert_statement)
+            sess.execute('INSERT INTO temp_zips '
+                         '(updated_at, created_at, zip5, zip_last4, county_number, state_abbreviation, '
+                         'congressional_district_no) VALUES {} '
+                         'ON CONFLICT DO NOTHING'.format(new_zip))
 
-            if i % 10000 == 0:
+            if i % 1000 == 0:
```
dpb-bah: Can we have a note/comment on the batches of 1000?

dataactvalidator/scripts/read_zips.py
```diff
@@ -25,8 +24,44 @@
...
+        Args:
+            sess: the database connection
+    """
+    # Getting indexes before dropping the table
+    indexes = Zips.__table__.indexes
+
+    logger.info("Hot swapping temporary zips table to official zips table.")
+
+    # Do everything in a transaction so it doesn't affect anything until it's completely done
+    sess.execute("BEGIN;")
```
dpb-bah: Instead of multiple `execute` calls, I wonder if it'll be cleaner or faster to have everything but the index changes be as one string/file and one `execute`. That's my preference but this is optional.

willkjackson: Agree.

PR #1534, opened 2019-03-12
DEV-2246 Close DB Session After Job

kwhickey:
**High level description:**
Reduce the gradual climb of memory utilization in the broker validator with each job processed. Tie the lifecycle of the SQL Alchemy Session object to the duration of a single validator job, which consists of processing a single message on the SQS queue for the broker validator component. After each message, the DB Session should be removed (closed).

**Technical details:**
See also [DEV-2246](https://federal-spending-transparency.atlassian.net/browse/DEV-2246)
Broker Validator seems to be having memory issues related to the SQLAlchemy `Session` life-cycle when being used by a python process (`app.py`). Here is some background on how it works in Broker:
https://github.com/fedspendingtransparency/data-act-broker-backend/pull/1431#discussion_r230556878

Specifically, this part is relevant, as it describes the life cycle used in validator, as determined by `dataactcore/interfaces/db.py` (but read the rest for context):

> 2. **Per Process.** If not [processing web requests], then we just store a reference to the `Session` instance ~statically on the `GlobalDB` class itself~ on the `flask.g` global dictionary and lives as long as the app context, which, in the case of a console app like this, resolves to basically one instance of `Session` per process using this code. This is all wrapped in retrieval of the "cached" `Session` object via `GlobalDb.db().session`

Given that the life cycle of the `Session` and a validaory process are tied one-to-one, this means that a `Session` object probably lasts for days and days processing 10s, 100s, or 1000s of jobs. Every query the session runs for each job stores the retrieved objects in its [`identity_map`](https://docs.sqlalchemy.org/en/latest/orm/session_api.html#sqlalchemy.orm.session.Session.identity_map) dictionary, causing the overall grown in memory consumption for that process and for the Linux EC2 server instance that hosts those processes. See DEV-2246 attachment for images of this. 

The fix is to call `GlobalDB.close()` after each message which closes/removes (see [`remove()`](https://docs.sqlalchemy.org/en/latest/orm/contextual.html#sqlalchemy.orm.scoping.scoped_session.remove) the `Session` object and all of its contained objects. The next message dequeued and processed will call `GlobalDb.db().session` again, find no `Session`, and instantiate a new one for processing that message.

**Link to JIRA Ticket:** [DEV-2246](https://federal-spending-transparency.atlassian.net/browse/DEV-2246)

_Checklist Details_
- Backend only change, no frontend
> - Unit & integration tests updated with relevant test cases
- We do not have automated integration tests that run `/dataactvalidator/app.py`, but rather run the processing functions within. Working to test this manually while the `do not merge` label is on.


kwhickey:
Closing this in favor of doing DEV-2250 instead.
See: https://federal-spending-transparency.atlassian.net/browse/DEV-2246?focusedCommentId=38580&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-38580 for details

PR #1586, opened 2019-05-07
Sprint 81: Dev -> Staging

alburde1:


dpb-bah:
Changelog:
* Updated location loader to only update on incoming file changes.
* Updated validator to use an SQS dispatcher to optimize memory, delegate queued jobs, and cleanup jobs on unexpected termination.
* Updated several python libraries used by the Broker.
* Added subtier agency(9505) to agency list and updated relevant transactions.
* Updated SQL rules to ensure blank entries are properly treated as zero dollars.
* Updated CFDA loader to properly decode CFDA values and cleaned data removing special characters involved with the CFDA loader.

PR #1597, opened 2019-05-23
DEV-2592: fabs exec comp

alburde1:
**High level description:**
Add executive compensation to FABS transactions

**Technical details:**
Adding columns to store executive compensation data in published FABS records and adding derivations for these values moving forward.

**Link to JIRA Ticket:**
[DEV-2592](https://federal-spending-transparency.atlassian.net/browse/DEV-2592)


dataactbroker/handlers/fileHandler.py
```diff
@@ -674,6 +675,16 @@ def publish_fabs_submission(submission):
                 state_dict[state.state_code.upper()] = state.state_name
             del states
 
+            duns_list = sess.query(DUNS).filter(DUNS.high_comp_officer1_full_na.isnot(None)).all()
```
dpb-bah: Making a general note that this list could be memory intensive if DUNS with executive compensation values increases but currently, that size of this list should be around five thousand and so it's not a concern for now.

PR #1599, opened 2019-05-24
Attempt to fix validator SQSWorkDispatcher unit test race condition

kwhickey:
**High level description:**
Race condition occurs during one of the unit tests, where the child worker process is supposed to have been killed during the 2nd attempt at running the exit_handler cleanup function, but is found to still be running.

**Technical details:**
- Put a 0.25 second buffer between when `os.kill(pid)` is called on the child worker process, and when the cleanup handler checks that processes status
- Also fix the assertions so that if it is found to still be running, that will fail the test, rather than allowing the test to carry on and eventually fail due to the worker still running
- Note, the tests should pass and because this is a race condition, this fix is only really tested by the tests _consistently_ passing without an intermittent failure

**Link to JIRA Ticket:**
[DEV-2250](https://federal-spending-transparency.atlassian.net/browse/DEV-2250)

- N/A Merged concurrently with [Frontend#1234](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/1234)
- N/A Frontend impact assessment completed


tests/integration/sqs_work_dispatcher_tests.py
```diff
@@ -1033,36 +1033,42 @@ def test_cleanup_second_try_succeeds_after_killing_worker_with_dlq(self):
         queue = sqs_queue()
         queue.send_message(MessageBody=msg_body)
         worker_sleep_interval = 0.05  # how long to "work"
+        cleanup_timeout = int(worker_sleep_interval + 3)  # how long to allow cleanup to run, max (integer seconds)
```
dpb-bah: Can we move this comment up? It's not officially in our style guide yet but what we generally do.

kwhickey: Can we leave it as-is?
- There's probably 100s of cases in this file that have comments on the same line of the code that would need to be fixed as well.
- Being that it's a unit tests, comment readability probably isn't super critical
- It can be helpful too to keep the flow of the code uninterrupted.

dpb-bah: I can leave it in for now but we will most likely put it in our style guide soon and resolve them in a tech improvement.

PR #1604, opened 2019-06-03
Sprint 83: Dev -> Staging

alburde1:


dpb-bah:
* Updated FSRS loader to populate Subaward table including derivations for File F and other FSRS data points that will be used by USASpending.gov.
* Updated FABS derivations and FPDS loader to include Executive Compensation data in transaction data to be used by USASpending.gov.

PR #1610, opened 2019-06-11
add new variable so that ssh key pass is not required to match user pass

willkjackson:


dataactcore/utils/duns.py
```diff
@@ -35,24 +35,29 @@ def get_client(ssh_key=None):
...
-    pkey = None
+
     if ssh_key:
-        with RetrieveFileFromUri(ssh_key, binary_data=False).get_file_object() as key_obj:
-            pkey = paramiko.RSAKey.from_private_key(key_obj, password=password)
+        host = sam_config.get('host_ssh')
+        username = sam_config.get('username_ssh')
+        password = sam_config.get('password_ssh')
+
+        ssh_key_file = RetrieveFileFromUri(ssh_key, binary_data=False).get_file_object()
```
dpb-bah: Pretty sure this still needs the `with`.

willkjackson: This was changed to clarify that it's getting the key file in one call and using `ssh_key_password` with it in the following call.

dpb-bah:
Tested with Jenkins.

PR #1611, opened 2019-06-12
DEV-2376: Multiple cfda fabs

alburde1:
**High level description:**
Changing how transaction unique key is derived to allow for multiple CFDA numbers and updating a script so we can use it to fix the existing keys.

**Technical details:**
Reorganize to put awarding agency sub tier code at the front of the afa_generated_unique so it can never start with `-none-` and mess Excel up. Also add CFDA number to the unique key. Updating the award key derivation script to allow it to be used for re-deriving unique transaction keys as well.

**Link to JIRA Ticket:**
[DEV-2376](https://federal-spending-transparency.atlassian.net/browse/DEV-2376)

- OPs Ticket: [OPS-735](https://federal-spending-transparency.atlassian.net/browse/OPS-735)


dataactcore/scripts/populate_unique_keys.py
```diff
@@ -0,0 +1,188 @@
...
+                                            func.coalesce(dap.agency_id, '-none-'), '_',
+                                            func.coalesce(dap.parent_award_id, '-none-'), '_',
+                                            func.coalesce(dap.referenced_idv_agency_iden, '-none-')))
+                # IDV
+                if 'IDV' in types:
+                    update_keys(dap, 'FPDS', key_type, 'IDV', years,
+                                func.concat('CONT_IDV_', func.coalesce(dap.piid, '-none-'),
+                                            '_', func.coalesce(dap.agency_id, '-none-')))
+            else:
+                # awards
```
dpb-bah: transactions?

dataactcore/scripts/populate_unique_keys.py
```diff
@@ -0,0 +1,188 @@
...
+                update_keys(dafa, 'unpublished FABS', key_type, '', None,
+                            func.concat(func.coalesce(dafa.awarding_sub_tier_agency_c, '-none-'), '_',
+                                        func.coalesce(dafa.fain, '-none-'), '_',
+                                        func.coalesce(dafa.uri, '-none-'), '_',
+                                        func.coalesce(dafa.cfda_number, '-none-'), '_',
+                                        func.coalesce(dafa.award_modification_amendme, '-none-')))
+
+        # published FABS
+        if 'publishedFABS' in models:
+            pafa = PublishedAwardFinancialAssistance
```
dpb-bah: Stupid question I can't tell with my eyes: are `unpublishedFABS` and `publishedFABS` the same code, just different models? If so, I'd consider refactoring but it's ultimately your call.

alburde1: There are a few differences including model and whether we use years or not so the only way to really refactor would be with a loop and some more tweaking. Probably overall not much cleaner code

PR #1615, opened 2019-06-14
DEV-2507: Unified Agency List

dpb-bah:
**High level description:**
Merged all the different sources of agency data from Broker and USASpending.gov to a single source and updating the Broker to use that.

https://files.usaspending.gov/reference_data/agency_codes.csv

**Technical details:**
* Includes script to generate the file for quick edits and ensuring its accuracy.

**Link to JIRA Ticket:**
[DEV-2507](https://federal-spending-transparency.atlassian.net/browse/DEV-2507)

- [x] Ensured discrepancies between data in the new agency file and agency data in broker production. 
- [x] [OPs ticket](https://federal-spending-transparency.atlassian.net/browse/OPS-755?oldIssueView=true&searchContainerId=11205&searchContentType=issue&searchObjectId=44233&searchSessionId=45f3bb6d-57d6-4a85-9466-3a6ac83cc5de) to reload the agencies on each environment


dpb-bah:
Marking `Do Not Merge` as there are slight differences between the master agency file data and the current broker production data.

dpb-bah:
Also noting that it was attempted to only reload based on new data with `load_agencies`. It led to needing to update `check_dataframe_diff` to handle lambda functions based on several columns and the requirement to drop additional columns from the dataframe, both of which are possible but would require more time/scope to properly test.

dpb-bah:
Script has been updated to resolve discrepancies.

dpb-bah:
Loader has been updated to only load on changes.

dataactbroker/helpers/pandas_helper.py
```diff
@@ -33,11 +36,16 @@ def check_dataframe_diff(new_data, model, id_col, sort_cols, lambda_funcs=None):
...
     current_data = pd.read_sql_table(model.__table__.name, sess.connection(), coerce_float=False)
+
+    # Apply any lambda functions provided to update values if needed
+    for col_name, lambda_func in lambda_funcs:
+        current_data[col_name] = current_data.apply(lambda_func, axis=1)
+
     # Drop the created_at and updated_at for the same reason as above, also drop the pk ID column for this table
     try:
-        current_data.drop([id_col, 'created_at', 'updated_at'], axis=1, inplace=True)
+        current_data.drop([id_col, 'created_at', 'updated_at'] + fk_cols, axis=1, inplace=True)
```
alburde1: If we're deleting more than one column potentially, would it make sense to just use one variable and name it `del_cols` or something? Then it would always have at least the ID column in it.

dataactbroker/helpers/pandas_helper.py
```diff
@@ -33,11 +36,16 @@ def check_dataframe_diff(new_data, model, id_col, sort_cols, lambda_funcs=None):
...
     except ValueError:
-        logger.info('id column, created_at, or updated_at not found, drop skipped.')
+        logger.info('id column, fk_columns, created_at, or updated_at not found, drop skipped.')
```
alburde1: Based on the above comment, this log should probably be more of `created_at, updated_at, or at least one of the columns provided for deletion not found, drop skipped.`

dataactvalidator/scripts/load_agencies.py
```diff
@@ -132,22 +173,38 @@ def load_frec(file_name):
...
         {"frec": "frec_code", "cgac_agency_code": "cgac_code", "frec_entity_description": "agency_name",
-         "agency_abbreviation": "agency_abbreviation"},
+         "agency_abbreviation": "agency_abbreviation", "frec_cgac_association": "frec_cgac"},
         {"frec": {"keep_null": False}, "cgac_code": {"pad_to_length": 3}, "frec_code": {"pad_to_length": 4}}
     )
     # de-dupe
+    data = data[data.frec_cgac == 'TRUE']
+    data.drop(['frec_cgac'], axis=1, inplace=True)
     data.drop_duplicates(subset=['frec_code'], inplace=True)
-    # create foreign key dicts
```
alburde1: Why did we remove this one and not a future one? We should be consistent.

dataactbroker/helpers/pandas_helper.py
dpb-bah: I was considering that but made `fk_cols` in case to not mess with the code too much. Fixing.

dataactbroker/helpers/pandas_helper.py
```diff
@@ -6,22 +6,24 @@
 logger = logging.getLogger(__name__)
 
 
-def check_dataframe_diff(new_data, model, id_col, sort_cols, lambda_funcs=None):
+def check_dataframe_diff(new_data, model, sort_cols, del_cols=None, lambda_funcs=None):
```
alburde1: Since the ID wasn't optional, I don't think this array should be either. It should contain at the very least the non-optional ID column. If there's a super weird exception at some point the value passed can just be an empty array

dpb-bah:
Tested locally again. Discovered loading on truncated tables led to a pandas error (due to the apply()). Checking for empty dataframes resolved it.

PR #1617, opened 2019-06-18
Sprint 84: Dev -> Staging

alburde1:


dpb-bah:
* Optimized FSRS loader scripts for both daily FSRS updates, resolving unlinked subawards, and entirely rederiving Subaward data.
* Reworked File F generation to use already derived Subaward data instead of deriving from FSRS each time.
* Updated checks on the unique FABS transaction key to be case-insensitive and properly dealt with any resulting duplicate transactions.
* Reordered elements in the unique FABS transaction key and included CFDA Number to the uniqueness.
* Updated several FABS 2 validation descriptions to include CFDA number.    
* Added an upper index on afa_generated_unique for performance.
* Updated country code loader to use a reference country codes list to be shared with USAspending.gov.

PR #1639, opened 2019-07-15
Sprint 86: Dev -> Staging

alburde1:


dpb-bah:
Release Notes:
* Added additional metadata with uploads for easier tracking.
* Enforced user uploads to only accept case-insensitive DAIMS or Terse 30 labels.
* Updated header error reports to specify DAIMS labels.
* Allowed FABS users to delete existing or public records without the normally required fields.
* Dropped the deprecated ExecutiveCompensation table (the DUNS table now includes this data).
* Removed dependency on the unified agency list's "FPDS Department ID" column.
* Updated SQL validations so checks on Action Type are case-insensitive.

PR #1642, opened 2019-07-19
DEV-1761: list submissions min last mod

alburde1:
**High level description:**
Include the minimum last modified date in the `list_submissions` endpoint

**Technical details:**
Filters should not affect the minimum date, only the type (FABS/DABS) and the publish status (true/false/both) along with user permissions

**Link to JIRA Ticket:**
[DEV-1761](https://federal-spending-transparency.atlassian.net/browse/DEV-1761)

### DIFF
--- a/dataactbroker/handlers/fileHandler.py
+++ b/dataactbroker/handlers/fileHandler.py
@@ -1569,7 +1569,7 @@ def list_submissions(page, limit, certified, sort='modified', order='desc', is_f
     return JsonResponse.create(StatusCode.OK, {
         'submissions': [serialize_submission(submission) for submission in query],
         'total': total_submissions,
-        'min_last_modified': str(min_last_mod)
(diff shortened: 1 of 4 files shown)


dataactbroker/handlers/fileHandler.py
```diff
@@ -1557,12 +1562,14 @@ def list_submissions(page, limit, certified, sort='modified', order='desc', is_f
     query = query.order_by(sort_order)
 
     total_submissions = query.count()
+    min_last_mod = min_mod_query.one()[0]
```
dpb-bah: While there's basically no chance this could return zero, could this use `first()` or `one_or_none()`
and the response include:
`'min_last_modified': str(min_last_mod[0]) if min_last_mod else None`.
Just a personal preference, you're welcome to ignore.

dpb-bah: Disregarding, `min()` returns `None` even if there are no results.

PR #1656, opened 2019-08-09
Sprint 87: Staging -> Master

alburde1:


dpb-bah:
* Updated FPDS loader to store "Additional Reporting" records in preparation for future DAIMS work.
* Updated the accuracy of file upload row counts, excluding the header and extra carriage returns.
* Improved the performance of cross-file rules A18 and A19.
* Updated Federal Hierarchy loader to include metrics on new cgacs, subtiers to be imported in order to update the unified agency list shared with USAspending.gov.
* Added functionality to generate pipe-delimited text files for D1 and D2.

PR #1678, opened 2019-09-09
Sprint 90: Dev -> Staging

alburde1:


dpb-bah:
* Added endpoint (`get_comments_file`) which generates a file containing all the comments related to a submission.
* Updated the Subaward loader to link subawards to awards, specifically taking into account dashes in the award ids.

PR #1701, opened 2019-10-08
Sprint 92: Dev -> Staging

alburde1:


dpb-bah:
* Added "/v1/historic_dabs_summary" endpoint which surfaces submission data filtered by agencies and the fiscal periods. This will be used by the Agency Data dashboard soon to come.
* Updated permissions documentation with details on each role's available functionality.

PR #1710, opened 2019-10-22
Sprint 93: Dev -> Staging

alburde1:


dpb-bah:
* Updated agency selectors on DABS pages to not include unaffiliated subtiers.
* Added endpoint "GET /v1/latest_certification_period/" which returns the latest quarter/fy of certifying DABS submissions.
* Added endpoint "POST /v1/get_rule_labels/" which returns a list of filtered rule labels.
* Added several endpoints to be used for the agency data dashboard (for trend analysis and listing error metadata)
* Users with only Edit-FABS permissions may now upload and delete but not publish submissions using the front-end interface.

PR #1713, opened 2019-10-29
DEV-3455: Unique ID File A - Display TAS

dpb-bah:
**High level description:**
Unique ID architecture with file A along with setting up display TAS values for staging A/B/C tables and the SF-133 table.

**Technical details:**
- Includes alembic migrations to staging tables and SF133 table.

**Link to JIRA Ticket:**
[DEV-3455](https://federal-spending-transparency.atlassian.net/browse/DEV-3455)

- [ ] Double check manual tests making sure everything works the way it should


dataactcore/migrations/versions/5e6d944e37be_adding_display_tas_to_sf133.py
```diff
@@ -0,0 +1,40 @@
+"""Adding display_tas to SF133
```
alburde1: Is there a reason this had to be a separate migration from the other one?

dataactcore/models/domainModels.py
```diff
@@ -33,6 +38,23 @@ def concat_tas_dict(tas_dict):
     return tas
 
 
+def concat_display_tas_dict(tas_dict):
+    """ Given a dictionary, create a concatenated display TAS string. Copied directly from USASpending.gov. """
+    tas_rendering_label = "-".join(filter(None, (tas_dict['allocation_transfer_agency'],
+                                                 tas_dict['agency_identifier'])))
+
+    typecode = tas_dict['availability_type_code']
+    if typecode is not None and typecode != "":
```
alburde1: Can we just do `if typecode:`? I think it'll cover all these bases

dataactcore/models/domainModels.py
```diff
@@ -33,6 +38,23 @@ def concat_tas_dict(tas_dict):
     return tas
 
 
+def concat_display_tas_dict(tas_dict):
+    """ Given a dictionary, create a concatenated display TAS string. Copied directly from USASpending.gov. """
```
alburde1: Format

dataactvalidator/config/sqlrules/a16_appropriations.sql
```diff
@@ -2,12 +2,15 @@
 WITH appropriation_a16_{0} AS
     (SELECT submission_id,
         row_number,
-        budget_authority_unobligat_fyb
+        budget_authority_unobligat_fyb,
+        tas,
```
alburde1: Why are we pulling `tas` in here? We don't use it

tests/integration/validatorTests.py
```diff
@@ -134,12 +134,13 @@ def test_schema_optional_field(self):
             "test3": "YES",
             "test4": "1",
             "test5": "1",
+            "afa_generated_unique": ""
```
alburde1: Why did we make these double quotes? I thought we were moving away from those

tests/unit/dataactvalidator/test_a34_appropriations.py
```diff
@@ -8,8 +8,8 @@
 
 
 def test_column_headers(database):
-    expected_subset = {'row_number', 'budget_authority_unobligat_fyb', 'expected_value_GTAS SF133 Line 2490',
-                       'difference'}
+    expected_subset = {'uniqueid_TAS', 'row_number', 'budget_authority_unobligat_fyb',
+                       'expected_value_GTAS SF133 Line 2490', 'difference'}
     actual = set(query_columns(_FILE, database))
     assert (actual & expected_subset) == expected_subset
```
alburde1: Is there a reason to leave any of the tests that use these bitwise operators as having them? They seem pretty redundant if we're already making sure it's `==`. We can probably just do `actual == expected_subset` for all of them.

Since I can't comment on all the spots that have this, here's a quick list for easy access: A32, A33.2, A34, A36

tests/unit/dataactvalidator/test_scripts_load_sf133.py
```diff
@@ -11,15 +11,15 @@
...
+    'period', 'created_at', 'updated_at', 'tas', 'display_tas']
 
 
 def test_fill_blank_sf133_lines_types():
     """Validate that floats aren't downgraded to ints in the pivot_table
     function (that'd be a regression)."""
     data = pd.DataFrame(
         # We'll only pay attention to two of these fields
-        [[1440, 3041046.31] + list('ABCDEFGHIJKL')], columns=['line', 'amount'] + FINGERPRINT_COLS
+        [[1440, 3041046.31] + list('ABCDEFGHIJKLL')], columns=['line', 'amount'] + FINGERPRINT_COLS
```
alburde1: Just checking, why did we add a duplicate letter at the end here? (and duplicate numbers lower down)

tests/unit/dataactvalidator/test_scripts_load_sf133.py
```diff
@@ -11,15 +11,15 @@
 FINGERPRINT_COLS = [
     'availability_type_code', 'sub_account_code', 'allocation_transfer_agency', 'fiscal_year',
     'beginning_period_of_availa', 'ending_period_of_availabil', 'main_account_code', 'agency_identifier',
-    'period', 'created_at', 'updated_at', 'tas']
+    'period', 'created_at', 'updated_at', 'tas', 'display_tas']
 
 
 def test_fill_blank_sf133_lines_types():
     """Validate that floats aren't downgraded to ints in the pivot_table
     function (that'd be a regression)."""
```
alburde1: We're already here, we might as well clean up the comments for these functions

tests/unit/dataactvalidator/validation_handlers/test_file_generation_manager.py
```diff
@@ -23,6 +23,113 @@
...
+        'beginning_period_of_availa': None,
+        'ending_period_of_availabil': None,
+        'availability_type_code': None,
+        'main_account_code': '0001',
+        'sub_account_code': '001'
+    }
+    tas2_str = concat_tas_dict(tas_dict)
+    tas2_dstr = concat_display_tas_dict(tas_dict)
+    assert tas2_str == '09701700000000 0001001'
+    assert tas2_dstr == '097-017-0001-001'
```
alburde1: Can we get a test showing what happens if we're missing just one of the dates? I know it isn't valid but during validation it's possible we'll have one like that, which is something the validations would catch but we'll still have to display it in the error/warning reports

dpb-bah:
- Reloaded the latest SF133 with my local DB, ensured the display_tas was correct.
- Going through the certification process.
- Good catch, we most definitely want to backfill the SF133. I can come up with a script for that (though it would have to backfill... 17M rows.

dataactcore/migrations/versions/5e6d944e37be_adding_display_tas_to_sf133.py
dpb-bah: No, it was done after the fact. I can merge them.

tests/unit/dataactvalidator/test_scripts_load_sf133.py
dpb-bah: This test was setup weirdly with the `FINGERPRINT` setup. Long story short, the L here (and the duplicate numbers lower down) represent the TAS string for each row. So to make it account for Display TAS, I simply copied the same letter/number to be consistent.

dataactcore/models/domainModels.py
dpb-bah: Agreed.

dataactvalidator/config/sqlrules/a16_appropriations.sql
dpb-bah: Most of the time the TAS was used to join as well. Good catch.

tests/integration/validatorTests.py
dpb-bah: I didn't want to change the rest of the file. I can fix it.

PR #1737, opened 2019-12-09
DEV-1720 fabs ppop scope

dpb-bah:
**High level description:**
Adding and deriving PrimaryPlaceOfPerformanceScope to `PublishedAwardFinancialAssistance`

**Technical details:**
- Alembic migration to add column to FABS

**Link to JIRA Ticket:**
[DEV-1720](https://federal-spending-transparency.atlassian.net/browse/DEV-1720)


dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -188,11 +197,19 @@ def derive_ppop_location_data(obj, sess, ppop_code, ppop_state_code, county_dict
             obj['place_of_performance_city'] = city_info.feature_name
             obj['place_of_perform_county_co'] = city_info.county_number
             obj['place_of_perform_county_na'] = city_info.county_name
+        elif re.match('^[A-Z]{2}\*{5}$', ppop_code):
+            obj['place_of_performance_scope'] = "State-wide"
+        elif re.match('^0{2}\*{5}$', ppop_code):
```
alburde1: This is actually more complicated/longer than just saying `'^00\*{5}$'`

dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -170,6 +170,14 @@ def derive_ppop_location_data(obj, sess, ppop_code, ppop_state_code, county_dict
         # deriving PrimaryPlaceOfPerformanceCityName
         city_info = sess.query(ZipCity).filter_by(zip_code=zip_five).one()
         obj['place_of_performance_city'] = city_info.city_name
+
+        # deriving PrimaryPlaceOfPerformanceScope
+        if ppop_code and (re.match('^[A-Z]{2}\d{5}$', ppop_code) or re.match('^[A-Z]{2}\d{4}R$', ppop_code)):
+            obj['place_of_performance_scope'] = "Single ZIP Code"
+    # if there is zip4 and it *is* city-wide, set the scope
+    elif obj['place_of_performance_zip4a'] and \
```
alburde1: Why don't we check for the existence of `ppop_code` here? We do in the above and it's just as likely to not exist here as it is up there. In fact, we could put this inside the `elif ppop_code` since that's required for this check anyway

dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -564,6 +581,7 @@ def fabs_derivations(obj, sess, state_dict, country_dict, sub_tier_dict, cfda_di
     obj['legal_entity_city_name'] = None
     obj['place_of_performance_zip5'] = None
     obj['place_of_perform_zip_last4'] = None
+    obj['place_of_performance_scope'] = None
```
alburde1: This doesn't have to be initialized, we never check it against anything. The initialization here is in case it somehow doesn't exist before we start comparing it against other things.

dataactcore/migrations/versions/ee70a910db2e_adding_primaryplaceofperformancescope_.py
```diff
@@ -0,0 +1,41 @@
...
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+
+
+
+def upgrade_data_broker():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('published_award_financial_assistance', sa.Column('place_of_performance_scope', sa.Text(),
```
alburde1: While we're already editing, just a general thing, we try to keep most of these on the same line (linting need not apply to migrations)

dpb-bah: TIL linting doesnt apply to migrations 👍

dataactbroker/helpers/fabs_derivations_helper.py
dpb-bah: So should I not include that check in the tests too?

alburde1: You set it to `None` in the derivations don't you?

dpb-bah: Not in every case. I can update it to do that.

dpb-bah: Not for every case. I can definitely do that though.

PR #1745, opened 2019-12-17
Sprint 97 Dev -> Staging

alburde1:


dpb-bah:
* Added "PrimaryPlaceOfPerformanceScope" derivation for FABS records and script to backfill it. Note, this is not currently included in D2 files.
* Updated D1/D2 filenames to include the date range requested.
* Updated subaward pipeline to exclude aggregate records when sending to FSRS.
* Removed "submission_narrative" endpoints. Please use "submission_comment" endpoints instead displayed in the README.
* Updated API to make the trailing slash for each endpoint to be optional.  
* Resolved various minor bugs.

PR #1751, opened 2020-01-02
OPS-1036 remove debug log level for supervisor

kwhickey:
**High level description:**
Reduce verbosity of supervisord.log by not duplicating app logs in it.

**Technical details:**
It turns out that having this set at debug will duplicate any console logging output from the running python processes to the supervisor logs

- see: debug puts process output into the normal log: http://supervisord.org/logging.html#activity-log-levels
- see: about how child process stderr and stdout gets logged when defaulted to AUTO: http://supervisord.org/logging.html#child-process-logs
- see that the defaults when not set puts it in the /tmp dir with 50MB rotated files up to a max of 10 files: http://supervisord.org/configuration.html

**Link to JIRA Ticket:**
[OPS-1036](https://federal-spending-transparency.atlassian.net/browse/DEV-1036)

- [NA] Backend review completed
- [NA] Merged concurrently with [Frontend#1234](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/1234)

### DIFF
--- a/dataactbroker/config/supervisord.conf
+++ b/dataactbroker/config/supervisord.conf
@@ -21,5 +21,4 @@ startsecs=1
 startretries=10
 
 [supervisord]
-loglevel=debug
 environment=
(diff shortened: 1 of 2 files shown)


kwhickey:
@myazidi or @egnoriega - can you take a look at this and approve? RE: limiting log output.

PR #1782, opened 2020-02-11
Sprint 100: Dev -> Staging

alburde1:


dpb-bah:
* Refactored the data loading process used in multiple features, speeding up the performance two to three times.
* Fixed FABS 2.2.1 error text for accuracy.
* Updated FABS derivations with record type "3" to account for "Place of Performance State Code".

PR #1786, opened 2020-02-13
DEV-4456: PPOP Scope Fix

dpb-bah:
**High level description:**
* Resolving bug with ppop scope derivation by moving it after most of the derivations.
* Adding case where ppop zip is blank and the code is XX##### or XX####R -> 'City-wide'

**Technical details:**
* Updating backfill ppop script which should be run again to account for the missed derivations.

**Link to JIRA Ticket:**
[DEV-4456](https://federal-spending-transparency.atlassian.net/browse/DEV-4456)

### DIFF
--- a/dataactbroker/helpers/fabs_derivations_helper.py
+++ b/dataactbroker/helpers/fabs_derivations_helper.py
@@ -171,11 +171,6 @@ def derive_ppop_location_data(obj, sess, ppop_code, ppop_state_code, county_dict
         city_info = sess.query(ZipCity).filter_by(zip_code=zip_five).one()
         obj['place_of_performance_city'] = city_info.city_name
 
-        # deriving PrimaryPlaceOfPerformanceScope
-        if ppop_code and (re.match('^[A-Z]{2}\d{5}$', ppop_code) or re.match('^[A-Z]{2}\d{4}R$', ppop_code)):
-            obj['place_of_performance_scope'] = "Single ZIP Code"
-        else:
-            obj['place_of_performance_scope'] = None
     # if there is no valid ppop zip4, we need to try to derive county/city info from the ppop code
     elif ppop_code:
         # if ppop_code is in county format,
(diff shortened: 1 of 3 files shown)


dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -196,30 +190,46 @@ def derive_ppop_location_data(obj, sess, ppop_code, ppop_state_code, county_dict
...
+
+
+def derive_ppop_scope(obj):
+    """ Deriving place of performance scope values from zip4 and place of performance code
+
+        Args:
+            obj: a dictionary containing the details we need to derive from and to
+    """
+    derived_ppop_scope = None
+    ppop_code = obj['place_of_performance_code'].upper() if obj['place_of_performance_code'] else None
```
alburde1: Is there a reason we're not just passing the one we already get?

dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -196,30 +190,46 @@ def derive_ppop_location_data(obj, sess, ppop_code, ppop_state_code, county_dict
...
-        obj['place_of_performance_scope'] = None
+
+
+def derive_ppop_scope(obj):
+    """ Deriving place of performance scope values from zip4 and place of performance code
+
+        Args:
+            obj: a dictionary containing the details we need to derive from and to
+    """
+    derived_ppop_scope = None
```
alburde1: Why are we making this a variable instead of just setting it?

dpb-bah: For me it's easier to see `derived_ppop_scope =` multiple times instead of `obj['place_of_performance_code'] =` multiple times when really the function is only calling it once.

dataactbroker/helpers/fabs_derivations_helper.py
dpb-bah: Trying to make this function more adaptable, meaning accounting for cases where we can't assume we've upper'd it beforehand.

alburde1: What cases are these?

dataactbroker/helpers/fabs_derivations_helper.py
alburde1: Is this the direction we're going with this file from now on? All other instances in it where there are multiple options simply set it.

dataactbroker/helpers/fabs_derivations_helper.py
dpb-bah: If someone were to call the function or if someone were to copy said function without knowing they were upper'd beforehand.

dataactbroker/helpers/fabs_derivations_helper.py
dpb-bah: It's something I'd prefer but it doesn't have to be one or the other.

PR #1788, opened 2020-02-13
DEV-3849: date format check

alburde1:
**High level description:**
Check format of GeneralLedgerPostDate

**Technical details:**
- Adding a date format checker
- Adding a new error type

**Link to JIRA Ticket:**
[DEV-3849](https://federal-spending-transparency.atlassian.net/browse/DEV-3849)


dataactbroker/helpers/validation_helper.py
```diff
@@ -427,6 +450,49 @@ def check_length(data, length_fields, report_headers, csv_schema, short_cols, fl
     return errors
 
 
+def check_field_format(data, format_fields, report_headers, short_cols, flex_data):
+    """ Check if all fields that are a type other than string match that type.
```
dpb-bah: Can we update this description to something like "Check if all the fields are in the appropriate format"? Similarly, with the other point, is this going to check the formatting of all types or just dates (considering the error message, rule label, and expected value are pretty geared towards the date format). If so, can we rename to `check_date_field_format` and the description to "Check if all date fields are in the appropriate format"?

dataactbroker/helpers/validation_helper.py
```diff
@@ -265,6 +267,27 @@ def expected_length(row, csv_schema):
     return 'Max length: {}'.format(current_field.length)
 
 
+def valid_format(row):
```
dpb-bah: This is pretty confined to just date formatting. For future reference, if there are other strings/types to format, is the plan to add the new type to the `parsed_fields['format']` and then in this function chooses the appropriate format for the given type? Otherwise, if this is just going to be used for `date`s, then I'd ask to update the name to `valid_date_format`.

PR #1795, opened 2020-02-21
Sprint 101: Dev -> Staging

alburde1:


dpb-bah:
* Updated FABS derivation for place_of_performance scope for accuracy.
* Adding `get_impact_counts` and `get_significance_counts` for future functionality of the active dabs dashboard.
* Reworked RuleSettings table to allow reloading of the SQL rules.
* Dropped the constraint on users creating submissions for a period that is already certified. Users will still be unable to certify for these duplicate submissions.

PR #1800, opened 2020-03-04
DEV-4617: submission metadata update

alburde1:
**High level description:**
Sending the ID of the certified submission rather than just indicating whether there is one or not

**Technical details:**
Note: README updates are already in another ticket so those will remain there to simplify things

**Link to JIRA Ticket:**
[DEV-4617](https://federal-spending-transparency.atlassian.net/browse/DEV-4617)

### DIFF
--- a/dataactbroker/handlers/submission_handler.py
+++ b/dataactbroker/handlers/submission_handler.py
@@ -1,4 +1,5 @@
 import logging
+import json
 
 from datetime import datetime
 from flask import g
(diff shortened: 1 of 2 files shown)


dataactbroker/handlers/submission_handler.py
```diff
@@ -136,6 +137,12 @@ def get_submission_metadata(submission):
     test_sub = find_existing_submissions_in_period(submission.cgac_code, submission.frec_code,
```
dpb-bah: Would it make more sense to refactor `find_existing_submissions_in_period` to return the list of submissions (`[]` if not found) and then have another function to JSONify the response for `check_year_and_quarter`? Totally a matter of preference.

PR #1804, opened 2020-03-09
Sprint 102: Dev -> Staging

alburde1:


dpb-bah:
* Created a "revert_submission" endpoint which will allow users to revert an updated DABS submission to its last certified state.
* Created and backfilled CertifiedAwardProcurement and CertifiedAwardFinancialAssistance tables to be used for the "revert_submission" endpoint.
* Updated README instructions for users installing the Broker.

PR #1818, opened 2020-03-13
DEV-4681: FABS 38.1 Update (including mixed aggregates 'E')

dpb-bah:
**High level description:**
Updating FABS 38.1 to include mixed aggregate records.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-4681](https://federal-spending-transparency.atlassian.net/browse/DEV-4681)

- Documentation Updated


tests/unit/dataactvalidator/test_fabs38_detached_award_financial_assistance_1.py
```diff
@@ -31,20 +32,29 @@ def test_success(database):
                                                           action_date='10/01/2018', correction_delete_indicatr='')
 
     # Rule ignored for CorrectionDeleteIndicator of D
-    det_award_5 = DetachedAwardFinancialAssistanceFactory(funding_office_code=None, action_type='A',
+    det_award_5 = DetachedAwardFinancialAssistanceFactory(funding_office_code=None, action_type='E',
                                                           action_date='10/01/2018', correction_delete_indicatr='D')
-    errors = number_of_errors(_FILE, database, models=[det_award_1, det_award_2, det_award_3, det_award_4, det_award_5])
+
+    # Rule ignored for action_type of None
```
alburde1: Should this section be checking not only `None` but also some other action types just to make sure it doesn't shoot off for other things?

Also, I can't comment that high up because GitHub but there are a couple of tests (`det_award_3` and `det_award_2`) that are using action type of `None` which means those tests are being ignored, they should be updated as well

dpb-bah: > Should this section be checking not only None but also some other action types just to make sure it doesn't shoot off for other things?
there's another section for "# Rule ignored for other action types"

Updated the None's above.

PR #1821, opened 2020-03-13
DEV-4672: B20 Update (adding program activity name)

dpb-bah:
**High level description:**
Adding program activity name to the main check to B20.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-4672](https://federal-spending-transparency.atlassian.net/browse/DEV-4672)

- Documentation Updated


dataactvalidator/config/sqlrules/b20_object_class_program_activity.sql
```diff
@@ -37,6 +44,9 @@ WHERE NOT EXISTS (
                 OR COALESCE(af.program_activity_code, '') = ''
                 OR af.program_activity_code = '0000'
             )
+            AND (UPPER(COALESCE(af.program_activity_name, '')) = UPPER(COALESCE(op.program_activity_name, ''))
```
alburde1: program activity name is not indexed in our tables. We may want to change that considering we'll be running this rule on it. I think we can just tweak the migration we already have, no need for a new one, to also index `UPPER(program_activity_name)`

tests/unit/dataactvalidator/test_b20_object_class_program_activity.py
```diff
@@ -15,42 +15,69 @@ def test_column_headers(database):
```
alburde1: I can't actually comment higher than this, but the `&` doesn't have to be there. You can just do `actual == expected_subset` with the same results (actually more accurate). Because this doesn't include all the columns we should be receiving back as-is

PR #1823, opened 2020-03-19
DEV-4080 Fixing Subgrant Linking Logic

dpb-bah:
**High level description:**
Updating the FSRS Grant linking logic to be fully accurate with FABS.  

**Technical details:**
* Matching case-insensitive
  * fain (dashes removed) <-> fain
  * federal_agency_id <-> awarding_sub_tier_agency_c 
* Alembic migration for a couple of indexes

**Link to JIRA Ticket:**
[DEV-4080](https://federal-spending-transparency.atlassian.net/browse/DEV-4080)

- [x] Reloaded Subgrants on Sandbox
- Documentation Updated


dataactcore/models/stagingModels.py
```diff
@@ -1694,6 +1694,7 @@ def __init__(self, **kwargs):
 Index("ix_pafa_uri_upper", func.upper(PublishedAwardFinancialAssistance.uri))
 Index("ix_pafa_awarding_subtier_c_upper", func.upper(PublishedAwardFinancialAssistance.awarding_sub_tier_agency_c))
 Index("ix_pafa_afa_generated_unique_upper", func.upper(PublishedAwardFinancialAssistance.afa_generated_unique))
+Index("ix_pafa_record_type", func.upper(PublishedAwardFinancialAssistance.record_type))
```
alburde1: 1) Why are we uppering the record type? Why not just set `index=True` in the column definition?
2) It is ALREADY defined that way and this index already exists

dataactcore/scripts/raw_sql/link_broken_subaward_grants.sql
```diff
@@ -34,7 +33,9 @@ aw_pafa AS
         AND EXISTS (
             SELECT 1
             FROM unlinked_subs
-            WHERE UPPER(TRANSLATE(unlinked_subs.award_id, '-', '')) = UPPER(TRANSLATE(pafa.fain, '-', ''))
+            WHERE pafa.record_type != 1
+                AND UPPER(TRANSLATE(unlinked_subs.award_id, '-', '')) = UPPER(TRANSLATE(pafa.fain, '-', ''))
+                AND UPPER(unlinked_subs.awarding_sub_tier_agency_c) IS NOT DISTINCT FROM UPPER(pafa.awarding_sub_tier_agency_c)
```
alburde1: Have you run any tests to see if `IS NOT DISTINCT FROM` is faster or slower than `COALESCE`?

This comment applies to all instances of `IS NOT DISTINCT FROM`

dataactcore/scripts/raw_sql/link_broken_subaward_grants.sql
```diff
@@ -34,7 +33,9 @@ aw_pafa AS
         AND EXISTS (
             SELECT 1
             FROM unlinked_subs
-            WHERE UPPER(TRANSLATE(unlinked_subs.award_id, '-', '')) = UPPER(TRANSLATE(pafa.fain, '-', ''))
+            WHERE pafa.record_type != 1
```
alburde1: Should be using `<>` not `!=`

dataactcore/scripts/raw_sql/populate_subaward_table_grants.sql
```diff
@@ -25,8 +22,10 @@ WITH aw_pafa AS
         AND EXISTS (
             SELECT 1
             FROM fsrs_grant
-            WHERE UPPER(TRANSLATE(fsrs_grant.fain, '-', '')) = UPPER(TRANSLATE(pafa.fain, '-', ''))
+            WHERE record_type != 1
```
alburde1: Should be using `<>` not `!=`

dataactcore/models/stagingModels.py
dpb-bah: Force of habit of UPPERCASE'ing everything and confirmed there are no letter record_types in the data. Removing.

dataactcore/scripts/raw_sql/link_broken_subaward_grants.sql
dpb-bah: Both are very similar in time with `IS NOT DISTINCT FROM` being just a little faster.
-- 6 rows, 509.5s IS NOT DISTINCT
-- 6 rows, 551.1s COALESCE

PR #1824, opened 2020-03-20
Sprint 102: Staging -> Master

alburde1:


dpb-bah:
* Created a "revert_submission" endpoint which will allow users to revert an updated DABS submission to its last certified state.
* Created and backfilled CertifiedAwardProcurement and CertifiedAwardFinancialAssistance tables to be used for the "revert_submission" endpoint.
* Updated README instructions for users installing the Broker.
* Users may be able to create test submissions for the same agency and the same quarter despite the certified submission already existing. Note one cannot certify these test submissions.

PR #1826, opened 2020-03-24
Sprint 103 Staging Deploy

dpb-bah:


dpb-bah:
* Upgraded python library (psutil) for security updates.
* Reworked linking between FSRS grants and FABS records for accuracy.
* Added functionality to revert already certified and updated submissions back to when they was last certified, allowing users to further test submissions in a safe scenario.

PR #1834, opened 2020-04-07
Sprint 104: Dev -> QAT

alburde1:


dpb-bah:
* Dropped outdated fpds_contracting_offices table.
* Added a "Reverting" publish status to prevent users from interrupting the reverting process.
* Updated older FABS records with outdated business types ('00', '01', etc.)
* Updated FABS rules and derivations for data quality (FABS38, "legal_entity_congressional" when record type is "1")

PR #1841, opened 2020-04-10
Fix: Updating Joins for Active DABS Dashboard Endpoints

dpb-bah:
**High level description:**
It was discovered that after reworking the RuleSettings table's unique identifiers, some of the endpoints still used the old way of joining, causing discrepancies with rules that have the same label (`B9`). This resolves that.

**Technical details:**
N/A

**Link to JIRA Ticket:**
N/A

- Documentation Updated


tests/unit/dataactbroker/test_dashboard_handler.py
```diff
@@ -193,7 +194,11 @@ def setup_submissions(sess, admin=False):
     rule_ab2 = RuleSql(rule_sql='', rule_label='B1', rule_error_message='second cross rule', query_name='',
                        file_id=FILE_TYPE_DICT_LETTER_ID['B'], target_file_id=FILE_TYPE_DICT_LETTER_ID['A'],
                        rule_severity_id=RULE_SEVERITY_DICT['warning'], rule_cross_file_flag=True, category='existence')
-    db_objects.extend([rule_a1, rule_a2, rule_ab1, rule_ab2])
+    # Adding a rule with the same rule label as another but different file
+    rule_cd1 = RuleSql(rule_sql='', rule_label='B1', rule_error_message='extra cross rule', query_name='',
```
alburde1, line 198: None of our existing "same label" rules are for the same job. One is for file B and the other is for file C. Should we be testing with that in mind?

tests/unit/dataactbroker/test_dashboard_handler.py
```diff
@@ -208,6 +213,9 @@ def setup_submissions(sess, admin=False):
     sub1_ab2 = ErrorMetadata(job=sub1_ab, original_rule_label=rule_ab2.rule_label, occurrences=130,
                              file_type_id=rule_ab2.file_id, target_file_type_id=rule_ab2.target_file_id,
                              rule_failed=rule_ab2.rule_error_message, severity_id=rule_ab2.rule_severity_id)
+    sub1_cd1 = ErrorMetadata(job=sub1_cd1, original_rule_label=rule_cd1.rule_label, occurrences=10,
```
alburde1, line 216: This has the same name as the new job (which shouldn't be an issue once the comment on the job is handled but is probably still bad practice)

PR #1850, opened 2020-04-21
Sprint 105: Dev -> QAT

alburde1:


dpb-bah:
* Updating DABS banner to inform the user of which state their submission is. This banner also allows users to revert their submission after it has been certified and updated on any step of the submission workflow.
* Updated "list_agencies" endpoint to filter out agencies based on the user's affiliations.
* Updated settings endpoints only to be accessible by DABS submitters/certifiers.

PR #1886, opened 2020-05-21
DEV-5131: Updating C5.2 Triggers

dpb-bah:
**High level description:**
Updating C5.2 to only trigger when:
* gross_outlay_amount_by_awa_cpe is not null
* any of the optional values is non-zero

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-5131](https://federal-spending-transparency.atlassian.net/browse/DEV-5131)

- Documentation Updated

### DIFF
--- a/dataactvalidator/config/sqlrules/c5_award_financial_2.sql
+++ b/dataactvalidator/config/sqlrules/c5_award_financial_2.sql
@@ -18,10 +18,10 @@ SELECT
 FROM award_financial
 WHERE submission_id = {0}
     AND gross_outlay_amount_by_awa_cpe IS NOT NULL
-    AND (COALESCE(gross_outlays_undelivered_cpe, 0) > 0
-         OR COALESCE(gross_outlays_undelivered_fyb, 0) > 0
-         OR COALESCE(gross_outlays_delivered_or_cpe, 0) > 0
-         OR COALESCE(gross_outlays_delivered_or_fyb, 0) > 0
+    AND (COALESCE(gross_outlays_undelivered_cpe, 0) <> 0
+         OR COALESCE(gross_outlays_undelivered_fyb, 0) <> 0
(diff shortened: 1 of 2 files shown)


dataactvalidator/config/sqlrules/c5_award_financial_2.sql
```diff
@@ -17,6 +17,12 @@ SELECT
     uri AS "uniqueid_URI"
 FROM award_financial
 WHERE submission_id = {0}
+    AND gross_outlay_amount_by_awa_cpe IS NOT NULL
```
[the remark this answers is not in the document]
dpb-bah: This was discussed with @schichester .  The wording of this may need updated in a future release but `GrossOutlayAmountByAward_CPE` *can* be zero and still fine. Sending the discussion over slack.

dataactvalidator/config/sqlrules/c5_award_financial_2.sql
```diff
@@ -17,6 +17,12 @@ SELECT
...
+    AND (COALESCE(gross_outlays_undelivered_cpe, 0) > 0
+         OR COALESCE(gross_outlays_undelivered_fyb, 0) > 0
+         OR COALESCE(gross_outlays_delivered_or_cpe, 0) > 0
+         OR COALESCE(gross_outlays_delivered_or_fyb, 0) > 0
+    )
```
alburde1, line 25: Would these always be checking for greater than zero? Do we just want to check for <> 0 to account for possible negatives? (not sure if that's something that should ever happen, probably not, but the situation should probably be accounted for)

dpb-bah: Good question, I don't think they should have negative values but we certainly should account for them.

PR #1922, opened 2020-07-21
[HOTFIX][DEV-5600] Optimize CSV data processing with vectorized use of Pandas

kwhickey:
**High level description:**
Addressing bottlenecks in ingest and processing of (very large) CSV files using Pandas data processing framework. Swapping use of code doing element-wise function calls (on each cell) with vectorized calls across column Series to reduce processing time.

**Technical details:**
Pandas is meant to be used to best effect when vectorizing functions across Series. Pandas DataFrames can be thought of as a matrix of Series, which are just numpy arrays, as rows/columns. As numpy arrays, you can apply a function in a vectored way to have it take effect on the whole array at once, rather than iterating cells in that Series/array.
We seemed to be following a pattern of applying validation functions element-wise. Handing a function to pandas that we then apply row by row, cell by cell. Things that could be validated one whole column at a time.

Main validations vectorized, in new functions, to biggest effect in this PR:
1. `clean_numbers`
2. `clean_cols`
3. `check_type`
4. `check_length` 

Additional unit tests and integration tests were added to verify corner cases with `None`, `numpy.NaN`, and mixed-type DataFrames.

**Link to JIRA Ticket:**
[DEV-5600](https://federal-spending-transparency.atlassian.net/browse/DEV-1234)

- [NA] Merged concurrently with [Frontend#1234](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/1234)
- [NA] Documentation Updated

### DIFF
--- a/dataactbroker/helpers/validation_helper.py
+++ b/dataactbroker/helpers/validation_helper.py
@@ -25,8 +25,8 @@ def is_valid_type(data, data_type):
     if data_type is None:
         # If no type specified, don't need to check anything
         return True
-    if data.strip() == '':
-        # An empty string matches all types
+    if data is None or data.strip() == '':
+        # All types can be None, and an empty string matches all types
         return True
     if data_type == 'STRING':
         return len(data) > 0
--- a/dataactbroker/helpers/validation_helper.py
+++ b/dataactbroker/helpers/validation_helper.py
@@ -373,17 +373,21 @@ def check_type(data, type_fields, type_labels, report_headers, csv_schema, short
             A dataframe containing error text that can be turned into an error report for non-string fields
     """
     # Get just the non-string columns along with the row number and unique ID
-    type_data = data[type_fields + ['row_number', 'unique_id']]
+    invalid_datatype = data[type_fields + ['row_number', 'unique_id']]
+
+    for type_field in type_fields:
+        # For each col-Series, null-out (set to NaN) any cells that meet datatype requirements
+        required_type = FIELD_TYPE_DICT_ID[csv_schema[type_field].field_types_id]
+        invalid_datatype[type_field].where(
+            invalid_datatype[type_field].map(lambda x: is_valid_type(x, required_type)), inplace=True
+        )
+
     # Flip the data so each header + cell combination is its own row, keeping the relevant row numbers and unique IDs
-    errors = pd.melt(type_data, id_vars=['row_number', 'unique_id'], value_vars=type_fields,
+    errors = pd.melt(invalid_datatype, id_vars=['row_number', 'unique_id'], value_vars=type_fields,
                      var_name='Field Name', value_name='Value Provided')
-    # Throw out all rows that don't have data, they don't have a type
-    errors = errors[~errors['Value Provided'].isnull()]
-    # If there is data that needs checking, keep only the data that doesn't have the right type
-    if not errors.empty:
-        errors['matches_type'] = errors.apply(lambda x: valid_type(x, csv_schema), axis=1)
-        errors = errors[~errors['matches_type']]
-        errors.drop(['matches_type'], axis=1, inplace=True)
+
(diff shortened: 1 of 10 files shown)


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -11,9 +11,10 @@
```
kwhickey, line 11: Best to look at this file with the diff settings above set to ignore whitespace (changed python indent)

dataactbroker/helpers/validation_helper.py
```diff
@@ -80,6 +81,41 @@ def clean_col(value, clean_quotes=True):
...
+            frame: pd.DataFrame to clean
+            convert_to_str: whether to apply .astype(str) on all Series in the DataFrame
+                - If the DataFrame was created or read in from CSV with dtype=str already, this may be unnecessary
+                  and will save on performance if avoided
+                - If left as False and there are mixed types, such as int or float, they will be set to None rather
+                  than converted to their str form
+            clean_quotes: whether to clean extra quotes or not
+
+        Returns:
+            The cleaned DataFrame, with a side-effect of an
```
tony-sappe, line 99: "The cleaned DataFrame, with a side-effect of an"

What.... What is the side-effect!?!??!

kwhickey: ha... good catch. Updated docstrings.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -24,7 +27,8 @@
     populate_job_error_info, get_action_dates
 )
 
-from dataactcore.models.domainModels import matching_cars_subquery, Office, concat_tas_dict, concat_display_tas_dict
+from dataactcore.models.domainModels import concat_tas_dict_vectorized
+from dataactcore.models.domainModels import Office, concat_display_tas_dict
```
dpb-bah, line 31: merge?

dataactbroker/helpers/validation_helper.py
```diff
@@ -126,6 +209,23 @@ def derive_unique_id(row, is_fabs):
...
+            is_fabs: a boolean indicating if the submission is a FABS submission or not
+
+        Returns:
+            A Series with properly formatted unique ID all rows depending on if it's a FABS or DABS submission
+    """
+    if not is_fabs:
+        return 'TAS: ' + frame['display_tas'][frame['display_tas'].notnull()].astype(str)
+    else:
+        return 'AssistanceTransactionUniqueKey: ' + \
+               frame['afa_generated_unique'][frame['afa_generated_unique'].notnull()].astype(str)
```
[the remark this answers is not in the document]
kwhickey: From what I could tell briefly reading the docs on Pandas, string concatenation of a column (a whole Series) and not just a value, is done like this using `+`. I couldn't find explicit docs on comparing to .format, but I was afraid it would "un"-vectorize it if causing each element to be passed to a string-formatting function. 

My guess is that pandas has overloaded the `+` operator on strings to do some kind of "add this to all strings in the Series" in one vectorized operation.

To be safe, and get the most out of performance I think we should stick to this kind of string concat until we can find docs that say how using formatting / interpolation is ok.

dataactvalidator/validation_handlers/validationManager.py
kwhickey: ah, good point. Probably a result of git merges. Not sure why flake didn't flake-out on that. Fixing.

dataactbroker/helpers/validation_helper.py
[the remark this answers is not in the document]
dpb-bah: Works for me 👍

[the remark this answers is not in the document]
tony-sappe: I haven't heard about a security concern, just a (very tiny) performance hit

PR #1923, opened 2020-07-21
[HOTFIX] Fixing Linking TAS Performance (preemptively merged into #1922 already)

dpb-bah:
**High level description:**
Alisa and I discovered some SQL improvements when linking TASs after loading files. The performance should be at least 75% faster.

**Technical details:**
N/A

**Link to JIRA Ticket:**
N/A

### DIFF
--- a/dataactvalidator/validation_handlers/validationManager.py
+++ b/dataactvalidator/validation_handlers/validationManager.py
@@ -4,6 +4,8 @@ import logging
 import os
 import traceback
 import pandas as pd
+import time
+from datetime import timedelta
 
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -827,9 +829,67 @@ def validate_job(self, job_id):
...
+                allocation_transfer_agency,
+                agency_identifier,
+                beginning_period_of_availa,
+                ending_period_of_availabil,
+                availability_type_code,
+                main_account_code,
+                sub_account_code
+        )
+        UPDATE {model}
+        SET tas_id = min_account_num
```
tony-sappe, line 872: Is this update run on data which already contains valid `tas_id` values?

dpb-bah: These are only run on new values we just added that don't have `tas_id` values

tony-sappe: good to know, thanks

PR #1925, opened 2020-07-22
WARMFIX: DEV-5600 Parallelize Data Loading

dpb-bah:
**High level description:**
Updating data loading logic such that it can be parallelized into separate concurrent processes to speed up performance. 

**Technical details:**
* Added `parallel_loading` and `multiprocessing_pools` config variables
* Set the validator jobs to not be daemons
* Updated pytest

**Link to JIRA Ticket:**
[DEV-5600](https://federal-spending-transparency.atlassian.net/browse/DEV-5600)

- [x] Tested locally
- [x] Tested on sandbox


dataactvalidator/sqs_work_dispatcher.py
```diff
@@ -187,7 +187,7 @@ def signal_reset_wrapper(func, args, kwargs):
             name=self.worker_process_name,
             target=signal_reset_wrapper,
             args=(job, job_args, job_kwargs),
-            daemon=True  # daemon=True ensures that if the parent dispatcher dies, it will kill this child worker
+            daemon=False  # daemon=True ensures that if the parent dispatcher dies, it will kill this child worker
```
kwhickey, line 190: One risk with this is that if the parent process does go down unexpectedly, child worker processes could keep on running. And considering the work they do (like validation), it could be resource-intense. 
On the USAS bulk downloader side, we had to set this to false too (since workers spawn psql and zip subprocesses) ... but to compensate, we put in a bunch of extra cleanup code in with `try-finally` blocks around areas that could potentially error-out, which attempted to find and destroy the child worker processes upon exit of the parent. 

Please look to add some similar cleanup to any try-finally wrappers in this code (and add the try-finally wrappers if missing).

Example cleanup functions and usages here: https://github.com/fedspendingtransparency/usaspending-api/commit/bc70a443f3eb0afa0b53b36ae5cd8b2de6f55544

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -416,17 +430,73 @@ def load_file_data(self, sess, bucket_name, region_name):
...
+
+            # Resetting these out here as they are used later in the process
+            self.total_rows = shared_data['total_rows']
+            self.has_data = shared_data['has_data']
+            self.error_rows = shared_data['error_rows']
+            self.error_list = shared_data['error_list']
+            self.reader = temp_reader
+
+    def iterative_data_loading(self, reader_obj):
+        shared_data = {
```
kwhickey, line 462: stylistically, would be better to see this dictionary and the same one in the parallel function being set the same way (curly assignment, or ['indexed'] = val assignment.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -518,10 +584,11 @@ def process_data_chunk(self, sess, chunk_df):
             del offices
 
         # Gathering flex data (must be done before chunk limiting)
-        if self.reader.flex_fields:
-            flex_data = chunk_df.loc[:, list(self.reader.flex_fields + ['row_number'])]
-        if flex_data is not None and not flex_data.empty:
-            flex_data['concatted'] = flex_data.apply(lambda x: concat_flex(x), axis=1)
+        with lockable:
+            if shared_data['flex_fields']:
```
kwhickey, line 588: Since this creates a lock around any processing of flex fields, it can create a bottleneck (sync point) for all chunks. It seems the only access to shared state is the bool flag for flex fields. Does this change on a per-chunk basis? Does anything during the processing of chunks write back to that value in the dict? If not, maybe take this out of the share state dict and just pass it in the function as a boolean, and that way no lock is needed.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -416,17 +430,73 @@ def load_file_data(self, sess, bucket_name, region_name):
...
 
             Args:
-                sess: the database connection
                 chunk_df: the chunk of the file to process as a dataframe
+                shared_data: dictionary of shared data among the chunks
+                m_lock: manager lock if provided to ensure processes don't override each other
         """
+        if m_lock:
+            # make a new connection per process
+            conn = db_connection()
```
kwhickey, line 489: Since the database transaction is an attribute of a connection, this should cause each chunk being processed in an iteration to run in a separate transaction (and separate SQLAlchemy `Session` object, which is not what you want. 
- When in parallel, failure of one chunk won't rollback data from other chunks that have succeeded

I spend a good deal of time ensuring that db sessions started in a parent process could continue to be used in child worker processes the the sqs dispatcher and child workers. I think you should be able to do that here, but I'm not sure if transactions can cross process boundaries and still work. This may make for tricky overall coordination - and perhaps needs some research on "coordinating sqlalchemy transactions over multiprocessing pool"

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -416,17 +430,73 @@ def load_file_data(self, sess, bucket_name, region_name):
...
+            sess = conn.session
+            lockable = m_lock
+        else:
+            sess = GlobalDB.db().session
+            lockable = NoLock()
```
kwhickey, line 494: good to put a comment here that says you're adhering to the locking pattern with a no-op object that bypasses any actual locking

dataactvalidator/validation_handlers/validationManager.py
kwhickey: You could test this out by running a large file in parallel and killing it halfway ... then see if all data was rolled back or not.

tests/integration/error_warning_file_tests.py
```diff
@@ -57,6 +60,9 @@ class ErrorWarningTests(BaseTestValidator):
             submission: the submission foundation to be used for all the tests
             val_job: the validation job to be used for all the tests
     """
+    CHUNK_SIZES = [5]
+    PARALLEL_OPTIONS = [True, False]
+    CONFIGS = list(itertools.product(CHUNK_SIZES, PARALLEL_OPTIONS))
 
     @classmethod
```
kwhickey, line 67: Create additional integration tests that test out what you expect to happen when one of many parallel processes has an exception
1. Are all other sibling processes in the pool that are working on chunks in parallel _also_ stopped when any one causes an error
1. Does the proper exception information bubble up to the sqs worker process, and get logged?
1. Do transactions all roll back and no data saved to the db (may need to run outside of the transactional-rollback test cases, and do your own setup teardown/cleanup for a test, with actual commits)
1. Do all processes get cleaned up: sibling pool processes, pool's parent worker process, worker's parent dispatcher process?
    - You can see if there are tests in the "reaping" code from USAspending. If not, may need to test this manually - kill things, then inspect running processes on the box with `ps -faux`

dataactvalidator/validation_handlers/validationManager.py
dpb-bah: They're technically two different objects but I can restyle them so they look initialized in the same way.

dataactvalidator/validation_handlers/validationManager.py
dpb-bah: Good catch, makes sense, just tested it out, and it all works out.

dataactvalidator/validation_handlers/validationManager.py
dpb-bah: Yeah I ran into many deadlocks using the same db session/connection objects and settled on this way after most of the research concluding something along these lines (https://docs-sqlalchemy.readthedocs.io/ko/latest/faq/connections.html#how-do-i-use-engines-connections-sessions-with-python-multiprocessing-or-os-fork):
 * The db socket gets confused when split/shared into multiple processes.
 * So either remove the poolling from the db engine or sever all connections before splitting and setting them up in the processes.
I _am_ curious how you currently have it setup with the dispatcher though so I'll look into that.

dpb-bah: Updated so that it uses temp tables in case something goes wrong with the data loading.

dataactvalidator/sqs_work_dispatcher.py
dpb-bah: Updated SQSDispatcher and updated ValidationManager to close all child processes upon completion.

tests/integration/error_warning_file_tests.py
dpb-bah: Added a test which should cover these points:
1. To my knowledge, they are not stopped immediately (only solution to that would be either joining the processes as they come in which would be similar to the iterative load or using the `errored` flag I added to the shared_dict on every line of the async method). That being said, the flag I added prevents other loads from starting if it's already errored.
2. Code's been updated such that the proper exception information bubbles up to the parent process which should be carried by the SQS worker process.
3. With the redesign of the database work in the function, the proper table only gets updated on a successful load. The temp tables also get dropped in the `finally` of the loader.
4. With the test I've added, I've checked to make sure the child processes are cleaned up by the end of the function. The cleanup outside of the function has remained the same.

PR #1933, opened 2020-07-29
WARMFIX: Update SQL Validations Memory

dpb-bah:
**High level description:**
Addresses a major source of memory usage when validating submissions.

**Technical details:**
* Yielding the Failure objects instead of storing them all in memory and passing that on
* Chunking out the results of the SQL call instead of fetching them all in memory
* Alembic migration to move function creations to permanent uses.

**Link to JIRA Ticket:**
TBD

- [x] Tested on sandbox
- [x] Smoke tests locally
- Documentation Updated


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -653,41 +652,42 @@ def run_sql_validations(self, short_colnames, writer, warning_writer):
...
-            self.error_list.record_row_error(self.job.job_id, self.file_name, field_name, failure.error,
-                                             self.total_rows, failure.original_label, failure.file_type_id,
-                                             failure.target_file_id, failure.severity_id)
-        return error_rows
+        for failure in validate_file_by_sql(self.job, self.file_type.name,
+                                            self.short_to_long_dict[self.file_type.file_type_id],
+                                            batch_results=BATCH_SQL_VAL_RESULTS):
+            self.process_failure(failure, short_colnames, writer, warning_writer)
+
+    def process_failure(self, failure, short_colnames, writer, warning_writer):
```
alburde1, line 660: Just curious why you turned this into a function? We don't use it elsewhere so I'm not sure why it was necessary rather than just leaving the code in the for loop

dpb-bah: Ah. I think that was a leftover from some earlier design I was toying around with it. I can move it back in.

dataactvalidator/validation_handlers/validator.py
```diff
@@ -41,36 +47,28 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict, job_id, error_c
...
-                       'Starting flex field gathering and file writing',
-            'message_type': 'ValidatorInfo',
-            'rule': rule.query_name,
-            'job_id': job_id,
-            'submission_id': submission_id
-        })
-        if failed_rows.rowcount:
-            rule_cols = failed_rows.keys()
+
+        def process_failures(failures, columns, batch_num=0):
```
alburde1, line 51: If this is a separate function entirely, should we move it out of this function and just call it? The only reason all this processing was in here to begin with was because it was all being done inside this function. If we're making a new function we could move it out and make it easier to read.

dataactvalidator/validation_handlers/validator.py
```diff
@@ -208,16 +243,16 @@ def validate_file_by_sql(job, file_type, short_to_long_dict):
...
             'action': 'run_sql_validation_rule',
             'status': 'start',
-            'start_time': rule_start
+            'start_time': rule_start,
+            'batch_results': batch_results
         })
 
-        failures = sess.execute(rule.rule_sql.format(job.submission_id))
-        if failures.rowcount:
+        def process_batch(failures, columns):
```
alburde1, line 250: Same comment as above for this function

dataactvalidator/validation_handlers/validator.py
```diff
@@ -120,60 +120,95 @@ def cross_validate_sql(rules, submission_id, short_to_long_dict, job_id, error_c
...
+                    for failure_key in columns:
                         # Difference
                         if failure_key == 'difference':
                             difference = str(row[failure_key] if row[failure_key] is not None else '')
                         if failure_key.startswith(diff_start):
                             diff_header = failure_key[len(diff_start):]
                             diff_array.append('{}: {}'.format(diff_header, str(row[failure_key] if
-                                                                               row[failure_key] is not None else '')))
+                                                                               row[
+                                                                                   failure_key] is not None else '')))
```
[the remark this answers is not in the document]
dpb-bah: Stange. I'm unsure how this turned out that way. Fixing.

dataactvalidator/validation_handlers/validator.py
dpb-bah: Agreeing with the sentiment though it turns out there are a bunch of shared scoped elements that would need to be converted to arguments. For example:
`def process_cross_failures(failures, columns, short_to_long_dict, rule, submission_id, job_id, error_csv, warning_csv, batch_num=0):`

PR #1939, opened 2020-08-10
Sprint 113: QAT -> Master

alburde1:


dpb-bah:
* Significantly improved the performance of FABS derivations and publications by using SQL calls and compressed tables.
* Reworked the data ingesting process to be parallelized simultaneously to improve data validation performance.
* Reworked the processing of SQL validations to be batched to save on large batches of memory and help performance.

PR #1955, opened 2020-08-25
Sprint 114: Dev -> QAT

alburde1:


dpb-bah:
* Renamed database tables and fields for clarity.
* Updated publishing status to prevent duplicate published data.
* Updated backfilling business categories script to account for transactions with multiple business types.
* Created DEFC table to store DEFC values to be used for rules B24 and C25.
* Added submission type filter to the list submissions endpoint.
* Added a day to the publish and certification deadlines as the deadline is the night of that date.
* Added option to generate D1 with FPDS Element numbers in the headers
* Updated submission table's submission history to update upon certification.

PR #1956, opened 2020-08-27
DEV-5955: metadata box last modified

alburde1:
**High level description:**
Updating the metadata box above submissions to use the proper modified date instead of only checking the submission's updated_at

**Technical details:**
Uses the same table as the frontend submission table to get the last_modified date so if a job has changed that didn't cause the submission to change, the most recent modification of any kind will be the one displayed.

**Link to JIRA Ticket:**
[DEV-5955](https://federal-spending-transparency.atlassian.net/browse/DEV-5955)

- Documentation Updated

### DIFF
--- a/dataactcore/interfaces/function_bag.py
+++ b/dataactcore/interfaces/function_bag.py
@@ -688,5 +688,5 @@ def get_last_modified(submission_id):
     submission_updated_view = SubmissionUpdatedView()
     sess = GlobalDB.db().session
     last_modified = sess.query(submission_updated_view.updated_at).\
-        filter(submission_updated_view.submission_id == submission_id).first().updated_at
-    return last_modified
+        filter(submission_updated_view.submission_id == submission_id).first()
+    return last_modified.updated_at if last_modified else None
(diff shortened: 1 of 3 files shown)


dataactcore/interfaces/function_bag.py
```diff
@@ -660,16 +661,32 @@ def get_fabs_meta(submission_id):
...
+            submission_id: submission ID to get the last modified for
+
+        Returns:
+            the last modified date of the provided submission or None if the submission doesn't exist
+    """
+    submission_updated_view = SubmissionUpdatedView()
+    sess = GlobalDB.db().session
+    last_modified = sess.query(submission_updated_view.updated_at).\
+        filter(submission_updated_view.submission_id == submission_id).first()
+    return last_modified.updated_at if last_modified else None
```
dpb-bah, line 692: Wondering if we should have it default to `submission.updated_at` if for some reason it's not in `SubmissionUpdatedView`

dpb-bah (changes requested):

dataactcore/interfaces/function_bag.py
alburde1: But a view is basically a virtual table. It gets updated automatically every time the elements it pulls from are updated. In order for it to be a submission and not in the view it'd have to either already not have an `updated_at` or the entirety of SQL would have to have a bug.

dpb-bah (approved):

PR #1960, opened 2020-09-04
Sprint 115: Dev -> QAT

alburde1:


dpb-bah:
* Modified C23 to be ignored when transaction obligated amount isn't provided.
* Updated submission_metadata's "Last Modified" values to match what is displayed on the Submission Table.
* Added a "Submission Type" filter to the Active DABS Submission Table, allowing users to filter out test or certifiable submissions. 
* Updated front-end behavior various DABS pages when a submission is reverting to prevent potential issues.

dpb-bah (approved):

PR #1969, opened 2020-09-22
Sprint 116: Dev -> QAT

alburde1:
Sprint 116: Dev -> QAT


dpb-bah (approved):

dpb-bah:
* Updated frontend behavior to prevent potential workflow bugs when publishing submissions
* Updated frontend behavior when making submissions to immediately notify users that a published submission already exists for that agency and period.
* Updated frontend behavior to immediately detect and notify users when they are uploading an invalid file type (i.e. not csv or txt) 
* Updated delete submission functionality to account for stored submission files
* Created nightly script to remove expired submissions, which we are defining as:
   * unpublished test DABS submissions that haven't been modified in over six months
   * unpublished DABS submissions from FY17Q1

PR #1976, opened 2020-10-02
DEV-6194: Starting with Long Rows Bug

dpb-bah:
**High level description:**
It was discovered that when uploading a file that started off with rows with too many fields, it failed the job. This resolves that by skipping said long rows in the beginning.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-6194](https://federal-spending-transparency.atlassian.net/browse/DEV-6194)

- [x] Merged after #1973
- [x] Manually tested
- Documentation Updated

### DIFF
--- a/dataactvalidator/validation_handlers/validationManager.py
+++ b/dataactvalidator/validation_handlers/validationManager.py
@@ -170,8 +170,10 @@ class ValidationManager:
         self.error_rows = []
         self.total_rows = 0
         self.total_data_rows = 0
+        self.short_rows = []
         self.short_pop_rows = []
         self.short_null_rows = []
+        self.long_rows = []
         self.long_pop_rows = []
         self.long_null_rows = []
         self.has_data = False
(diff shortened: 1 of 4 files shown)


dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -591,9 +602,6 @@ def process_data_chunk(self, chunk_df, shared_data, m_lock=None):
         # index gets reset for each chunk, adding the header, and adding previous rows
         chunk_df['row_number'] = chunk_df.index + 2
 
-        # Some long null rows may get included in the resulting dataframe, this ensures they are removed
-        chunk_df = chunk_df[~chunk_df['row_number'].isin(self.long_null_rows)]
-
```
alburde1, line 596: Is it no longer possible for a chunk to have long rows? It can't have long rows at the beginning but what if a chunk has a long row at the end? Or in the middle? We still don't want those right?

alburde1 (changes requested):

dataactvalidator/validation_handlers/validationManager.py
dpb-bah: They still shouldn't get surfaced via pandas like normal. They only randomly surface when the whole chunk is filled with rows with format errors. The reason why we skip them in the front is that pandas processes the rest of the rows wrong if it tries to process the first long rows.

alburde1 (approved):

PR #1984, opened 2020-10-20
Sprint 118: Dev -> Qat

alburde1:


dpb-bah (approved):

dpb-bah:
* Updated frontend pages to notify users of expiring submissions.
* Adding FY21 Submission window schedule to the Help page and in the database.
* Updated the design of the maintenance page when deploying the API.

PR #1997, opened 2020-11-20
Hotfix: WSDL SOAP SUDS HTTP Proxies

dpb-bah:
**High level description:**
The current implementation and use of suds doesn't account for http proxies. This makes the FSRS and DUNS historical loads break on the DTI environment. This PR simply uses the HTTP proxy environment variables if there are any. 

**Technical details:**
N/A

**Link to JIRA Ticket:**
N/A

- [x] Successfully tested with the [DTI subaward load](http://10.117.0.20:8080/job/Broker-FSRS-Subaward-Load/906/console)
- Documentation Updated

### DIFF
--- a/dataactbroker/fsrs.py
+++ b/dataactbroker/fsrs.py
@@ -12,6 +12,7 @@ from suds.xsd import doctor
 from dataactcore.config import CONFIG_BROKER
 from dataactcore.models.fsrs import FSRSProcurement, FSRSSubcontract, FSRSGrant, FSRSSubgrant
 from dataactcore.models.domainModels import States
+from dataactbroker.helpers.generic_helper import WellBehavedHttpTransport
 
 logger = logging.getLogger(__name__)
 PROCUREMENT = 'procurement_service'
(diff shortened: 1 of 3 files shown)


dataactbroker/fsrs.py
```diff
@@ -146,6 +147,16 @@ def new_client(service_type):
...
 
+    proxy_options = {}
+    http_proxy = os.environ.get('HTTP_PROXY')
+    if http_proxy:
+        proxy_options['http'] = http_proxy
+    https_proxy = os.environ.get('HTTPS_PROXY')
+    if https_proxy:
+        proxy_options['https'] = https_proxy
+    if proxy_options:
+        options['proxy'] = proxy_options
```
bminahan-kc, line 158: `no_proxy` should be set the same way, so that you don't accidentally try and use a proxy for something that is no_proxy'd

dataactbroker/helpers/generic_helper.py
```diff
@@ -112,8 +113,19 @@ def format_internal_tas(row):
...
+    options = {}
+    proxy_options = {}
+    http_proxy = os.environ.get('HTTP_PROXY')
+    if http_proxy:
+        proxy_options['http'] = http_proxy
+    https_proxy = os.environ.get('HTTPS_PROXY')
+    if https_proxy:
+        proxy_options['https'] = https_proxy
+    if proxy_options:
+        options['proxy'] = proxy_options
```
bminahan-kc, line 125: same as above

dataactbroker/fsrs.py
dpb-bah: Good catch. I don't know if the suds package accounts for no_proxy (at least it isn't well documented). I'm gonna try another solution first, one that bypasses suds' own proxy settings.

dpb-bah: There we go. I had to somewhat tweak this solution but seems like it's working.
https://stackoverflow.com/a/64974205/4166584

alburde1 (approved):

PR #2013, opened 2021-01-13
DEV-6628: Submission Comment in API

dpb-bah:
**High level description:**
* Adding an overall submission comment to the comments and certified comments tables.
* Updating the endpoints which get/update these comments. 

**Technical details:**
* Alembic upgrade to include new columns

**Link to JIRA Ticket:**
[DEV-6628](https://federal-spending-transparency.atlassian.net/browse/DEV-6628)

- [x] Documentation Updated


dataactbroker/handlers/fileHandler.py
```diff
@@ -1242,10 +1242,14 @@ def get_submission_comments(submission):
...
     result = {letter: '' for letter in FILE_TYPE_DICT_LETTER.values() if letter != 'FABS'}
+    result['submission_comment'] = ''
     comments = sess.query(Comment).filter_by(submission_id=submission.submission_id)
     for comment in comments:
-        letter = FILE_TYPE_DICT_LETTER[comment.file_type_id]
-        result[letter] = comment.comment
+        if comment.submission_comment and comment.file_type_id is None:
+            comment_type = 'submission_comment'
+        else:
+            comment_type = FILE_TYPE_DICT_LETTER.get(comment.file_type_id, None)
```
alburde1, line 1251: couldn't we just do `comment_type = FILE_TYPE_DICT_LETTER.get(comment.file_type_id, 'submission_comment')` ? I'm trying to figure out what other situation we have no file type ID for.

dataactcore/models/jobModels.py
```diff
@@ -196,8 +196,9 @@ class Comment(Base):
     submission_id = Column(Integer, ForeignKey("submission.submission_id", name="fk_submission", ondelete="CASCADE"),
                            nullable=False)
     submission = relationship(Submission, uselist=False, cascade="delete")
-    file_type_id = Column(Integer, ForeignKey("file_type.file_type_id", name="fk_file_type"), nullable=False)
+    file_type_id = Column(Integer, ForeignKey("file_type.file_type_id", name="fk_file_type"), nullable=True)
     file_type = relationship(FileType, uselist=False)
+    submission_comment = Column(Boolean, nullable=False, default=False, server_default="False")
```
alburde1, line 201: I'm curious why we need a boolean flag here actually. We could just search on `file_type_id IS NULL` if we want submission comments.

alburde1 (changes requested):

dataactcore/models/jobModels.py
dpb-bah: That's definitely doable. The intent was to be explicit/obvious to future developers. If we move to that, I can add an extra comment in the model to account for that.

dataactbroker/handlers/fileHandler.py
dpb-bah: That works. The code was based on the `comment.submission_comment` in addition to the none `file_type_id` and can be updated.

alburde1 (approved):

PR #2031, opened 2021-02-09
Sprint 126: Dev -> QAT

alburde1:


dpb-bah (approved):

dpb-bah:
* Updated the DABS Review page:
    * Cleaned up the look of the page for clarity.
    * Added an overall Submission Comment for general comments about the whole submission.
    * Updated the UI to expand/collapse comments.
    * Comment boxes now detect and warn users if they are including non-ASCII text which may affect their submission Assurance Statements.
    * The page will also inform the user if there are unsaved comments on the page and if they leave the page as well.

PR #2044, opened 2021-02-24
DEV-6926: DUNS/Exec Comp Forced Reloads

dpb-bah:
**High level description:**
Adding `--force_reload` arguments to the DUNS and Exec Comp loaders which forces a reload of data from a specified date.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-6926](https://federal-spending-transparency.atlassian.net/browse/DEV-6926)

- [x] Successfully run on dev and production to reload data from 2021-02-06
- Documentation Updated

### DIFF
--- a/dataactcore/scripts/load_duns.py
+++ b/dataactcore/scripts/load_duns.py
@@ -16,7 +16,7 @@ from dataactvalidator.health_check import create_app
 logger = logging.getLogger(__name__)
 
 
-def process_duns_dir(sess, historic, local, benchmarks=None, metrics=None, force_reload=None):
+def process_duns_dir(sess, historic, local, benchmarks=None, metrics=None, reload_date=None):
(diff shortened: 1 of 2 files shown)


dataactcore/scripts/load_duns.py
```diff
@@ -16,7 +16,7 @@
 logger = logging.getLogger(__name__)
 
 
-def process_duns_dir(sess, historic, local, benchmarks=None, metrics=None):
+def process_duns_dir(sess, historic, local, benchmarks=None, metrics=None, force_reload=None):
```
alburde1, line 19: Normally we use `force_reload` to require it to reload the data even if we have no changes in the file. This is normally just a flag. I think we should keep to that standard and name this something like `reload_date` instead

alburde1 (changes requested):

alburde1 (approved):

PR #2056, opened 2021-03-22
Sprint 129: Dev -> Qat

alburde1:


dpb-bah (approved):

dpb-bah:
* Updated FREC agency names to use the FREC abbreviation instead of the parent CGAC abbreviation.
* Resolved a minor FE bug with total obligations when reverting.
* Updated the error state messages of the DABS Review page.

PR #2066, opened 2021-04-06
Sprint 130: Dev -> Qat

alburde1:


dpb-bah (approved):

dpb-bah:
* Added V to list of DEF codes.
* Updated the GTAS SF133 loader to de-pad any multiple character DEFC to their single-character equivalent until it's the character limit has been established.
* Renamed `tas_id` to `account_num` in SF133 table for database clarity.
* Cleared out error messages when restarting or preparing jobs for database clarity.
* Removed/updated a couple packages out of security concerns and docker setup.

PR #2067, opened 2021-04-08
DEV-7030: only failing differences c20 c21 c26

alburde1:
**High level description:**
Only display the difference values in rules C20, C21, and C26 for values that are incorrect

**Technical details:**
The three rules now return one column labeled "difference" with the entire string instead of a separate column for each difference that then was concatenated for the report.

**Link to JIRA Ticket:**
[DEV-7030](https://federal-spending-transparency.atlassian.net/browse/DEV-7030)

- Documentation Updated


dpb-bah (approved):
Reapplied the changes myself and confirmed no typos.

PR #2073, opened 2021-04-20
Sprint 131: Dev -> QAT

alburde1:


dpb-bah:
* Updated C20, C21, and C26 to only return relevant differences in the error/warning reports.
* Updated Federal Hierarchy loader to process inactive offices and run on incremental updates.
* Preventing users from starting revalidation while a job is still running.
* Resolving some minor frontend bugs (sorting errors and warnings, fixing banners, and displaying error messages properly)

PR #2098, opened 2021-05-24
Hotfix: FPDS Loader Retrying Empty Responses

dpb-bah:
**High level description:**
Updating the FPDS loader such that:
- it'll only request the pages we expect data for
- it'll retry requests if they don't include data for connection issues.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-7336](https://federal-spending-transparency.atlassian.net/browse/DEV-7336)

- [x] Tested locally and on QAT

### DIFF
--- a/dataactcore/scripts/pull_fpds_data.py
+++ b/dataactcore/scripts/pull_fpds_data.py
@@ -9,6 +9,7 @@ import datetime
 import time
 import re
 import json
+import math
 
 from sqlalchemy import func
 
@@ -47,7 +48,7 @@ FPDS_NAMESPACES = {'http://www.fpdsng.com/FPDS': None,
 
 # Used for asyncio get requests against the ATOM feed
 MAX_ENTRIES = 10
-REQUESTS_AT_ONCE = 100
+MAX_REQUESTS_AT_ONCE = 100
 SPOT_CHECK_COUNT = 10000
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


dataactcore/scripts/pull_fpds_data.py
```diff
@@ -1477,16 +1481,21 @@ def get_data(contract_type, award_type, now, sess, sub_tier_list, county_by_name
 
     entries_processed = 0
     while True:
-        async def atom_async_get(entries_already_processed):
+        # pull in the next MAX_ENTRIES * REQUESTS_AT_ONCE until we get anything less than the MAX_ENTRIES
+        async def atom_async_get(entries_already_processed, total_expected_records):
             response_list = []
             loop = asyncio.get_event_loop()
+            requests_at_once = MAX_REQUESTS_AT_ONCE
+            if total_expected_records - entries_already_processed < (MAX_REQUESTS_AT_ONCE * MAX_ENTRIES):
```
alburde1, line 1489: What if we expect 60 entries and we end up with 61 entries in the feed? Pulling more entries than expected protected us from that. This way, sure, if we get more than expected when it's not a round number it works, but if it's a multiple of 10 we miss things. Most often the "different amounts" are from us getting less records than expected but we can't rule out the opposite.

dpb-bah: Solution was adding an extra call to check for additional records.

alburde1 (approved):

PR #2128, opened 2021-07-07
Sprint 136: Qat -> Staging

alburde1:


dpb-bah (approved):

dpb-bah:
* Deprecating Rule B18
* Removing all instances of DEFC 9 from the rules and code.
* Updating Rules B11 and B12 to account for the new Object Class Updates 
* Updating the shared object class file and loader to only support the 3-digit codes and their 4-digit equivalents suffixed with an additional zero.

PR #2143, opened 2021-07-27
DEV-7367: FABS File Updates

dpb-bah:
**High level description:**
Adding/removing various columns to the FABS schema per DAIMS 2.1.

**Technical details:**
* Adding `IndirectCostFederalShareAmount`, `FundingOpportunityNumber`, `FundingOpportunityGoalsText`, `AwardeeOrRecipientDUNS`, `AwardeeOrRecipientUEI` 
* Removing `AwardeeOrRecipientUniqueIdentifier`
* Includes Alembic migration
* To prevent USAS from renaming columns, added a quick derivation upon loading to rename the recipient cols from the DAIMS short to how they're already stored in the models. 
* Note: this does not include the column updates to the SQL rules, that will be included in another ticket.

**Link to JIRA Ticket:**
[DEV-7367](https://federal-spending-transparency.atlassian.net/browse/DEV-7367)

- [x] Documentation Updated
- [x] Tested locally, able to upload and publish FABS data and seeing that data in the database


alburde1 (changes requested):
We can't remove AwardeeOrRecipientUniqueIdentifier until April FY22. I don't know why it didn't get put into the description but it was discussed separately from the ticket and in the comments of the ticket. They can submit either one until then and if they submit DUNS then we have to be able to use it, which we cannot if it is removed from the expected columns.

alburde1 (approved):

PR #2196, opened 2021-10-19
DEV-7967: Public Loads Export Updates

dpb-bah:
**High level description:**
Minor updates to loaders based on the changes to surface some public files.

**Technical details:**
* Adding `public_files_bucket` config value
* Fixed typo in DUNS loader
* `state_list.txt` -> `state_list.csv`
* load `GOVT_UNITS.txt` and `NationalFedCodes.txt` from the public files bucket instead of the SF133 bucket for the location loader
* `load_zips` exports `state_congressional.csv` to public S3 

**Link to JIRA Ticket:**
[DEV-7967](https://federal-spending-transparency.atlassian.net/browse/DEV-7967)

- [ ] Merged before [the Broker Config PR](https://github.com/fedspendingtransparency/data-act-broker-config/pull/656)
- Documentation Updated

### DIFF
--- a/dataactvalidator/scripts/read_zips.py
+++ b/dataactvalidator/scripts/read_zips.py
@@ -439,7 +439,7 @@ def read_zips():
         hot_swap_zip_tables(sess)
         update_state_congr_table_current(sess)
         update_state_congr_table_census(census_file, sess)
-        if not CONFIG_BROKER['use_aws']:
+        if CONFIG_BROKER['use_aws']:
             export_state_congr_table(sess)
 
         logger.info("Zipcode script complete")


dataactvalidator/scripts/load_location_data.py
```diff
@@ -297,19 +297,23 @@ def load_location_data(force_reload=False):
     """
     if CONFIG_BROKER["use_aws"]:
         s3_client = boto3.client('s3', region_name=CONFIG_BROKER['aws_region'])
-        city_file = s3_client.generate_presigned_url('get_object', {'Bucket': CONFIG_BROKER['sf_133_bucket'],
-                                                                    'Key': "NationalFedCodes.txt"}, ExpiresIn=600)
-        county_file = s3_client.generate_presigned_url('get_object', {'Bucket': CONFIG_BROKER['sf_133_bucket'],
-                                                                      'Key': "GOVT_UNITS.txt"}, ExpiresIn=600)
-        state_file = s3_client.generate_presigned_url('get_object', {'Bucket': CONFIG_BROKER['sf_133_bucket'],
-                                                                     'Key': "state_list.txt"}, ExpiresIn=600)
+        city_file = s3_client.generate_presigned_url('get_object',
```
alburde1, line 300: This is a question for all of these. If they're going in a public bucket, is there any reason we need to separate this and the local versions? Do we have to generate presigned URLs at all or can we just get the file like we do with the agencies files and skip boto for everything except ctystate?

dataactvalidator/scripts/read_zips.py
```diff
@@ -418,6 +439,7 @@ def read_zips():
         hot_swap_zip_tables(sess)
         update_state_congr_table_current(sess)
         update_state_congr_table_census(census_file, sess)
+        export_state_congr_table(sess)
```
alburde1, line 442: Should this be in an "if" and only done if not running locally? We wouldn't want our locals to accidentally overwrite important things.

dpb-bah: It's a good point. It'd be super hard nowadays for local users to have AWS access to the buckets but regardless that would prevent them from it being an issue at all.

PR #2213, opened 2021-12-01
Sprint 143: Qat -> Staging

alburde1:


dpb-bah:
* Implemented progress bars for FABS and DABS validations, keeping users informed on the states of their submissions.

dpb-bah (approved):

PR #2223, opened 2021-12-20
Sprint 144: Qat -> Staging

alburde1:


dpb-bah (approved):

dpb-bah:
* Modified TAS loader to include TAS/Display TAS values when updating data.
* Updated program activity loader to export a public version of the data to be shared on the Data Sources page.
* Modified rule C18 to only trigger when all values are present.
* Moved temporary FABS Business Categories SQL function to a more-permanent migration.
* Extended the functionality of the dynamic banners to be presented on the login page (for when there are MAX issues).
* Added cross-file validation progress bars to keep the user informed on its status.
* Updated rules B19, C28, and FABS 2.1 to be consistent in what they mark as duplicate rows.
* Updated the quarter picker on the Generate File A page to display new periods on the first of the next month instead of waiting for the GTAS period to complete.
* Updated DABS Publish History to include full timestamps instead of just dates.
* Removed link to FY21 Submission Calendar on Broker help page.
* Updated DABS creation page to automatically default new submissions as monthly.

PR #2272, opened 2022-02-28
Sprint 147: Staging -> Master

alburde1:


dpb-bah:
* Updated FABS45 rule description for clarity.
* Updated A36 to account for multiple amounts in SF133 grouped by the same TAS.
* Updated the default time period when creating DABS submissions to be the next period after the publication deadline has been passed. 
* Added deadline to the `latest_publication_period` endpoint.
* Improved the logout functionality to ensure users remained logged out until they log back in.

dpb-bah (approved):

PR #2274, opened 2022-03-07
DEV-8493: ExecComp using UEI, V1/V2 updates

dpb-bah:
**High level description:**
* Updating the Exec Comp loader to use UEI for V2
* Updating the DUNS and Exec Comp loaders to update the same recipient based on whether DUNS and/or UEI is provided.

**Technical details:**
The technical details can be placed here for the knowledge of other developers. Any detailed caveats or specific deployment steps should be outlined here.

**Link to JIRA Ticket:**
[DEV-1234](https://federal-spending-transparency.atlassian.net/browse/DEV-1234)

- Merged concurrently with [Frontend#1234](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/1234)
- [x] Documentation updated


dataactcore/utils/duns.py
```diff
@@ -270,32 +269,58 @@ def update_duns(sess, duns_data, table_name='duns', metrics=None, deletes=False,
             'updated_duns': []
         }
 
+    if key_cols is None:
+        key_cols = ['awardee_or_recipient_uniqu']
+    elif key_cols not in (['awardee_or_recipient_uniqu'], ['awardee_or_recipient_uniqu', 'uei']):
```
alburde1, line 274: If we're requiring DUNS and making UEI optional, why not always have DUNS and just have a boolean flag to include UEI? It seems simpler and cleaner overall.

alburde1 (changes requested):

dataactcore/utils/duns.py
dpb-bah: I think that makes sense. I'll name it something like `includes_uei`.

alburde1 (approved):

PR #2344, opened 2022-07-19
Sprint 154: Qat -> Staging

alburde1:


dpb-bah:
* Added tests for SF133 and failed TAS loaders for ensured accuracy.
* Added DEFC 7 to the list of accepted DEFCs.

dpb-bah (approved):

PR #2409, opened 2023-03-14
DEV-9364: subaward loader source

alburde1:
**High level description:**
Updating where we get prime-level subaward data for the subaward table

**Technical details:**
Changing many of the columns in subawards to be derived from our existing data rather than what FSRS sends us

**Link to JIRA Ticket:**
[DEV-9364](https://federal-spending-transparency.atlassian.net/browse/DEV-9364)

- Style Guide check completed
- Documentation updated


dataactcore/scripts/raw_sql/link_broken_subaward_contracts.sql
```diff
@@ -10,44 +10,215 @@ WITH unlinked_subs AS
...
+    place_of_perform_state_name = ldap.place_of_perfor_state_desc,
+    place_of_performance_zip = ldap.place_of_performance_zip4a,
+    place_of_perform_congressio = ldap.place_of_performance_congr,
+    place_of_perform_country_co = ldap.place_of_perform_country_c,
+    place_of_perform_country_na = ldap.place_of_perf_country_desc,
+    award_description = bdap.award_description,
+    naics = ldap.naics,
+    naics_description = ldap.naics_description,
+
+    -- File F Prime Awards
```
dpb-bah, line 204: Feels out of place, especially given all of these are at the prime award level. I know this was written here beforehand when we were originally making the File F/Subaward table. I think we were using these to keep track of "stuff we pull from FSRS, stuff we pull from Transactions, stuff we use in File F, and stuff we just have for tracability/tracking." I dunno if we want to make those distinctions again now or after we reapproach these tables again if that's slated for anytime soon (probably alongside when we update USAS's subaward approach, so we can then update it alongside the Ultimate Subaward Crosswalk spreadsheet). For now we can leave these in as a reminder to review this.

dpb-bah (changes requested):

dpb-bah (approved):

PR #2423, opened 2023-04-18
Fix: ZipCity Loader updates

dpb-bah:
**High level description:**
Includes several fixes to the zip_city loader
* using the right name
* grouping by zip + state + city (and not just zip)

**Technical details:**
N/A

**Link to JIRA Ticket:**
N/A

- [ ] Unit & integration tests updated with relevant test cases
- [x] Documentation updated
- [ ] Approved by POs

### DIFF
--- a/dataactvalidator/scripts/load_location_data.py
+++ b/dataactvalidator/scripts/load_location_data.py
@@ -176,9 +176,10 @@ def parse_zip_city_file(f):
                 zip_code = curr_row[1:6]
                 city_name = curr_row[62:90].strip()
                 state_code = curr_row[99:101]
-                data_dict[zip_code] = {'zip_code': zip_code,
-                                       'city_name': city_name,
-                                       'state_code': state_code}
+                zip_city_key = zip_code + city_name + state_code
+                data_dict[zip_city_key] = {'zip_code': zip_code,
+                                           'city_name': city_name,
+                                           'state_code': state_code}
 
             # cut the current line out of the chunk we're processing
             curr_chunk = curr_chunk[line_size:]
--- a/dataactvalidator/scripts/load_location_data.py
+++ b/dataactvalidator/scripts/load_location_data.py
@@ -174,7 +174,7 @@ def parse_zip_city_file(f):
             curr_row = curr_chunk[:line_size]
             if curr_row[0] == "D":
                 zip_code = curr_row[1:6]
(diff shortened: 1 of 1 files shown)


dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -362,10 +362,12 @@ def derive_ppop_location_data(sess, submission_id):
     # Deriving city info for transactions with zips
     query = """
         UPDATE tmp_fabs_{submission_id}
-        SET place_of_performance_city = city_name
+        SET place_of_performance_city = preferred_city_name
         FROM zip_city
         WHERE place_of_performance_zip5 IS NOT NULL
-            AND zip_city.zip_code = place_of_performance_zip5;
+            AND zip_city.zip_code = place_of_performance_zip5
+            AND place_of_perfor_state_code IS NOT NULL
```
alburde1, line 369: This is extraneous. If it's NULL it'll just not match on anything

dataactbroker/helpers/fabs_derivations_helper.py
```diff
@@ -628,10 +630,12 @@ def derive_le_location_data(sess, submission_id):
     # Deriving city info for records with zips (type 2 and 3)
     query = """
         UPDATE tmp_fabs_{submission_id}
-        SET legal_entity_city_name = city_name
+        SET legal_entity_city_name = preferred_city_name
         FROM zip_city
         WHERE legal_entity_zip5 IS NOT NULL
-            AND zip_city.zip_code = legal_entity_zip5;
+            AND zip_city.zip_code = legal_entity_zip5
+            AND legal_entity_state_code IS NOT NULL
```
alburde1, line 637: Same as above comment

alburde1 (changes requested):

dataactbroker/helpers/fabs_derivations_helper.py
dpb-bah: Should the `place_of_performance_zip5 IS NOT NULL`s be removed as well?

alburde1: Those can be left. If they're not null it's most likely that the state will also not be null. One check can narrow the list, the second check will end up being extraneous.

alburde1 (dismissed):

alburde1 (approved):

PR #2452, opened 2023-06-20
DEV-9489: CAIA Login

dpb-bah:
**High level description:**
Adds a new `/v1/caia_login` endpoint taking in the code and redirect_uri to log the user into the Broker and set their permissions based on their roles.

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-9489](https://federal-spending-transparency.atlassian.net/browse/DEV-1234)

- [x] Merged concurrently with [Frontend#1234]()
- [x] Client Secrets for BE updated in all environments
- [x] Documentation updated


dataactbroker/handlers/account_handler.py
```diff
@@ -165,6 +170,79 @@ def max_login(self, session):
             # Return 500
             return JsonResponse.error(e, StatusCode.INTERNAL_ERROR, error=str(e))
 
+    def caia_login(self, session):
+        """ Logs a user in if their password matches using CAIA
```
alburde1, line 174: Minor thing but I don't think "password" is the right way to explain this. Perhaps something like "if their CAIA validation succeeds"

dataactbroker/handlers/account_handler.py
```diff
@@ -429,6 +529,88 @@ def get_max_dict(ticket, service):
...
+    """
+    headers = {"Authorization": f"Bearer {accces_token}"}
+    caia_resp = requests.get(f"{CONFIG_BROKER['caia']['url_root']}/idp/userinfo.openid", headers=headers)
+    caia_resp.raise_for_status()
+
+    return json.loads(caia_resp.content.decode())
+
+
+def revoke_caia_access(refresh_token):
+    """ Simple revokes access to the given CAIA refresh token. We are revoking the refresh token (and not the access
```
alburde1, line 599: Since we're already editing, this should probably be "Simply" not "Simple"

alburde1 (changes requested):
Can we update the `max_login` endpoint documentation as well? See the `get_certified_file` endpoint for an example

PR #2463, opened 2023-07-27
DEV-9272: automatic file a generation

alburde1:
**High level description:**
Automatically generate file A when creating a new submission

**Technical details:**
Ignore file As that are sent in by the user only when first creating a submission. If no file A is sent in that's also fine.

**Link to JIRA Ticket:**
[DEV-9272](https://federal-spending-transparency.atlassian.net/browse/DEV-9272)

- [ ] Merged concurrently with [Frontend#1787](https://github.com/fedspendingtransparency/data-act-broker-web-app/pull/1787)
- Documentation updated


dataactbroker/handlers/fileHandler.py
```diff
@@ -959,14 +976,15 @@ def publish_fabs_submission(submission):
         response_dict = {'submission_id': submission_id}
         return JsonResponse.create(StatusCode.OK, response_dict)
 
-    def build_file_map(self, file_dict, file_type_list, upload_files, submission):
+    def build_file_map(self, file_dict, file_type_list, upload_files, submission, existing_submission=False):
```
dpb-bah, line 979: Instead of adding another parameter and making this less agnostic to the submissions and file types, could this logic be outside of the call and simply update the incoming `file_type_list` to exclude appropriations when it's not an existing submission? It's not a pressing point, just felt like `file_type_list` could do the same and keep the function simpler.

dataactbroker/handlers/generation_handler.py
```diff
@@ -72,6 +72,9 @@ def generate_file(submission, file_type, start, end, agency_type, file_format):
     try:
         if file_type in ['D1', 'D2']:
             generation_helper.start_d_generation(job, start, end, agency_type, file_format=file_format)
+        elif file_type == 'A':
+            agency_code = submission.frec_code if submission.frec_code else submission.cgac_code
```
dpb-bah, line 76: Could use Python's coalesce
```suggestion
            agency_code = submission.frec_code or submission.cgac_code
```

dpb-bah (changes requested):

dpb-bah (dismissed):

dpb-bah (approved):

PR #2620, opened 2024-07-22
DEV-11179: bidirectional boc comparison report

alburde1:
**High level description:**
Making the BOC Comparison Report bidirectional so values from file B and also values from the BOC data that would have been bucketed into this agency but weren't in file B for any reason

**Technical details:**
Moving the bucketing logic to a helper file so any file that needs to access it can do so.

**Link to JIRA Ticket:**
[DEV-11179](https://federal-spending-transparency.atlassian.net/browse/DEV-11179)

- Documentation updated


dpb-bah (approved):
LGTM. Nice refactor so we still only need to update that logic in two places (not three or more) if it changes.

PR #2631, opened 2024-08-29
DEV-11376: deprecation fixes

alburde1:
**High level description:**
Fixing utcnow deprecation warnings in advance of updating python/sqlalchemy versions in the future

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-11376](https://federal-spending-transparency.atlassian.net/browse/DEV-11376)

- Documentation updated

### DIFF
--- a/dataactcore/utils/timeStampMixin.py
+++ b/dataactcore/utils/timeStampMixin.py
@@ -1,17 +1,17 @@
-from datetime import datetime
+from datetime import datetime, UTC
+
 from sqlalchemy import Column, DateTime, event
 
 
 class TimeStampMixin(object):
-    """ Timestamping mixin
-    """
-    created_at = Column(DateTime, default=datetime.utcnow)
-    updated_at = Column(DateTime, default=datetime.utcnow)
+    """ Timestamping mixin """
+    created_at = Column(DateTime, default=lambda: datetime.now(UTC).replace(tzinfo=None))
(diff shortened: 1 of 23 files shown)


dataactcore/utils/timeStampMixin.py
```diff
@@ -1,17 +1,17 @@
...
-    updated_at = Column(DateTime, default=datetime.utcnow)
+    """ Timestamping mixin """
+    created_at = Column(DateTime, default=lambda: datetime.now(UTC).replace(tzinfo=None))
+    updated_at = Column(DateTime, default=lambda: datetime.now(UTC).replace(tzinfo=None))
 
     @staticmethod
     def _updated_at(mapper, connection, target):
         if not getattr(target, "ignore_updated_at", None):
-            target.updated_at = datetime.utcnow()
+            target.updated_at = datetime.now(UTC).replace(tzinfo=None)
```
dpb-bah, line 14: Does the `TimeStampMixin` not work if you use `get_utc_now()`?

dpb-bah (changes requested):

dataactcore/utils/timeStampMixin.py
alburde1: It causes circular imports because this base is imported into every single model so we can't then import models into function bag if we do that

dpb-bah (approved):

PR #2655, opened 2024-10-10
DEV-10377: b12 updates

alburde1:
**High level description:**
Removing B12.1 and instead making `ByDirectReimbursableFundingSource` required for file C. Removing all USSGL exceptions for B12.3

**Technical details:**
N/A

**Link to JIRA Ticket:**
[DEV-10377](https://federal-spending-transparency.atlassian.net/browse/DEV-10377)

- [x] Merged after [Backend#2654](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/2654)
- Documentation updated


dpb-bah (changes requested):
Could we do these renames for consistency/clarity given this ticket has moved things around?

* `b12_object_class_program_activity_1` -> `b12_object_class_program_activity_3`
* `b12_object_class_program_activity_2` -> `b12_object_class_program_activity_4`

alburde1:
> Could we do these renames for consistency/clarity given this ticket has moved things around?
> 
>     * `b12_object_class_program_activity_1` -> `b12_object_class_program_activity_3`
> 
>     * `b12_object_class_program_activity_2` -> `b12_object_class_program_activity_4`

That is actually less consistent. If you look at A1 we don't do numberings at all because they're different file types and these are the 1st and 2nd for this file type rather than the 3rd and 4th. Another example of this is directly above this file with B11 where we have this exact same numbering vs label split

dpb-bah:
Nah that's fine. I still find this labeling confusing alongside A1 and B11 but given that's not in the scope of this, we can just move on normally for now.

PR #2680, opened 2024-12-13
[DEV-11811] - Add legacy city codes (PW, FM, MH) to the city code loader

zachflanders-frb:
**High level description:**
This PR re-adds missing legacy city codes that is not present for certain "states".

**Technical details:**
I tested this locally using a  minio container to replicate the S3 functionality.  

**Note:**
I also fixed an unrelated test that was randomly failing. See [comment](https://github.com/fedspendingtransparency/data-act-broker-backend/pull/2680#discussion_r1884475378).

**Link to JIRA Ticket:**
[DEV-111811](https://federal-spending-transparency.atlassian.net/browse/DEV-11811)

### DIFF
--- a/tests/unit/dataactcore/utils/test_tracing.py
+++ b/tests/unit/dataactcore/utils/test_tracing.py
@@ -138,8 +138,8 @@ def test_logging_trace_spans(caplog, capsys):
 
     captured = capsys.readouterr()
     assert test_msg in caplog.text, "caplog.text did not seem to capture logging output during test"
-    assert hex(trace_id) in captured.out, "trace_id not found in logging output"
-    assert hex(span_id) in captured.out, "span_id not found in logging output"
+    assert f'{trace_id:x}' in captured.out, "trace_id not found in logging output"
+    assert f'{span_id:x}' in captured.out, "span_id not found in logging output"
     assert f"{test}_resource" in captured.out, "traced resource not found in logging output"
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/dataactcore/utils/test_tracing.py
```diff
@@ -138,8 +138,8 @@ def test_logging_trace_spans(caplog, capsys):
 
     captured = capsys.readouterr()
     assert test_msg in caplog.text, "caplog.text did not seem to capture logging output during test"
-    assert hex(trace_id) in captured.out, "trace_id not found in logging output"
-    assert hex(span_id) in captured.out, "span_id not found in logging output"
+    assert f'{trace_id:x}' in captured.out, "trace_id not found in logging output"
+    assert f'{span_id:x}' in captured.out, "span_id not found in logging output"
```
zachflanders-frb, line 142: The `hex` function was not including leading zeros after the `0x` in the hex string.  This was causing this test to randomly fail.  I used python's built in formatter to format this as a string without the leading `0x` in order to address this issue.  See [docs](https://docs.python.org/3/library/functions.html#hex).

dpb-bah: Great catch! I was so curious about the root cause of that test randomly failing. Thank you.

dpb-bah (approved):
Nice improvements, tested locally and confirmed it works, LGTM ✅

PR #2690, opened 2025-01-15
Sprint 172: Qat -> Staging

alburde1:


dpb-bah:
* Updated several PARK and USSGL columns in File B and C to match the terse 30 headers.
* Removed the FY24 Submission Schedule calendar and updated the Validation Rules link in the Resources Help page per GSDM 1.1.
* Removed the sample file links in the DABS Submission Guide. They are still available and current via the GSDM link in the same section.

PR #2697, opened 2025-01-24
[DEV-11948] Add new B29 validation rules

zachflanders-frb:
**High level description:**
This PR adds two new rules (B29 and B29.1) and renames the old B29 rule to B29.2.  B29 checks that the prior year adjustment field is a valid value (X, B, or P) for file C and B29.1 checks that the prior year adjustment field is a valid value (X, B, P, or Blank) for file B.

**Technical details:**
The current code does not allow `None` values in file C;  they must always be (X, B, or P).  `None` values are treated the same as a blank string `''` in file B.   

**Link to JIRA Ticket:**
[DEV-11948](https://federal-spending-transparency.atlassian.net/browse/DEV-11948)


dataactvalidator/config/sqlrules/b29_object_class_program_activity.sql
```diff
@@ -0,0 +1,11 @@
+-- For File B PriorYearAdjustment must be one of X, B, or P.
+SELECT
+    row_number,
+    prior_year_adjustment,
+    display_tas AS "uniqueid_TAS"
+FROM object_class_program_activity
+WHERE (submission_id = {0}
+    AND (UPPER(prior_year_adjustment) NOT IN ('X', 'B', 'P')
+        OR prior_year_adjustment IS NULL
```
zachflanders-frb, line 9: Do we want to allow NULLs?  This will fail if the value is null.  Since we do not allow "blank" values in this field (unlike file C) I set this to fail validation for blank strings and nulls.

dataactvalidator/config/sqlrules/b29_award_financial_1.sql
```diff
@@ -0,0 +1,10 @@
+-- For File C PriorYearAdjustment must be one of X, B, P, or Blank
+SELECT
+    row_number,
+    prior_year_adjustment,
+    display_tas AS "uniqueid_TAS"
+FROM award_financial
+WHERE (submission_id = {0}
+    AND UPPER(prior_year_adjustment) NOT IN ('X', 'B', 'P', '')
+    AND prior_year_adjustment IS NOT NULL
```
zachflanders-frb, line 9: I am treating nulls the same here as a blank string.  A null value will pass validation.

dpb-bah (changes requested):

dataactvalidator/config/sqlrules/b29_award_financial_1.sql
```diff
@@ -0,0 +1,9 @@
+-- For File C PriorYearAdjustment must be one of X, B, P, or Blank
+SELECT
+    row_number,
+    prior_year_adjustment,
+    display_tas AS "uniqueid_TAS"
+FROM award_financial
+WHERE (submission_id = {0}
+    AND UPPER(COALESCE(prior_year_adjustment, '')) NOT IN ('X', 'B', 'P', '')
+);
```
dpb-bah, line 9: Can you also remove the extra parens around the entire `WHERE` for both rules? We'll update the styleguide such that parens should only be used as logical groupings within the `WHERE`, not the entire `WHERE` itself.

dpb-bah (changes requested):

dpb-bah (approved):

PR #2704, opened 2025-02-05
Sprint 198 Staging Deploy

dpb-bah:


dpb-bah:
• Reworked rule B29 include a check for PYA domain values in Files B and C. Rules B29 and B29.1 will do these additional checks while rule B29.2 will include the primary logic.
• Created new raw subaward tables in preparation of migrating from FSRS to SAM for subaward data.
• Corrected a minor typo in the email signup link on the Help page.

PR #2712, opened 2025-03-03
[DEV-11953] - Broker Subaward SAM Deletions

zachflanders-frb:
**High level description:**
Adds the delete_subawards method.

**Technical details:**

**Link to JIRA Ticket:**
[DEV-11953](https://federal-spending-transparency.atlassian.net/browse/DEV-11953)

- [ ] Style Guide check completed


dataactbroker/helpers/script_helper.py
```diff
@@ -93,8 +93,14 @@ def validate_load_dates(arg_start_date, arg_end_date, arg_auto, load_type, arg_d
         logger.error('start_date, end_date, or auto setting is required.')
         raise ValueError('start_date, end_date, or auto setting is required.')
 
-    if start_date and end_date \
-            and datetime.strptime(start_date, output_date_format) >= datetime.strptime(end_date, output_date_format):
+    if (
```
dpb-bah, line 96: Did this come from `black`? We generally try avoiding short lines in conditions unless the grouping makes sense (like the original code). Down the road we can reconsider using black or automatic formatting but it should be all at once instead of piecemeal and inconsistent.

dataactcore/config.py
```diff
@@ -10,7 +10,7 @@
 CONFIG_DB = {}
 CONFIG_LOGGING = {}
 CONFIG_CATEGORIES = {"broker": CONFIG_BROKER, "services": CONFIG_SERVICES, "db": CONFIG_DB,
-                     "logging": CONFIG_LOGGING}
+                     "logging": CONFIG_LOGGING, "sam": CONFIG_BROKER}
```
dpb-bah, line 13: This is a clever idea to have the categories include each type of load/data but it's a little confusing if it's still just linking back to one of the three ("broker"). We can maybe address having configs for every loader down the road.

dataactcore/scripts/pipeline/load_sam_subaward.py
```diff
@@ -288,36 +295,29 @@ def parse_raw_subaward(raw_subaward_dict, data_type):
...
-def store_subawards(sess, new_subawards, data_type):
-    """ Load the new subawards into the database
-
-        Args:
-            sess: database connection
-            new_subawards: dataframe of the new subaward data to be stored
-            data_type: either "contract" or "assistance"
-    """
+def store_subawards(sess: Session, new_subawards: pd.DataFrame, model: SAMSubgrant | SAMSubcontract) -> None:
+    """ Load the new subawards into the database"""
```
dpb-bah, line 299: StyleGuide: can you keep the `Args:` and just update `data_type` to model?

Additionally I know on USAS they utilize Python's type checking. We can have an tech debt effort to enforce throughout the code but for now can we remove it for consistency?

dpb-bah (changes requested):

dataactbroker/helpers/script_helper.py
zachflanders-frb: No, adding `datetime.datetime.strptime` instead of `datetime.strptime` pushed the line length past 140, so I put each condition on its own line.  I thought it helped with readability.

dataactcore/config.py
zachflanders-frb: I will remove this, I originally didn't understand that the  `"sam"` was under `"broker"` so was adding it as a top level config category.

dataactbroker/helpers/script_helper.py
dpb-bah: Ah I see, my mistake. I see what you had to work with - it being kinda a long mess regardless. We've generally tried to fit as much as we can on every line to the flake8 width limit with exceptions to specific logical groupings. We can add that to the StyleGuide as beforehand it was just a casual agreement between the two lead devs.

```suggestion
    if start_date and end_date and (datetime.datetime.strptime(start_date, output_date_format)
                                    >= datetime.datetime.strptime(end_date, output_date_format)):
```

dpb-bah (changes requested):

dataactbroker/helpers/script_helper.py
```diff
@@ -93,8 +93,15 @@ def validate_load_dates(arg_start_date, arg_end_date, arg_auto, load_type, arg_d
...
-            and datetime.strptime(start_date, output_date_format) >= datetime.strptime(end_date, output_date_format):
+    if start_date and end_date and (datetime.datetime.strptime(start_date, output_date_format)
+                                    >= datetime.datetime.strptime(end_date, output_date_format)):
+        start_date
+        and end_date
+        and (
+            datetime.datetime.strptime(start_date, output_date_format)
+            >= datetime.datetime.strptime(end_date, output_date_format)
+        )
+    ):
```
zachflanders-frb, line 104:
```suggestion
```

dpb-bah (approved):

PR #2746, opened 2025-04-22
[DEV-12100] - Update assistance unique award key generation

zachflanders-frb:
**High level description:**
This PR updates the FABS unique_award_key derivation to replace the sub tier agency code with the agency code.

**Technical details:**
I replaced the sql found in the fabs_derivations_helper with this function using pandas and sqlalchemy to create the mapping of subtier codes to agency codes.
```python
    def retrieve_sub_tier_agencies(chunk_df, sess):
        return pd.read_sql(
            sess.query(
                SubTierAgency.sub_tier_agency_code,
                case((SubTierAgency.is_frec, FREC.frec_code), else_=CGAC.cgac_code).label('agency_code')
            )
            .join(CGAC, SubTierAgency.cgac_id==CGAC.cgac_id)
            .join(FREC, SubTierAgency.frec_id==FREC.frec_id)
            .filter(SubTierAgency.sub_tier_agency_code.in_(chunk_df['awarding_sub_tier_agency_c']))
            .statement,
            sess.connection()
        )
```
I also vectorized the `derive_fabs_unique_award_key` functino to operate on the entire database instead of driving the unique key row-wise.

**Link to JIRA Ticket:**
[DEV-12100](https://federal-spending-transparency.atlassian.net/browse/DEV-12100)

[DEV-12100]: https://federal-spending-transparency.atlassian.net/browse/DEV-12100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dataactcore/migrations/versions/ac14a84e80a3_add_agency_code_and_name_to_fabs.py
```diff
@@ -0,0 +1,42 @@
+"""empty message
```
dpb-bah, line 1:
```suggestion
""" Adding agency code/name to fabs
```

dpb-bah: This is usually populated by `alembic revision -m "[message here describing the migration"]`

dpb-bah (changes requested):
I apologize on my part for not being explicit in the ticket.

We were thinking of adding the awarding agency code/name to the staging `FABS` table since the new `unique_award_key` was going to need it anyways. So the hope was deriving the awarding agency code/name in the dataframe, and then simply using the new code column to derive the `unique_award_key`, instead of deriving them later on when publishing. Right now it looks like the two new columns will be added but blank in FABS and thatd be copied into PublishedFABS too.

Also realizing that `awarding_sub_tier_agency_n` would lose its derivation here. Could you include `awarding_sub_tier_agency_n ` in the migration as well and derive it here too? Let me know if this is too much scope creep and I can jump on it if need be.

Still think its a good idea updating the dataframe as a whole versus updating each row individually. I think it'd be a performance boost overall this way but we should make sure running locally on a not super small submission to ensure of that performance.

dpb-bah:
Upon further discussion, we've settled on:
* only adding `awarding_agency_code` for `FABS`, deriving that on staging FABS upload/validation, and using that to derive the `unique_award_key`
* leaving the derivations for `awarding_sub_tier_agency_n ` and `awarding_agency_name` in FABS derivations on publishing.

dataactvalidator/validation_handlers/validationManager.py
```diff
@@ -895,6 +901,20 @@ def process_data_chunk(self, chunk_df, shared_data, file_row_count, sess=None, l
                 'status': 'end'
             })
 
+    @staticmethod
+    def retrieve_sub_tier_agencies(chunk_df, sess):
```
dpb-bah, line 905: Can this be renamed to something like `retreive_agency_codes`? Same with the resulting variable `sub_tier_agencies` -> `agency_codes`.

dpb-bah (changes requested):
Looks pretty good. In addition to the unit tests, can you run it locally with a small and relatively sized submission to make sure performance isn't impacted? If not I can check for performance, just making sure it's been tested outside of unit/integration tests as it can reveal something unexpected.

dpb-bah (approved):

PR #2916, opened 2026-08-25
[DEV-15812] - Sam Loader Memory Improvements

zachflanders-frb:
**High level description:**
This PR improves the memory efficiency of the sam loader.

**Technical details:**
The biggest spike in memory occurred in the insert_into_db function where the code unpacked each row and each value from the dataframe into a python dictionary of strings.  Keeping the data in a dataframe format and instead inserting into a temp table and then doing an upsert from the temp table reduces the memory demand of this operation.

**Link to JIRA Ticket:**
[DEV-15812](https://federal-spending-transparency.atlassian.net/browse/DEV-15812)

[DEV-15812]: https://federal-spending-transparency.atlassian.net/browse/DEV-15812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dataactcore/scripts/pipeline/load_sam_contract.py
+++ b/dataactcore/scripts/pipeline/load_sam_contract.py
@@ -11,21 +11,7 @@ import boto3
 import numpy as np
 import pandas as pd
 
-from sqlalchemy import (
-    func,
-    case,
-    types,
-    Boolean,
-    Numeric,
-    ARRAY,
-    BigInteger,
-    text,
-    Integer,
-    Float,
-    DateTime,
-    Date,
-)
-from sqlalchemy.inspection import inspect
+from sqlalchemy import func, case, types, Boolean, text
 
 from dataactbroker.helpers.script_helper import validate_load_dates
 from dataactbroker.helpers.generic_helper import unzip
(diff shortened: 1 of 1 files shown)


dataactcore/scripts/pipeline/load_sam_contract.py
```diff
@@ -1425,66 +1425,37 @@ def main():
...
+        "metrics": metrics_json,
+    }
     if args.feed in ["add", "both"]:
         insert_start = get_utc_now()
         logger.info(f"Starting data collection at: {str(insert_start)}")
+        if args.local_file is not None:
+            get_data(**get_data_params)
+        else:
+            for award_type in award_types_idv:
+                get_data(**get_data_params, contract_type="IDV", award_type=award_type)
```
zachflanders-frb, line 1448: If we are using a local file, we just need to call `get_data` once since the contract_type and award_type are not used when using a local file and was resulting in the same operation multiple times.

dpb-bah: I like this generally, but it does look like those _do_ get used later in `process_data` to determine which mappings to use.

```
    contract_mappings = SAM_CONTRACT_MAPPINGS
    if contract_type == "IDV":
        contract_mappings = contract_mappings | IDV_MAPPINGS
    else:
        contract_mappings = contract_mappings | AWARD_MAPPINGS
 ```
 
We could work from the filename and passing those in but then its dependent on that filename format. Could make additional commandline arguments with `--local_file` to communicate the `award_type` and `contract_type`?

Or could just leave it alone as its more local testing which can edit the file to your liking anyway.

dpb-bah (changes requested):
LGTM, just needing to figure out that local situation

dpb-bah:
FYI we tested this code in Staging to compare with existing FPDS data and it ran much faster/smoother 🎉

dataactcore/scripts/pipeline/load_sam_contract.py
zachflanders-frb: Good point.  I reverted this back to use the for loops for award type with a local file.  I kept the params dictionary since most the params were reused and this reduces code duplication.

dpb-bah (approved):

